PR #40, opened 2021-08-02
mature.js updated for full matches

### LINKED ISSUE #31: Adult filter sometimes filters too much.
`All my life, my parents have told me not to open the ba******t door, but I got curious and disobeyed them.`

Here, it's just "semen" which is removed but actually it should not. I don't have a solution right now.

gokaykucuk:
closes #31, as solution I've added some spaces before and after the keywords. So they shouldn't match partial words anymore.


app/src/backend/mature.js
```diff
@@ -1,8 +1,7 @@
+import {matureWords} from "./matureWords.json"
```
voodoohop, line 1: The netlify function is built using `esbuild` which as far as I understood doesn't allow directly importing of .json's like this. Maybe better to replace with a `readFile` operation

voodoohop (approved):

app/src/backend/mature.js
gokaykucuk: makes sense, i'll update it

PR #46, opened 2021-08-09
Package json removed

gokaykucuk:
We had both yarn.lock and package-json.lock. This PR removes package-json.lock


voodoohop:
I read a few articles and came to the conclusion that I would prefer the npm package-lock.json and move away from yarn. It seems yarn doesn't offer that many more benefits anymore and I find the v2 of yarn goes in the wrong direction. But I think in the end it doesn't make a big difference.

Would you mind if we went back to npm though? it's just a gut feeling

voodoohop:
Ah cool! Already merged. I was reading something about the ins and outs of yarn. We should stick with v1

PR #48, opened 2021-08-10
Easy mode for hiding all the params except primaryInput

gokaykucuk:
We still have to carry it to state but for now we can just hide the other parameters.

### DIFF
--- a/app/src/components/Form.js
+++ b/app/src/components/Form.js
@@ -59,13 +59,11 @@ export default React.memo(FormView)
 function getFormInputs(ipfs, metadata) {
     if ((metadata === undefined) || (metadata === null)) return;
     ipfs = ipfs || {};
-    debug("properties", metadata);
     const allParams = Object.fromEntries(Object.entries(metadata.form.properties).map(
         ([formKey, prop]) => [formKey, formKey in ipfs ? {...prop, "default": ipfs[formKey]} : prop]));
 
     // TODO: Carry it to state so users can disable it.
     const easyMode = true;
-    // debug("all params", fil);
     if(easyMode){
         return filter((param) => (param.title === metadata.primaryInput), allParams);
     }else{
--- a/app/src/components/Form.js
+++ b/app/src/components/Form.js
@@ -2,31 +2,32 @@ import React from "react";
 import Form from "@rjsf/material-ui";
 import Button from '@material-ui/core/Button'
 import Debug from "debug";
-import { Box, Typography } from "@material-ui/core";
+import {Box, Typography} from "@material-ui/core";
 import HelpModal from "./HelpModal";
+import {filter} from "ramda";
+
 
 const debug = Debug("Form");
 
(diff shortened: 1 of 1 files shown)


app/src/components/Form.js
```diff
@@ -58,9 +59,16 @@ export default React.memo(FormView)
 function getFormInputs(ipfs, metadata) {
     if ((metadata === undefined) || (metadata === null)) return;
     ipfs = ipfs || {};
-
-    return Object.fromEntries(Object.entries(metadata.form.properties).map(
-            ([formKey, prop]) => [formKey, formKey in ipfs ? { ...prop, "default": ipfs[formKey] } : prop]))
+    const allParams = Object.fromEntries(Object.entries(metadata.form.properties).map(
```
gokaykucuk, line 62: @voodoohop real change is here, the rest got auto-fixed by my ide.

voodoohop:
I see 4 options to enable this:
1) a query parameter that enables "easy" mode or the opposite to enable "expert" mode
2) another prefix e.g. /e/[hash] is the easy mode, /p/[hash] is the (current) expert mode
3) localStorage
4) we don't code an easy mode into the pollinations code but have expert and easy notebooks using different notebook.ipynb's so one would have only one parameter and the expert one would make more available using the built-in colab parameters

voodoohop:
> I see 4 options to enable this:
> 
> 1. a query parameter that enables "easy" mode or the opposite to enable "expert" mode
> 2. another prefix e.g. /e/[hash] is the easy mode, /p/[hash] is the (current) expert mode
> 3. localStorage
> 4. we don't code an easy mode into the pollinations code but have expert and easy notebooks using different notebook.ipynb's so one would have only one parameter and the expert one would make more available using the built-in colab parameters

@gokaykucuk which one do you think makes most sense?

gokaykucuk:
> I see 4 options to enable this:
> 
> 1. a query parameter that enables "easy" mode or the opposite to enable "expert" mode
> 2. another prefix e.g. /e/[hash] is the easy mode, /p/[hash] is the (current) expert mode
> 3. localStorage
> 4. we don't code an easy mode into the pollinations code but have expert and easy notebooks using different notebook.ipynb's so one would have only one parameter and the expert one would make more available using the built-in colab parameters

@voodoohop I think a combination of localStorage and query parameter makes sense. We can bind the query parameter to a value in localStorage and use it as a way of switching modes.

ex:

user1 navigates to pollinations first time => redirect to easy mode with ?mode=easy and store mode=easy in localStorage
user1 navigates to pollinations with mode=expert in query => store mode=expert in localStorage
user1 navigates to pollinations with mode=expert in localStorage => update url to ?mode=expert.

voodoohop:
Another option would be simple collapsible advanced options. I kind of like it because it adds no state complexity

advanced-options-collapsed
advanced-options-expanded

voodoohop:
@gokaykucuk I really like just having the expandable "> advanced option". if you think it's good enough for now I'd suggest we go ahead with it. I can implement it and merge or if you have time.

PR #57, opened 2021-08-16
[WIP] ContentID updates shouldn't update history

### LINKED ISSUE #23: Don't create history when updating ContentID

gokaykucuk:
Closes #23


voodoohop:
https://stackoverflow.com/questions/26793130/history-replacestate-still-adds-entries-to-the-browsing-history

looks like this is not possible. let's close for now. later we can think of only changing url at end of run and keeping contentid in regular react state during notebook processing. lets make that 0.2.0

voodoohop:
OK. I found a solution. We only update the URL when the run is done. Intermediate state is just in react

app/src/backend/matureWords.json
```diff
@@ -1,5 +1,8 @@
 {
- "words": [
+ "words": [,
+   "muhammad",
```
gokaykucuk, line 3: @voodoohop why "muhammad" is a mature word?

PR #58, opened 2021-08-16
Disable connecting from browser to local IPFS node by default

voodoohop:
Running a local IPFS node is breaking pollinations. If I remember correctly it needed some special configuration to work.


gokaykucuk (approved):
LGTM!

PR #102, opened 2021-09-20
Unify and simplify IPFS handling

voodoohop:


autohashtag.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+curl -H "Authorization: Bearer 1SQJNTJ-KSZ41EF-N2HXXZN-BEDYDYC" -d '{"post": "the next time i found this walrus", "max": 3, "position": "auto"}' -X POST https://app.ayrshare.com/api/auto-hashtag
```
gokaykucuk, line 2: @voodoohop there is a token written here, is it safe?

PR #257, opened 2022-03-17
supabase user management initialised

gokaykucuk:


app/src/components/TopBar.js
```diff
@@ -28,58 +45,121 @@ const TopBar = ({ showNode }) => {
...
+                        <RouterLink {...linkProps}/>
+                    </li>)
+                }
+                <SocialLinks style={{alignSelf: 'end'}}/>
             </MenuItems>
         </HiddenContentStyle>
-        
+
+
+        <Dialog open={loginOpen}>
```
voodoohop, line 78: Move to separate component

app/src/components/TopBar.js
```diff
@@ -28,58 +45,121 @@ const TopBar = ({ showNode }) => {
...
+                {loginProviders?.map((provider) => (
+                    <ListItem button onClick={() => handleSignIn(provider)} key={provider}>
+                        <ListItemAvatar>
+                            <Avatar src={`/socials/${provider}_white.png`}/>
+                        </ListItemAvatar>
+                        <ListItemText primary={provider}/>
+                    </ListItem>
+                ))}
+            </List>
+            <DialogActions>
```
voodoohop, line 90: also separate component

app/src/components/TopBar.js
```diff
@@ -28,58 +45,121 @@ const TopBar = ({ showNode }) => {
...
+            </DialogActions>
+        </Dialog>
+
+
+
     </Container>
 }
 
+
+const LoggedUser = ({ user }) => {
```
voodoohop, line 101: move to separate component

app/src/supabase/pollen.js
```diff
@@ -0,0 +1,39 @@
...
+    return supabase.from("baseapp_pollen").select("*").order("id", {ascending: false}).then(response => {
+        return response.data
+    })
+}
+
+export function createPollen(input, cid, ipns) {
+    return supabase.from("baseapp_pollen").insert([{
+        "cid": cid,
+        "ipns": ipns,
+        "text_prompt": text_prompt,
```
voodoohop, line 13: this is too specific to image notebooks. i think this should also not be saved in a table but as an ipfs folder with just one CID

gokaykucuk: I just put those ipns and CID as a guess we can leave the ones we don't use as empty. text_prompt was an idea for search in the future, but again we can leave it empty

voodoohop: let's create an indexing db at some point that just indexes all the inputs of the pollens so we can search through any metadata. maybe a document storage would make more sense since the fields can be completely different depending on which notebook is used?

voodoohop: like mongo-style but on postgres. i think it allows saving arbitrary jsons which you can search over

gokaykucuk: @voodoohop yes, there was already an "inputs" field which was json, I'll add that back in. I'll also add an "output" field as json which we can use

PR #259, opened 2022-03-17
Eslint

gokaykucuk:
Eslint rules added with StandardJS preset


voodoohop (approved):
Shall we just go with this one for now?

PR #270, opened 2022-03-25
feat: masonry logic added for notebook cards

fatihgozenc:


voodoohop:
@fatihgozenc I'm not 100% sure about the utility of this masonry logic the way the notebook list is at the moment. Maybe I would prefer keeping it simpler with CSS and just defining a fixed size per card. Can we postpone the masonry decision until later when we are more clear about the overall design?

gozenc:
> @fatihgozenc I'm not 100% sure about the utility of this masonry logic the way the notebook list is at the moment. Maybe I would prefer keeping it simpler with CSS and just defining a fixed size per card. Can we postpone the masonry decision until later when we are more clear about the overall design?

sure fine by me, I can continue on other pr's

gokaykucuk:
@fatihgozenc @voodoohop I'm closing this one then, we can take another look in the future if we need masonry logic

Issue #273, opened 2022-03-28
Create a cron job to enter the public-ipfs docker container and run ipfs repo gc


gokaykucuk:

voodoohop:
@gokaykucuk  have you tried the --enable-gc for runtime gc? maybe it was just failing on our other node because it was so much data and a slow hd

gokaykucuk:
@voodoohop i think we can merge it with writing our own dockerfile than. Makes sense, I'll take this as the next item.

PR #282, opened 2022-03-31
[WIP] - Added public ipfs dockerfile

### LINKED ISSUE #258: Create a dockerfile for public ipfs instance

gokaykucuk:
Closes #258  and #273 

@voodoohop do you have any idea how should we deal with private key?


gokaykucuk:
@voodoohop one idea from my side is to put them through ENV, the same goes for nft.storage keys

PR #284, opened 2022-04-03
Fine grained progress bar

voodoohop:
Screenshot 2022-04-03 at 19 03 47


gokaykucuk (approved):
looks good! I liked that it's not replacing the existing functionality but supporting it

PR #320, opened 2022-05-09
Markdown expos

alexreiling:
Implemented basic architecture for "made with pollinations" articles. Made some more or less unrelated adjustments to the code base during the process:
- Added prettier config
- Introduced generic `Icon` component that renders 1024x1024 (unitless) vector paths as svgs, also set up initial vector path lib
- refactored main nav a little bit to make it more visible in the app hierarchy, also should increase robustness by declaring routes in central config object instead of passing strings as content ids
- refactored `MarkdownContent` component to accept content url (instead of content id) as prop to make it more flexible
- split up `useContent()` hook into `useFetchText()` and `useMarkdown()` hook to separate concerns a little bit


alexreiling:
@voodoohop, @ale-rls can we merge this?

PR #620, opened 2023-12-13
Update matureWords.js

flowa-ai:


voodoohop:
I don't think we're using the matureWords.js anymore so I don't know if we need this. I'm going to clean up the repository and add the generative image url backend code soon

PR #660, opened 2024-09-15
Is link image

diogo-karma:


voodoohop (commented):
I like the documentation updates and the example apps.

I'm not sure about the link functionality though. It seems like a functionality for a specific use case.

 one can quite easily use the hook and add a link around an image if one needs it.

chakra-gold:
You're absolutely right. I was looking into onClick events on the image to build out several gallery examples using the models, but there's no need for the <a> tag. The  tag already supports the necessary events, including mouseover, mouseenter, mouseleave, mousemove, mousedown, mouseup, and all other mouse events, as well as touch events like touchstart, touchmove, and touchend. Let's focus on the  tag in the Readme and make the most of it. I'm currently focused on building out the examples.

PR #1099, opened 2025-01-27
replaced copy email by google form to submit project

ElliotEtag:


voodoohop:
@ElliotEtag After thinking about it I think the Google Form doesn't fit so well into our architecture. I think we could take people to a template issue they can fill in on GitHub which should trigger our issue resolver bot to add it by itself. This feature is for Devs so they should be comfortable with GitHub

check https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

they have a feature called issue forms which automatically creates a form for an issue template it seems.

voodoohop:
@mentatbot is it possible to implement my suggestion to use issue templates through configuration files or do i need to update the settings of the repository?

voodoohop:
@mentatbot create a PR pleas

ElliotEtag:
> @ElliotEtag After thinking about it I think the Google Form doesn't fit so well into our architecture. I think we could take people to a template issue they can fill in on GitHub which should trigger our issue resolver bot to add it by itself. This feature is for Devs so they should be comfortable with GitHub
> 
> check https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository
> 
> they have a feature called issue forms which automatically creates a form for an issue template it seems.

okay i like it too ✌️

PR #1102, opened 2025-01-28
1088 pollinationsai re add projects that were lost in last UI update

voodoohop:


voodoohop:
@mentatbot can you review?

PR #1133, opened 2025-02-03
Add `nohup.out` to gitignore

MinecraftPublisher:
`nohup.out` is a log file, and pushing it to the repository causes the repository's size to increase with no clear benefits.


MinecraftPublisher:
Replaced by #1134.

PR #1234, opened 2025-02-25
[text.pollinations.ai] Streaming and Function Calling

voodoohop:


ekaship8-cmyk:
I'm experiencing the same issue with my AI photo editor desktop app (Python/PyQt6). 

The `openai` model works fine for image descriptions via `gen.pollinations.ai`, but when I try `gemini` or `claude`, I get a `400 Invalid model` error.

It would be great to have access to Gemini for vision tasks since it's optimized for that. Is there an ETA on when these models might be available again on the free tier?

Thanks for the great service!

PR #1253, opened 2025-02-27
added referrer rule (news + apidocs)

ElliotEtag:


voodoohop:
@mentatbot 

a ideia e que se voce quer tirar logo ou um link pro pollinations que adicionamos depois do texto do text API ou na imagem que giramos tem que nos passar um referrer
BrainDiver — Yesterday at 19:47
Excelente. É o minimo
Um ótimo passo
Até para entender como tão usando, ter métricas mais precisas 
Isso vai ajudar muito nas tomadas de decisões...ótimo passo mesmo
Quer que eu crie um bot para postar esses announcements? hahaha
Nunca criei um bot de discord, seria divertido
Você ja deve ter né
Vc vai postar quando?
thomash — Yesterday at 19:54
pensei depois de terminar um outro feature branch tipo mais tarde ou amanha
vamos botar nos api docs tambem
no mesmo momento
BrainDiver — Today at 18:20
I just realized that this might affect my app. I will probably have to roll back the closed testing of the Abyss Ascending app from the PlayStore to add the referer otherwise the responses from Llama will come with a link to Pollinations once this new condition is implemented, right?
thomash — Today at 18:27
yeah im not sure about adding the link yet to each text response
could be a bit disruptive to existing apps
i was thinking of having some kind of classifier that only adds it at the end of say markdown style blog responses
but not say in a conversation or something because it doesnt make sense
to drop a link after each message. also people ask for json etc
we didnt really think the text link through yet
but if its a conversational thing then i would not add the link anyway
BrainDiver — Today at 18:34
Yes, that is something to consider. Anyway, for my app is not a problem because there are no actual users yet, I can add later an update with the code. But for ppl that already have a lot of users that can be a bit disruptive indeed.
The earliest before the implementation that they get the announcement better to have able time to add the referer to their apps


what do you think of this feedback?

PR #1374, opened 2025-03-16
consolidate api documentation

voodoohop:


voodoohop:
@mentatbot what do you think? just review

voodoohop:
WithThatWay commented


After conducting several experiments with Pollinations' documentation, if you use the OpenAI endpoint, it says:
"Just follow OpenAI standard format."

However, in practice, some parameters are only supported when using:

The developer role to replace system, which applies to all OpenAI models in Pollinations and Claude 3.5 Haiku.
Other models will ignore it and fallback to their own base model format.

maxTokens Handling
Following the OpenAI format, max_tokens has been deprecated in favor of max_completion_tokens.
However, this change is only supported in OpenAI models. Other models still accept only max_tokens.

seed Parameter
Specifically for Mistral, they only accept random_seed instead of seed.

Conclusion
Following the OpenAI standard format only works if the selected model is truly an OpenAI model. Otherwise, each model follows its own parameters as defined by its origin.

@elliot @thomash

PR #1757, opened 2025-04-21
added param to remove cloudflare invocation logs

ElliotEtag:


voodoohop:
@ElliotEtag this is a configuration in cloudflare dashboard i think

ElliotEtag:
@voodoohop 
It worked, i followed Cloudflare instructions, there is also a way to go this via dashboard i think but i prefered the programatic way

PR #1927, opened 2025-05-08
Update generateTextPortkey.js

withthatway:
Just my suggestion: that portkey should no longer be accessible. If a user tries to access a portkey like `https://text.pollinations.ai/Hai?model=openai-xlarge`, ideally, it should fall back to nano since the specified model doesn't exist—right, @voodoohop?


withthatway:
We ran out of credits for some of those models, which is why we "removed" them from the portkey to make them inaccessible.

PR #2003, opened 2025-05-19
Added the web search agent and it's test html file

Circuit-Overtime:
## Added the web search agent for the improvement of the current searchgpt model
- Uses the current `openai-small` and `openai-large` as a combination to form the results
- Finds the most accurate and the recent result from dynamic searching and leverages true thinking to dynamically navigate routes 

## Future Plans 
- Integrate function calling
- Integrate streaming mode 

Functional Difference between searchgpt and the code which i provided 
`Query: What's the time in kolkata IST right now?`
> query was made on 19/05/2025, 20:22 PM
Here's some result analytics 
- SearchGPT
image
---
- My Algorithm
image

## Drawbacks 
- Mine is a bit slower since it focuses more on accuracy but with a more powerful machine and network speed, it will be almost 95% the speed of searchgpt

Hoping forward to integrate this with pollinations


Circuit-Overtime:
@voodoohop  i have added all the necessary updated files for the initial version of the improved searchgpt model, it's not using bing search but leverages a native search engine mechanism of ddgs which is created smartly to avoid timeout and also uses much more context and recent toolkits for searching and parsing

PR #2103, opened 2025-05-26
Update README for SillyTavern

Cohee1207:
We now support Pollinations.ai as a chat source.


voodoohop:
@mentatbot can you also update our project list?

voodoohop:
ah i think we probably need to create a new branch and pr since this branch is on st's github

voodoohop:
ah i think we probably need to create a new branch and pr since this branch is on st's github @MentatBot

PR #2138, opened 2025-05-27
Fix: Update referrer extraction priority - query and body params now override headers

voodoohop:
## Summary

This PR updates the `extractReferrer` function in `shared/auth-utils.js` to ensure that explicitly provided referrer values (via query parameters or request body) take precedence over browser-sent headers.

## Changes

- **Query parameters** (`?referrer=...` or `?referer=...`) now have highest priority
- **Body parameters** (`referrer` or `referer` fields) have second priority  
- **Headers** (`Referer`, `Referrer`, `Origin`) now have lowest priority
- Removed POST-only restriction for body parameter checking

## Rationale

Previously, the function only checked headers and body, with headers taking precedence. This change ensures that when users explicitly provide a referrer value (either in the URL or request body), it will override any automatic browser-sent referer headers.

## Priority Order (highest to lowest)

1. URL query parameters
2. Request body parameters
3. Request headers

This provides more control to API consumers and ensures consistent behavior across different request methods.

### DIFF
--- a/image.pollinations.ai/src/index.js
+++ b/image.pollinations.ai/src/index.js
@@ -315,6 +315,21 @@ const checkCacheAndGenerate = async (req, res) => {
       'X-Auth-Status': isAuthenticated ? 'authenticated' : 'unauthenticated'
     };
     
+    // Add Content-Disposition header with sanitized filename
+    if (originalPrompt) {
+      // Create a filename from the prompt, limiting length and sanitizing
+      const baseFilename = originalPrompt
+        .slice(0, 100) // Limit to 100 characters
+        .replace(/[^a-z0-9\s-]/gi, '') // Remove special characters except spaces and hyphens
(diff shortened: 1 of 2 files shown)


voodoohop:
## Additional Updates

Added two more commits to this PR:

1. **Added Content-Disposition header for image downloads** 
   - Images now have a sanitized filename based on the prompt
   - Removes special characters and limits length to 100 characters
   - Example: prompt "a beautiful sunset" → filename "a-beautiful-sunset.jpg"

2. **Fixed TypeError crash with referrer handling**
   - Fixed server crash when `ref.includes()` was called on non-string values
   - Simplified the code using `String()` conversion instead of verbose type checking
   - Used optional chaining (`?.`) for cleaner body parameter access
   - Ensures `extractReferrer()` always returns a string or null

These changes improve both the user experience (meaningful download filenames) and server stability (no crashes from non-string referrers).

PR #2216, opened 2025-06-01
feat: Add transparency support for gptimage

Zingzy:
@voodoohop do tell me if I need to modify anything else to make this work 😃


voodoohop:
Updated the implementation:

- **Kept PNG format for gptimage model**: The code now preserves PNG format when using the gptimage model instead of converting to JPEG
- **Commented out output_compression parameter**: This parameter appears to be causing issues with the Azure GPT Image API, so it's been temporarily disabled
- **Reverted API docs changes**: Removed the transparent parameter documentation as it's not being implemented in this PR

The gptimage model will now output PNG format by default, which should provide better quality for generated images.

PR #2393, opened 2025-06-11
2336 add tier field to model list in availablemodelsjs

ElliotEtag:


voodoohop:
@mentatbot - do you think the sorting should happen in availableModels.js instead of server.js?

voodoohop:
@mentatbot implement it

PR #2502, opened 2025-06-15
2470 update pricing and remove from models list endpoint keep in data available models use pollen or similar instead of dollar

ElliotEtag:
- updated token price and allocated tier
- token price filtered out of the function output

@voodoohop 
- ignore commented out details values (other issue)
- o3 + o3-mini as alias? 
- tools=true ok for bidara?
- tier: "anonymous" could be just no tier?


ElliotEtag:
image

PR #3465, opened 2025-08-05
[text.] server: Raise JSON body size limit.

hadadarjt:
* The previous 5MB cap often wasn’t enough for high-resolution images, especially those coming from the camera or used in vision model inputs. Since Base64 encoding increases the payload size by around 33%, even a 5MB image could turn into something much larger once wrapped in JSON. This led to failed requests, server hangs, or 413 errors when the payload went over the limit.

* With the new limit, the server can now handle larger image payloads more reliably, which is especially helpful when dealing with full-quality images for OpenAI-compatible API endpoint or other use cases where resolution matters.


voodoohop:
Thanks

PR #3527, opened 2025-08-06
3348 delete operation root folder and move visual assets to the root

ElliotEtag:


ElliotEtag:
/claude check if we didn't break any link for the files in the assets folder

PR #3676, opened 2025-08-13
Pollen MVP

eulervoid:
- Add new auth backend based on better-auth
- Add polar.sh integration
- Use drizzle to manage database interaction
- Add mocks and initial test for github social sign-in

I would suggest we keep this as a draft until we are ready to merge.


voodoohop:
@claude can you attempt summarizing what was done and do a high level review?

enter.pollinations.ai/.encrypted.env
voodoohop, line 1: why is there a .encrypted.env but a .dev.vars.example?

enter.pollinations.ai/.encrypted.vars.env
voodoohop, line 1: feels to me like we should choose either .dev.vars or .env if we choose  .dev.vars wouldnt it be more coherent to use .encrypted.dev.vars.env?

enter.pollinations.ai/README.md
```diff
@@ -0,0 +1,161 @@
...
+- `GET /api/user-tiers` - Get all users with their tiers (admin only)
+
+### User Preferences Endpoints (NEW)
+- `GET /preferences` - Get user preferences
+  - Optional: `?user_id=...` to get another user's preferences (requires auth)
+- `POST /preferences` - Update user preferences
+  - Body: `{"key": "preference_name", "value": "preference_value"}` for single preference
+  - Body: `{"preferences": {"key1": "value1", "key2": "value2"}}` for multiple preferences
+
+### User Metrics Endpoints (Admin Only - NEW)
```
voodoohop: the user metrics that are saved in the user db were kind of a remnant / test. i think we dont need them. tinybird is responsible for those now

enter.pollinations.ai/drizzle/0000_dear_boom_boom.sql
```diff
@@ -0,0 +1,93 @@
...
+);
+--> statement-breakpoint
+CREATE TABLE `apikey` (
+	`id` text PRIMARY KEY NOT NULL,
+	`name` text,
+	`start` text,
+	`prefix` text,
+	`key` text NOT NULL,
+	`user_id` text NOT NULL,
+	`refill_interval` integer,
```
voodoohop, line 25: what are these? refill?

voodoohop: are these fields basically the schema the api key plugin uses?

enter.pollinations.ai/drizzle/0000_dear_boom_boom.sql
```diff
@@ -0,0 +1,93 @@
...
+	`rate_limit_time_window` integer DEFAULT 86400000,
+	`rate_limit_max` integer DEFAULT 10,
+	`request_count` integer,
+	`remaining` integer,
+	`last_request` integer,
+	`expires_at` integer,
+	`created_at` integer NOT NULL,
+	`updated_at` integer NOT NULL,
+	`permissions` text,
+	`metadata` text,
```
voodoohop, line 39: wouldnt a json type be better here or is that represented as text?

enter.pollinations.ai/drizzle/0000_dear_boom_boom.sql
```diff
@@ -0,0 +1,93 @@
...
+	FOREIGN KEY (`user_id`) REFERENCES `user`(`id`) ON UPDATE no action ON DELETE cascade
+);
+--> statement-breakpoint
+CREATE TABLE `session` (
```
voodoohop, line 43: what is this session management used for? is it an oauth thing?

enter.pollinations.ai/drizzle/0000_dear_boom_boom.sql
```diff
@@ -0,0 +1,93 @@
...
+	`user_agent` text,
+	`user_id` text NOT NULL,
+	`impersonated_by` text,
+	FOREIGN KEY (`user_id`) REFERENCES `user`(`id`) ON UPDATE no action ON DELETE cascade
+);
+--> statement-breakpoint
+CREATE UNIQUE INDEX `session_token_unique` ON `session` (`token`);--> statement-breakpoint
+CREATE TABLE `user` (
+	`id` text PRIMARY KEY NOT NULL,
+	`name` text NOT NULL,
```
voodoohop, line 59: id keep it as minimal as possible i.e. we dont have a user name yet. i dont even know if github gives us an email? if yes then we should fill it in.

voodoohop: image and email_verified i would also remove. what is role used for?

enter.pollinations.ai/scripts/put-secrets.ts
voodoohop, line 1: i think we talked about this and it may not be necessary. but if it is keep it.

enter.pollinations.ai/src/client/components/api-key.tsx
```diff
@@ -0,0 +1,400 @@
...
+
+    return (
+        <div className="space-y-4">
+            <div className="bg-amber-100 border-2 border-amber-300 rounded-lg p-4">
+                <h3 className="text-sm font-medium text-amber-800">
+                    Security Guide
+                </h3>
+                <ul className="mt-2 text-sm text-amber-800 list-disc pl-3">
+                    <li className="font-bold text-amber-900">
+                        This is the only time you'll see your API key. <br />{" "}
```
voodoohop, line 214: i know its good security practice to not be able to see the api key again but at the same time quite a few platforms i use do allow it. since we're more focused at beginner's rather than corporate id give the possibility to view the key again

enter.pollinations.ai/src/client/routeTree.gen.ts
```diff
@@ -0,0 +1,113 @@
...
+import { Route as PrivacyRouteImport } from './routes/privacy'
+import { Route as ImprintRouteImport } from './routes/imprint'
+import { Route as IndexRouteImport } from './routes/index'
+
+const SignInRoute = SignInRouteImport.update({
+  id: '/sign-in',
+  path: '/sign-in',
+  getParentRoute: () => rootRouteImport,
+} as any)
+const PrivacyRoute = PrivacyRouteImport.update({
```
voodoohop: I think having privacy and imprint routes is not necessary. this lives on pollinations.ai. im not sure we need to use a router here. id rather a very minimal spa.  but i'm also ok with just removing privacy and imprint if the router makes sense right now. i wouldnt keep it because maybe in the future we may need it

enter.pollinations.ai/src/registry.ts
voodoohop, line 1: would it be possible to move the registry and cost calculation, etc into a separate pr? it feels like the other part is kind of finished as a block. or are they intertwined? if it makes sense then id suggest we tie off a v1 with everything together as in with the current cost logic. but then the we branch of master with a more focused pr

enter.pollinations.ai/src/style.css
voodoohop, line 1: what is this style used for? i was under the impression the web interface is under src/client

flake.nix
voodoohop, line 1: remove deps that are not essential - as we discussed

enter.pollinations.ai/.encrypted.env
eulervoid: This is on purpose because `.encrypted.env` contains env variables that are not supposed to be passed to the worker, like `CLOUDFLARE_SECRET`, while the other one is equivalent to .dev.vars. I called it `.envcrypted.dev.env` in my current dev branch. I'm not using the vars ending because `sops` uses the ending to detect the file type and if it is `.vars`, it tries to parse it as json by default. But we could also change it if we want, thats just what was easiest.

enter.pollinations.ai/README.md
eulervoid: I didn't touch the readme, so this is just a remnant from the old auth service, i think it makes most sense to just delete it for now and then create a new one.

enter.pollinations.ai/drizzle/0000_dear_boom_boom.sql
eulervoid: Yes, this is the auto-generated schema of better-auths api-key plugin, refill is used for rate limiting. (see https://www.better-auth.com/docs/plugins/api-key)

enter.pollinations.ai/drizzle/0000_dear_boom_boom.sql
eulervoid: Yeah internally it is json, the schema is auto-generated, they probably do it like this to be compatible with different databases without having a lot of special handling in the schema generation.

enter.pollinations.ai/drizzle/0000_dear_boom_boom.sql
eulervoid: This is where better-auth stores the user sessions, so not necessarily oauth specific, if we used email/password, the session would also be stored there.

enter.pollinations.ai/drizzle/0000_dear_boom_boom.sql
eulervoid: We get the name and e-mail from github and it is stored there. `email_verified` keeps track if the email was verified (e.g. by sending a link and the user clicking on it, or in the oauth case i think it will be verified once the user logs in for the first time). It is a relatively important field for security as far as i'm aware, so i would leave it there, its handled automatically by better-auth.

enter.pollinations.ai/scripts/put-secrets.ts
eulervoid: Yes, its obsolete.. I'll remove it.

enter.pollinations.ai/src/client/components/api-key.tsx
eulervoid: The reason is that the keys are hashed before being stored in the database, so we don't have access to the original value after creating them. We could disable that technically, but i really think we shouldn't. The docs have this to say about it:
```
⚠️ Security Warning: It's strongly recommended to not disable hashing. 
Storing API keys in plaintext makes them vulnerable to database breaches, 
potentially exposing all your users' API keys.
```

enter.pollinations.ai/src/client/routeTree.gen.ts
eulervoid: Ok, i'll remove the routes for now. The router is currently only doing the routing on the frontend, so it is an SPA. We don't have many routes, so its not strictly necessary, but it brings other features like data loading and state management that are really worth the dependency in my opinion.

enter.pollinations.ai/src/registry.ts
eulervoid: Yep, makes sense, i'll remove it for this PR.

flake.nix
eulervoid: Ok, i'll remove pipelight for now.

enter.pollinations.ai/src/style.css
eulervoid: I think there was some issue with the bundling process that led me to put it there, but i don't remember exactly.. i'll try to move it into client.

eulervoid:
Ok, i applied the changes that we discussed, could you have another look @voodoohop ?

enter.pollinations.ai/src/style.css
voodoohop: no problem if its because of that i just thought it may be a leftover file or something

enter.pollinations.ai/README.md
voodoohop: i dont think we need the metrics thing at all. but i thought id seen related code in the schema too, i think we can remove that

voodoohop:
@eulervoid good to merge after the adjustments. i added some comments

PR #3804, opened 2025-08-21
Add moderation and cache fields to event tracking

ElliotEtag:
Extended the text_events datasource schema and event payload to include moderation severity, protected material detection, and cache details. Moderation telemetry is now sent inline with the main event, removing the separate moderation request logic.


ElliotEtag:
@claude

PR #4019, opened 2025-09-11
Fix `.trim()` crash for array-style `message.content` in `messageSani…

MeatyChunks:
Note that this needs to be tested as I was sleuthing with GPT for the fix, the vision example from apidocs works well to test.

This PR fixes a TypeError that occurs when `message.content` is an array of structured objects instead of a string. The crash manifests as:

```
TypeError: message.content.trim is not a function
```

It happens because `messageSanitizer.js` currently assumes `message.content` is always a string:

```js
if (message.role === 'user' && (!message.content || message.content.trim() === '')) {
  replacedCount++;
  return { ...message, content: 'Please provide a response.' };
}
```

The OpenAI-style schema allows `message.content` to be either:

* A plain string
* An array of objects, e.g. `[{type: "text", text: "..."}, {type: "image_url", image_url: {...}}]`

**Changes made:**

* Added type checking for `message.content`.
* If it’s an array, concatenate all text parts and trim them.
* Only replace content with placeholder if the resulting text is empty.

**Example:**

```js
if (typeof content === "string") {
  isEmpty = content.trim() === "";
} else if (Array.isArray(content)) {
  const concatenated = content
    .map(part => part.type === "text" && typeof part.text === "string" ? part.text : "")
    .join(" ")
    .trim();
  isEmpty = concatenated === "";
}
```

**Benefits:**

* Prevents `.trim()` crashes on structured messages.
* Preserves existing behavior for string-only messages.
* Maintains support for OpenAI-style image+text inputs.

**Steps to Test:**

1. Send a POST to `/openai` with `messages[].content` containing both `{"type": "text"}` and `{"type": "image_url"}`.
2. Verify no `TypeError` occurs.
3. Verify empty text content is still replaced with the placeholder string.

<details>
<summary>Python vision example</summary>

```python
import requests
import base64
import json

url = "https://text.pollinations.ai/openai"
headers = {"Content-Type": "application/json"}

# Helper function to encode local image to base64
def encode_image_base64(image_path):
    try:
        with open(image_path, "rb") as image_file:
            return base64.b64encode(image_file.read()).decode('utf-8')
    except FileNotFoundError:
        print(f"Error: Image file not found at {image_path}")
        return None


def analyze_local_image(image_path, question="What's in this image?"):
    base64_image = encode_image_base64(image_path)
    if not base64_image:
        return None

    # Determine image format (simple check by extension)
    image_format = image_path.split('.')[-1].lower()
    if image_format not in ['jpeg', 'jpg', 'png', 'gif', 'webp']:
         print(f"Warning: Potentially unsupported image format '{image_format}'. Assuming jpeg.")
         image_format = 'jpeg' # Default or make more robust for unknown formats

    payload = {
        "model": "openai", # Ensure this model supports vision
        "messages": [
            {
                "role": "user",
                "content": [
                    {"type": "text", "text": question},
                    {
                        "type": "image_url",
                        "image_url": {
                           "url": f"data:image/{image_format};base64,{base64_image}"
                        }
                    }
                ]
            }
        ],
        "max_tokens": 500
    }
    try:
        response = requests.post(url, headers=headers, json=payload)
        response.raise_for_status()
        return response.json()
    except requests.exceptions.RequestException as e:
        print(f"Error analyzing local image: {e}")
        # if response is not None: print(response.text) # Show error from API
        return None



result_local = analyze_local_image('path/to/your/image.jpg', question="Describe the main subject.")
if result_local:
    print("Local Image Analysis:", result_local['choices'][0]['message']['content'])
```
</details>

### DIFF
--- a/text.pollinations.ai/utils/messageSanitizer.js
+++ b/text.pollinations.ai/utils/messageSanitizer.js
@@ -23,7 +23,6 @@ export function sanitizeMessagesWithPlaceholder(messages, modelConfig, originalM
     return { messages, replacedCount, bedrockAdjusted };
   }
 
-  // Step 1: Ensure user messages have non-empty content to prevent provider errors
   let result = messages.map((msg) => {
     if (!msg || msg.role !== "user") return msg;
     const m = { ...msg };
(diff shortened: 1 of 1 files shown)


2013224691:
Hello I got the same error but using the curl in php: {"error":"message.content.trim is not a function","status":500} . Can you help me to fix it I use the openai format api in php

voodoohop:
sorry for taking so long to merge this

Issue #4020, opened 2025-09-11
Track cost for image generation


eulervoid:
How to track cost for image generation is still an open question for me:
- How do we pass the cost data from the image service to auth?
- Do we use tokens, compute seconds or some other unit?

I think we should just decide relatively quickly on tokens/seconds and then just do it like this. We can still easily change it in the future, as the system is set up to support different kinds of units. 

As for how to pass the data: 
- Similar to the text API maybe? So have the actual image service only supports a POST API that returns json, maybe with a link to the image, and then we map the simpler APIs, like the GET `/:prompt` to this via the proxy. 
- BUT: This might not be the best way to do it in the short term as it requires changes to the service and potentially the cache, so the simple alternative would be to pass it as headers.

What do you think @voodoohop ? I'm totally open for suggestions as well, just thinking out loud here basically.

voodoohop:
My suggestion is to use the same system as we use for TXT. The difference is that we calculate a cost per image based on the average throughput of images per week and the cost of the GPUs per week.

<img width="856" height="430" alt="Image" src="https://github.com/user-attachments/assets/f2bf551b-dbce-43c7-922a-b2a26cbe3168" />

The number of images generated per month is not changing drastically. If we use the number of a month with slightly more usage, the worst that can happen is we slightly overestimate the cost.

---

### Current Data

* August: **8.5M images per week (avg)**
* Last 2 weeks: **18M images total**
* GPU cluster cost per week: **\$789.60**

---

### Cost per Image

* August: `0.0103¢`
* September: `0.0088¢`
* October (projected, +10% MoM image growth): `0.0079¢`

*At the current utilization, cost per image is decreasing \~10% per month (\~2.2% per week).*

---

### Formula

$$
\text{WeeklyGrowth} = 2.2\%
$$

$$
\text{EstimatedImages} = \text{ImagesLastWeek} \times (1 + \text{WeeklyGrowth})
$$

$$
\text{CostPerImage} = \frac{\text{GPUCostWeekly}}{\text{EstimatedImages}}
$$

1. Each week: take the number of images generated **last week**.
2. Apply **+2.2% Week-over-Week (WoW)** growth (≈10% MoM).
3. Compute cost per image with the formula above.

---

### What unit do we use?

* **Unit = `IMAGE`** (1 generated image = 1 billable unit).
* Rationale: simple, matches how we already think about throughput, and integrates smoothly with Grafana and JSON `cost_per_unit`.

---

[...]

Issue #3885, opened 2025-08-28
Cache polar requests (balance, customer state) in KV


eulervoid:
We need to request user balance from polar for each request to the image/text generation apis. To make this fast and efficient, and not exceed polars rate limits, we should cache these in Cloudflare KV.

eulervoid:
Implemented by #4039

PR #4039, opened 2025-09-12
Pollen MVP v0.2

eulervoid:
- **Add logging and cost/price calculation**
- **Add models/services to registry**
- **Add tests for model/service registry and cost/price calculations**
- **Add fixtures for signIn and creating an api key, add tests for calling all defined models**
- **Add test to ensure only free services are available without API key**
- **Add tinybird cli to flake.nix**

Still a draft for now.

### DIFF
--- a/enter.pollinations.ai/vitest.config.ts
+++ b/enter.pollinations.ai/vitest.config.ts
@@ -33,8 +33,9 @@ export default defineWorkersConfig(async () => {
                         enabled: true,
                         include: [
                             "@polar-sh/sdk",
-                            "@polar-sh/better-auth",
                             "better-auth",
+                            "kysely",
+                            "drizzle-orm",
                         ],
                     },
                 },


eulervoid:
Ok, i think we can merge this now and i will create a new branch/pr for the next features.

voodoohop:
@eulervoid 

What is the next branch about? Let's base branches around specific features next week. I want to contribute PRs to the new pollen/service system. How about we do a small task swap: I take a pollen feature, you do the TypeScript + Hono conversion of the text service?

The text service is cleaned up and ready for conversion. Understanding the text service better will likely help you make decisions in the pollen system too. I need to understand the enter, pollen, and databases anyway properly. Maybe we can then throw out the auth system of the text service together at the end of the week.

eulervoid:
I would do #3885 and #4020 next, each in their own PR.

Regarding the task swap: sure, makes sense. I completely agree that we should each try to understand as much of the codebase as possible.

voodoohop:
@eulervoid cool shall we merge this one? another way of moving forward would be for me to convert the text service to typescript + hono using your image.pollinations code as reference and then integrating it with enter.pollinations.ai. what would you prefer?

PR #4135, opened 2025-09-22
Only allow paid generations with enough balance

eulervoid:
Adds proper authorization that only allows requests to paid models if the user has sufficient balance. Requests to polar for getting the balance are cached in Cloudflare KV, in order to improve performance and avoid hitting polars rate limit. Tracking was also improved, adding cache information that is also used for billing (cached responses are not billed).


eulervoid:
I fixed the conflicts and and an issue with how cache-hits and errors where handled, i think this is ready to be merged now @voodoohop

PR #4151, opened 2025-09-23
Update image cost and pricing logic

eulervoid:
- Handle image generations as 1 token per generation
- Add DPT unit to conveniently enter price per image
- Add `cost_type` to tracking events to record if the cost is a fixed weekly cost or per generation


eulervoid:
Closing this because I created a new version that branches off from master, which makes it much easier to review the actual changes and avoids conflicts.

PR #4167, opened 2025-09-24
Track moderation results for text generations

eulervoid:
Adds moderation results from OpenAI responses to tinybird events. If we want to track moderation results for image requests, we should pass them as headers.


voodoohop:
@claude can you review

PR #4181, opened 2025-09-25
Update PromptEnhancer.ts and add support for image-based enhancement 

Itachi-1824:


redac1ed:
@Itachi-1824 Suggestion:

Instead of using only one model, we should use multiple (ex- gemini) as fallback if, in any case, the model fails to generate a prompt. This would reduce the chances of just passing the original prompt in case any model fails.

voodoohop:
@claude can you review?

voodoohop:
@Itachi-1824 prompt enhancing is currently disabled but i will re-enable it asap and then merge your branch. we were having some issues with rate limiting. although maybe what we could do is set enhance to false by default.

Itachi-1824:
Sure @voodoohop 
I've also updated the ParameterProcessor.js to strip unsupported params from GPT-5 models!

PR #4184, opened 2025-09-25
fix(telemetry): Guard cost resolution to prevent error events from being dropped

ElliotEtag:
## Problem

Since commit `77a8197e530e6d0816b8ce009df47c82173e7786`, no `response_status = 'error'` events are appearing in the Tinybird `text_events` datasource. All logs now show as `success` even when errors occur.

## Root Cause

The recent commit changed `observability/costCalculator.js` to throw an error when `resolveCost()` is called without a model name:

```javascript
export function resolveCost(responseModel) {
    if (!responseModel) {
        throw new Error('No model name provided for cost resolution');
    }
    // ...
}
```

In error scenarios, `genericOpenAIClient.js` calls `sendTinybirdEvent()` without providing `modelUsed`, causing:
1. `tinybirdTracker.js` calls `resolveCost(undefined)`
2. `resolveCost()` throws an error
3. The entire telemetry event is dropped before reaching Tinybird
4. No error events are recorded

## Solution

Guard the cost resolution in `tinybirdTracker.js` to:
- Skip cost resolution when `modelUsed` is falsy (common in error paths)
- Catch any cost resolution errors and log them
- Continue sending telemetry events even if cost lookup fails
- Default token price fields to 0 when cost data is unavailable

## Changes

- **Modified**: `observability/tinybirdTracker.js`
  - Wrapped `resolveCost(modelUsed)` in try/catch
  - Added null check before calling `resolveCost()`
  - Added error logging for cost resolution failures
  - Ensured telemetry events are still sent when cost lookup fails

## Testing

After this fix:
- Error events should appear again in Tinybird with `response_status = 'error'`
- Cost data will still be captured when available
- No telemetry events should be silently dropped due to cost resolution failures

## Verification

To verify the fix works, check Tinybird for error events:

```sql
SELECT count() FROM text_events 
WHERE response_status = 'error' 
AND timestamp >= now() - interval 1 hour
```

Fixes the regression where error telemetry was silently dropped, restoring observability into API failures.

### DIFF
--- a/text.pollinations.ai/observability/tinybirdTracker.js
+++ b/text.pollinations.ai/observability/tinybirdTracker.js
@@ -33,15 +33,19 @@ export async function sendTinybirdEvent(eventData) {
     try {
         // Extract model and cost information
         const modelUsed = eventData.modelUsed ?? null;
-        // Cost resolution must be tolerant: in error paths we may not have a model
+        const isErrorEvent = eventData.status === "error";
+        
+        // Skip cost resolution for error events - no token usage data available
         let cost = null;
-        try {
-            if (modelUsed) {
-                cost = resolveCost(modelUsed);
+        if (!isErrorEvent) {
+            try {
+                if (modelUsed) {
+                    cost = resolveCost(modelUsed);
+                }
+            } catch (e) {
+                errorLog(`Cost resolution failed for modelUsed='${modelUsed || ''}': ${e.message}`);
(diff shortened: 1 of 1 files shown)


voodoohop:
@ElliotEtag i dont think we can resolve cost for errors. we dont get token counts back. an error should just have cost 0.

voodoohop:
@claude can you review? is error.status guaranteed to be set?

PR #4208, opened 2025-09-26
Fix signup when a polar customer already exists

eulervoid:
When signing up, a polar customer for each user gets created. This PR makes it so that if a customer with this email address already exists, it is linked to the new account. This is mostly relevant for development, where the staging environment uses the same sandbox as the dev environment, but should also make the production version more robuts.


voodoohop:
Great. Loving the fine-grained PRs <3

PR #4276, opened 2025-10-01
Enhance the information about the image models to be shown to the users

Circuit-Overtime:
This pull request enhances the API response for listing available models by including additional metadata for each model. Instead of returning just the model names, the endpoint now provides details such as the model's tier and maximum side length. The changes also ensure that all models in the configuration explicitly specify their tier.

API response improvements:

* The `/models` endpoint now returns an array of objects with `name`, `tier`, and `maxSideLength` for each model, instead of just a list of model names.

### DIFF
--- a/image.pollinations.ai/src/index.ts
+++ b/image.pollinations.ai/src/index.ts
@@ -586,9 +586,23 @@ const server = http.createServer((req, res) => {
             Pragma: "no-cache",
             Expires: "0",
         });
+        
+        res.end(JSON.stringify(Object.keys(MODELS)));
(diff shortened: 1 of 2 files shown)


voodoohop:
i think that will break compatibility with existing apps
i think better to create a separate endpoint maybe call it /modelDetails 
we can make breaking changes when we roll out pollen
to the public

Circuit-Overtime:
@voodoohop I have fixed and created a /about route

voodoohop:
@claude can you review

PR #4283, opened 2025-10-02
Simplify registry to DPT-only pricing

voodoohop:
- Remove unit field from pricing definitions (rates are now just numbers in DPT)
- Consolidate ZERO_COST and ZERO_PRICE into single ZERO_PRICE constant
- Add price-helpers.ts with fromDPMT() converter and costAsPrice() helper
- Use costAsPrice() for services that charge provider cost directly
- Reduce boilerplate by extracting .cost internally in costAsPrice()
- Match minimal commenting style of codebase


voodoohop:
@claude can you review

PR #4305, opened 2025-10-02
Add api docs

eulervoid:
Adds Scalar API docs at `https://enter.pollinations.ai/docs`
- Only show generation endpoints when not in development mode
- Add request validation and response schemas
- Add error response schemas
- Add `/api/generate/image/models` endpoint

This branches off from the improve-ui PR, which is why the diff is messy. Sorry about that.. i can rebase it on master if we want to get a clean diff.


voodoohop:
@claude can you review?

PR #4400, opened 2025-10-07
feat(enter): Add tier field to user table

voodoohop:
- Add `tier` column to user schema with default `"seed"` value
- Add tier to better-auth user additional fields
- Update tracking middleware to extract tier from user object
- Add migration script with tier index for performance

Tier is now stored as user metadata and automatically tracked in all generation events. Usage/enforcement will be determined later.


voodoohop:
@claude can you review?

PR #4475, opened 2025-10-10
Simplify registry: Remove factory pattern, test with real functions

voodoohop:
Removes factory pattern and unused code from shared registry. Tests now use production functions directly instead of mocks.

**Changes:**
- Removed `createRegistry()` factory
- Removed `createTestRegistry()`
- Tests use direct imports: `calculateCost()`, `resolveServiceId()`, etc.
- Added YAGNI principle to development guidelines

**Benefits:**
- Simpler codebase
- Tests validate real pricing/behavior
- Direct function exports (tree-shakeable)
- No mock/reality divergence


voodoohop:
@claude can you review

PR #4550, opened 2025-10-14
🐛 Fix Tinybird model name logging to use actual models instead of URL…

ElliotEtag:
… parameters

**Problem:**
Tinybird database was being flooded with invalid/non-existent model names like:
- Old removed models: analog-diffusion, midjourney-v4, dreamlike-photoreal-2.0
- Non-existent models: dalle-3, gpt-4, gpt-4o (text models)
- Error messages: "error loading models"
- Typos: fluxxx, midijourney
- Malformed: .nanobanana, ImageModel.FLUX

This happened because telemetry was blindly logging whatever users put in the ?model= URL parameter, not the actual model that processed the request.

**Solution:**
1. **Cloudflare Cache (tinybird.ts):**
   - Now reads x-model-used response header (actual model from backend)
   - Falls back to URL parameter only if header missing
   - Prevents invalid model names in cache hit telemetry

2. **Main Image Service (index.ts):**
   - Success telemetry uses trackingData.actualModel
   - Only falls back to requested model if tracking data unavailable
   - Error telemetry unchanged (no tracking data in error cases)

3. **Bonus: Cache Logging (exact-cache.ts):**
   - Changed cache hit/miss from console.log → console.debug
   - Removed verbose cache key logging
   - Reduces production log noise

**Result:**
Tinybird will now only log valid model names that actually exist and processed requests: flux, turbo, kontext, nanobanana, seedream, gptimage

**Files Changed:**
- cloudflare-cache/src/middleware/tinybird.ts: Use x-model-used header
- image.pollinations.ai/src/index.ts: Use trackingData.actualModel
- cloudflare-cache/src/middleware/exact-cache.ts: Reduce log verbosity

### DIFF
--- a/image.pollinations.ai/src/index.ts
+++ b/image.pollinations.ai/src/index.ts
@@ -301,11 +301,9 @@ const imageGen = async ({
         // Send telemetry to Tinybird
         const endTime = new Date();
         const duration = endTime.getTime() - startTime;
-        // Use actual model from tracking data, fallback to requested model
-        const actualModel = maturity.trackingData?.actualModel || safeParams.model || "unknown";
         sendImageTelemetry({
             requestId,
-            model: actualModel,
+            model: safeParams.model || "unknown",
             duration,
             status: "success",
             authResult,


ElliotEtag:
@claude be critical

ElliotEtag:
@claude what o you think now? changed the appraoch

voodoohop:
@ElliotEtag is this still relevant? id close it because it seems to be a bit outdated

ElliotEtag:
We can close this is resolved differently

PR #4759, opened 2025-10-24
feat(enter): Add flag to enable/disable anonymous usage of generation endpoints

eulervoid:
Reject anonymous requests in staging/production, can be configured via `ALLOW_ANONYMOUS_REQUESTS` env variable.


voodoohop:
@eulervoid just made anonymous false for local dev too by default

@claude i added some small changes. can you also review?

PR #4761, opened 2025-10-24
feat(enter): Separate session and API key authentication middlewares

voodoohop:
**Authentication Middleware (`src/middleware/authenticate.ts`)**
- Split unified `authenticate()` into two specialized middlewares:
  - `authenticateSession`: Session cookies only (dashboard routes)
  - `authenticateAPI`: Bearer tokens only (API routes)
- Removed fallback logic between auth methods
- Renamed `requireActiveSession()` → `requireAuth()`
- Updated error messages to be context-specific

**Route Updates**
- `polar.ts`: Use `authenticateSession` 
- `proxy.ts`: Use `authenticateAPI`
- All `requireActiveSession()` calls → `requireAuth()`

## Notes
- No breaking changes - internal refactoring only
- Builds on the API key auth introduced in master (commit 5be8da60d)

### DIFF
--- a/enter.pollinations.ai/src/middleware/authenticate.ts
+++ b/enter.pollinations.ai/src/middleware/authenticate.ts
@@ -53,9 +53,15 @@ export const authenticateSession = createMiddleware<AuthEnv>(async (c, next) => 
 
 /**
  * API key authentication for API routes.
- * Only checks Bearer token (no session cookies).
+ * Checks Bearer token. Can be chained with authenticateSession for fallback.
  */
 export const authenticateAPI = createMiddleware<AuthEnv>(async (c, next) => {
+    // Skip if already authenticated (e.g., by session middleware)
+    const existingAuth = c.get("auth");
+    if (existingAuth?.user) {
+        return await next();
+    }
+
     const client = createAuth(c.env);
     const authHeader = c.req.header("authorization");
     const apiKey = extractApiKey(authHeader);
@@ -72,7 +78,7 @@ export const authenticateAPI = createMiddleware<AuthEnv>(async (c, next) => {
     const requireAuth = (message?: string) => {
         if (!user) {
             throw new HTTPException(401, {
-                message: message || "API key required. Provide a valid Bearer token.",
+                message: message || "Authentication required. Sign in or provide a valid API key.",
             });
         }
         return { user, session: undefined };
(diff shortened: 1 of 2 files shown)


eulervoid (commented):
I'm not sure why this is needed, whats the harm in allowing cookie auth in the api routes? Also getting the user is done automatically by better-auth which also creates a session object, i think doing this manually adds unnecessary complexity. Unless there is a good reason?

voodoohop:
> I'm not sure why this is needed, whats the harm in allowing cookie auth in the api routes? Also getting the user is done automatically by better-auth which also creates a session object, i think doing this manually adds unnecessary complexity. Unless there is a good reason?

I wasn't able to authenticate via API key at all before. From my investigation it seemed like there was extra logic needed to be able to do API key authentication. So I created a middleware for it.

Enabling or disabling session-based authentication for API routes is quite easy just re-adding the middleware to that route. That wasn't supposed to be the point of this code.

I tried to get API-key based authentication to work before the commit eaeaf11567960911e9365aeea343da0744fbd237 - in which we started to the change to Bearer authentication but it seemed like that logic wasn't complete yet. It it seemed like it was trying to use session authentication which was failing when an API key was supplied.

voodoohop:
@eulervoid I re-enabled session auth for API routes.

I also created this branch: https://github.com/pollinations/pollinations/tree/demo/api-key-issue-before-pr-4761. 

It has a new `/enter.pollinations.ai/test-api-key-before-pr.sh` script which shows my problems to get `x-api-key` working before any changes made for *Bearer* authentication.

Try requesting something from `openai-large` for example with an API key. I don't think it was working. If there is a simpler way, fine to discard this PR.

eulervoid:
We don't need this middleware, the issue is most likely that when you extract the bearer token, you are basically replacing the function that extracts the x-api-key header, so if we handle both in the custom extraction function everything should work as expected without an extra middle ware and the added complexity. I can make a small PR for this today, it should only be a couple of lines of code.

voodoohop:
@eulervoid can you do this before doing other things? a few tasks of elliot and mine are depending on it. 

i think it would be best to start with all the bearer token stuff reverted? because when I created the first commit related to this I potentially had the wrong assumptions

voodoohop:
Was included in https://github.com/pollinations/pollinations/pull/4783

@eulervoid it seems to be working we can simplify more later on i would say, unless there is something important that was missed

PR #4793, opened 2025-10-26
fix(enter): Fix x-api-key and authorization header, add tests and fix discovered issues, improve logging

eulervoid:
- Improves on the authentication by keeping it as a single middleware with config options
- Fixes authentication fixture that lets tests create and use api-keys and session cookies
- Adds tests that go through each model and check that the availablility is as expected
- Adds tests for the polar endpoints that confirm correct access rights
- Update tests to handle different auth mechanisms (x-api-key header, authorization header, session cookie)

### DIFF
--- a/enter.pollinations.ai/test/models.test.ts
+++ b/enter.pollinations.ai/test/models.test.ts
@@ -48,9 +48,9 @@ describe.for([true, false])(
     },
 );
 
-// Send a request to each text model, using the x-api-key header
+// Send a request to each text model, using the Authorization Bearer header
 test.for(getTextServices())(
-    "%s should respond with 200 when x-api-key header",
+    "%s should respond with 200 when using API key via Bearer token",
     { timeout: 30000 },
     async (serviceId, { apiKey }) => {
         const response = await SELF.fetch(
@@ -59,7 +59,7 @@ test.for(getTextServices())(
                 method: "POST",
                 headers: {
                     "content-type": "application/json",
-                    "x-api-key": apiKey,
+                    "authorization": `Bearer ${apiKey}`,
                     "referer": env.TESTING_REFERRER,
                 },
                 body: JSON.stringify({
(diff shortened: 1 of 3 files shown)


voodoohop:
@eulervoid 

i think we should remove x-api-key

## 🔴 Security Issue: Unsafe Bearer Token Extraction

**File**: [src/middleware/auth.ts](cci:7://file:///Users/thomash/Documents/GitHub/pollinations/enter.pollinations.ai/src/middleware/auth.ts:0:0-0:0) line 50

**Current Code**:
```typescript
function extractApiKey(headers: Headers): string | null {
    if (headers.has("x-api-key")) {
        return headers.get("x-api-key");
    }
    if (headers.has("authorization")) {
        return headers.get("authorization")!.split(" ")[1] || null;a
    }
    return null;
}
```

**Problems**:
1. **No "Bearer" prefix validation** - accepts ANY authorization header format
2. **No bounds checking** - crashes if header has no space
3. **Non-null assertion** - unsafe after `has()` check

**Attack Vectors**:
```bash
# Accepted as valid (wrong):
Authorization: NotBearer some-random-string

# Causes crash:
Authorization: NoSpace
```

**Priority**: Critical - must fix before merge

voodoohop:
@claude can you review?

enter.pollinations.ai/test/models.test.ts
```diff
@@ -48,9 +48,9 @@ describe.for([true, false])(
     },
 );
 
-// Send a request to each text model, using the x-api-key header
+// Send a request to each text model, using the Authorization Bearer header
 test.for(getTextServices())(
-    "%s should respond with 200 when x-api-key header",
+    "%s should respond with 200 when using API key via Bearer token",
```
eulervoid, line 53: This test can be deleted completely as below there is another one checking bearer auth.

PR #4806, opened 2025-10-26
Add simplified GET text endpoint

eulervoid:
- Add `/api/generate/text/:prompt`
- Add moderation data as headers
- Read usage data from headers

I'm currently adding the moderation headers in the text endpoint as a shortcut, we should add them in the upstream services in a future pr.

### DIFF
--- a/enter.pollinations.ai/src/db/schema/event.ts
+++ b/enter.pollinations.ai/src/db/schema/event.ts
@@ -1,6 +1,7 @@
 import { PriceDefinition, TokenUsage } from "@shared/registry/registry.ts";
 import { integer, real, sqliteTable, text } from "drizzle-orm/sqlite-core";
-import type { ContentFilterResult, OpenAIResponse } from "@/usage.ts";
+import type { CreateChatCompletionResponse } from "@/schemas/openai";
+import { removeUnset } from "@/util.ts";
 
 const eventTypeValues = ["generate.text", "generate.image"] as const;
 export type EventType = (typeof eventTypeValues)[number];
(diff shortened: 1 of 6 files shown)


voodoohop:
@eulervoid it seems like it's mixed with a formatter that changed most files. I think it's better to do only the endpoint change here and do the formatting from the merged master

eulervoid:
I cherry picked the commit on master to get a clean PR. 

That it shows so many changes happens because i don't want to wait for the last PR to be merged to continue working, but i also don't want to start from an outdated master branch, so i branch off from my lastest PR. I would hope that git can handle this but it seems it can't. But its usually not a problem to just apply the commits to a new branch off of master once the previous PR is merged.

The formatter changes are just in `usage-types.ts`. But it also contains a real change though, so i think it should stay in this commit. We should all use biome with format on save, then this will not happen anymore ;).

eulervoid:
> ### 👷 Deploy Preview for _pollinations-react-hooks-docs_ processing.
> Name 	Link
> 🔨 Latest commit 	[403ad23](https://github.com/pollinations/pollinations/commit/403ad237e11b6fdabc9d7635765d9d75f435aa7e)
> 🔍 Latest deploy log 	https://app.netlify.com/projects/pollinations-react-hooks-docs/deploys/68feaff09e006a00085c3198

Do we need these comments, could we turn that off? It's a bit too much noise for my taste. There are already the netlify checks as well..

voodoohop:
@eulervoid i think just editing the netlify workflow. make a mini-pr if you want for that. 

i did migrate all the dns stuff from netlify to cloudflare. and it only uses two small functions. we want to move netlify to cloudflare pages asap. 

maybe makes more sense doing that and disabling the netlify bot alltogether

PR #4821, opened 2025-10-27
feat: support token query parameter for GET requests

### LINKED ISSUE #4820: Support token= parameter in enter.pollinations.ai GET requests
## Problem
No way to pass API token directly in URL. Especially useful for the frontend

## Proposal
Support `token=` query parameter for authentication in GET requests:
- `https://enter.pollinations.ai/api/generate/text/hello?token=sk_xxx`
- `https://enter.pollinations.ai/api/generate/image/cat?token=sk_xxx`

## Implementation
- Extract token from query params in proxy route
- Pass to backend services via existing auth headers

voodoohop:
Fixes #4820

- Add  query parameter support for GET requests
- Authorization header takes precedence when both present
- Token parameter filtered before proxying to backend
- Added tests for valid and invalid token scenarios

**Usage:**
```
https://enter.pollinations.ai/api/generate/text/hello?token=sk_xxx
https://enter.pollinations.ai/api/generate/image/cat?token=sk_xxx
```


voodoohop:
@eulervoid i think maybe just `key` is nice. 

I just went with the simple solution instead of checking if it is specifically a front-end key. Was in two minds whether this is something the users are responsible for or not. Like maybe someone created a private key because they are doing an app that is running in their intranet. Then  we're not being the nanny telling them they can't use it that way. When I'm in doubt I usually go for the approach that involves less logic. But either way is good for me.

the reason i'm using token is that it's how it was before. I'm not sure how much wer are trying to maintain compatibility with the old API but maybe we can just go for what we like most. if we really want we can later make image. and text. services maintain backwards compatibility

eulervoid:
In this case i would go with being a nanny, in the end we will have the trouble if people use it in the wrong way and rack up a huge bill because the key was stolen. I would try to enforce good practices wherever we can, its a small thing now but would be a huge hassle if something happens down the line.

voodoohop:
@eulervoid i'm going to override your suggestion for now and only change the naming. feel free to propose a PR with the extra check. my intuition is simple is good. we don't need to write extra docs explaining that for the token only that key works. in the end there are already different ways in which the user is taking responsability for their pollen. the worst thing is they spend the pollen that they charged. it's not like the cloud providers where you pay afterwards.

voodoohop:
@claude - can you review? whats your opinion on the token discussion? be concise

voodoohop:
@claude ok whatever. so go ahead and implement your changes. im just like whatever, the dna of pollinations is open for hackers and i dont know i feel like being flexible is better than being beaurocratic. but i think im outvoted here and that im trying to sway the discussion is not really fair. so go ahead and implement it

voodoohop:
@claude can you remove temporary test files that were created? if a test is needed it should be part of the testing framework. add tests if it make sense but follow the exact testing convention that was used. dont create any extra files.

voodoohop:
@claude fyi im going to ask deepseek next time. i think you're just too square.

PR #4835, opened 2025-10-28
Fix GitHub Action failure: Add missing `actions: write` permission for debounce step

CloudCompile:
This PR fixes the workflow failure caused by missing permissions during the debounce step.

### ✅ Problem

The workflow `.github/workflows/auto-condense.yml` was failing with this error:

```
Resource not accessible by integration
```

This was thrown when the debounce step attempted to cancel previous runs using the GitHub API, but the workflow did not have sufficient permissions.

### 🔧 Fix

Added the required `actions: write` and `contents: read` permissions to allow workflow run cancellation.

#### Updated permissions block

```yaml
permissions:
  actions: write
  contents: read
  pull-requests: write
  issues: write
```

### ✅ Result

* Debounce step now works
* No more `403 Resource not accessible by integration` errors
* Workflow executes normally

---

### DIFF
--- a/.github/workflows/condense-descriptions.yml
+++ b/.github/workflows/condense-descriptions.yml
@@ -14,15 +14,17 @@ jobs:
   condense:
     runs-on: ubuntu-latest
     permissions:
+      actions: write
+      contents: read
       pull-requests: write
       issues: write
-    
+
     steps:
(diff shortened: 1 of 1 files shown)


CloudCompile:
That was fast...

PR #4876, opened 2025-10-30
fix: tier system not refilling daily - update to V2 products

ElliotEtag:
## Changes

- Update Polar product IDs to V2 across all environments (dev, staging, production)
- Simplify frontend tier config - only stores badge colors now
- Use Polar API as single source of truth for product names and daily pollen amounts
- Add detailed logging for tier detection, product info, and DB/Polar mismatches
- Improve UI - make tier name larger and bolder (text-3xl font-bold)
- Remove duplicate subscription check in backend
- Add `/upgrade` endpoint for legacy to V2 subscription migration

## Key Improvements

**Product IDs Updated:**
- All three tiers (Seed, Flower, Nectar) now use new V2 product IDs
- Legacy product IDs preserved in code for detection only

**Cleaner Architecture:**
- Frontend no longer hardcodes tier metadata
- Polar product data flows directly to UI
- TIER_CONFIG reduced from 3 properties to 1 (badge colors only)

**Better Observability:**
- Minimal, readable logs for tier detection
- Clear warnings when DB tier doesn't match Polar tier
- Product ID, tier, and daily pollen all logged

## Testing

Tested with both legacy and V2 subscriptions - system correctly:
- Recognizes V2 subscriptions
- Detects legacy subscriptions  
- Displays correct product names and pollen amounts from Polar
- Shows mismatch warnings when DB tier differs from Polar tier

### DIFF
--- a/enter.pollinations.ai/src/routes/tiers.ts
+++ b/enter.pollinations.ai/src/routes/tiers.ts
@@ -13,13 +13,6 @@ type ActivatableTier = "seed" | "flower" | "nectar";
 // Central tier definition
 const TIERS: readonly ActivatableTier[] = ["seed", "flower", "nectar"] as const;
 
-// Legacy product IDs (old subscriptions before migration)
-const LEGACY_PRODUCT_IDS: Record<string, ActivatableTier> = {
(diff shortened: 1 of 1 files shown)


ElliotEtag:
@claude review please

PR #4941, opened 2025-11-03
Add taskfile and process-compose config for better local development across services

eulervoid:
Adds repo-scoped tasks for easier development across services.
- `task install` Installs dependencies for enter, text and image services
- `task dev` Runs dev servers for enter, text and image services with automatic restart on code changes

### DIFF
--- a/enter.pollinations.ai/wrangler.toml
+++ b/enter.pollinations.ai/wrangler.toml
@@ -14,6 +14,7 @@ directory = "dist/client"
 not_found_handling = "single-page-application"
 run_worker_first = [ "/api/*" ]
 
+
 [[d1_databases]]
 binding = "DB"
 database_name = "development-pollinations-enter-db"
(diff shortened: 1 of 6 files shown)


voodoohop (changes requested):
@eulervoid 

i just ran it locally and I'm a bit unsure. It involved installing two extra binary dependencies and two new config files. 

it could be done with one well-structured bash script, a Makefile or  a package.json in the root (optionally with a lightweight node-based executer like `concurrently` or `npm-run-all`)

i'm always thinking of the overhead needed when we add extra tooling. We'd have to document where to download and how to install the two new deps `process-compose` and the task runner.

eulervoid:
My thinking in using `task` as the top level task runner was that we might have services other than node, or want to run docker containers, etc. `make` is definitely an option there, i just find the syntax super oldschool and error prone. Shell scripts have the same drawbacks as make, anything a bit more complicated gets difficult to maintain in my view. 

If we use nix, installing the dependencies is done by the dev shell, so i personally don't think its that much of an issue, but i understand if it's not for you.

Maybe we throw out task, replace it with node, but keep process-compose? I find it very handy how it lets me see the individual logs, restart individual services etc. Or maybe the node runners have similar functions?

voodoohop:
@eulervoid I prefer a solution that doesnt require an additional install unless it's light and node-based as in it uses the same tooling we use everywhere. i think there's probably an easy way to set up process-compose for your local environment which is outside of version control. 

even pm2 feels like overkill to me because it has all kinds of features for longer running processes and sometimes makes it difficult for ai agents to debug and interact with the testing environment. i think grepping or filtering for individual service logs is quite trivial if we set it up nice. concurrently seems quite good for that in terms of features but also open just to bash or make or everything in the scripts part in the root package.json

eulervoid:
Ok, lets close this PR and i will add my process-compose.yml to gitignore.

voodoohop:
@eulervoid I quickly vibecoded the solution with concurrently and a simple package.json

Screenshot 2025-11-04 at 00 14 44

i think it fulfills the purpose at least for my use cases and is a bit more lightweight considering everything is already node.

```json

{
  "name": "pollinations-monorepo",
  "private": true,
  "scripts": {
    "install:all": "npm install --prefix enter.pollinations.ai && npm install --prefix text.pollinations.ai && npm install --prefix image.pollinations.ai",
    "dev": "concurrently --kill-others-on-fail --prefix '[{name}]' --names 'enter,text,image' -c 'auto' 'npm run dev --prefix enter.pollinations.ai' 'npm run dev --prefix text.pollinations.ai' 'npm run dev --prefix image.pollinations.ai'",
    "dev:enter": "npm run dev --prefix enter.pollinations.ai",
    "dev:text": "npm run dev --prefix text.pollinations.ai",
    "dev:image": "npm run dev --prefix image.pollinations.ai"
  },
  "devDependencies": {
    "concurrently": "^9.1.0"
  }
}


```

thats the root package.json now. check if you think its ok. i think concurrently also has ways of dynamically hiding / unhiding specific services

voodoohop:
@claude can you look at how we can start the enter service in a way it is connecting the local running services? create an env for it? override the variaboes?
 respond concisely

ElliotEtag:
@voodoohop should we keep this or close?

voodoohop:
@ElliotEtag close. i merged another pr related to this

Issue #4795, opened 2025-10-26
Add test for streaming responses


eulervoid:
Streaming text responses should be automatically tested, especially making sure that cost/price tracking works as expected as this will probably be a special case.

eulervoid:
It turns out `trailers` are poorly supported in standard web APIs unfortunately, so i think that will not work easily. So i see two options:

1. Keep the service as is, split the stream, parse the events and accumulate the usage in the enter service
2. Send a last event that contains the accumulated usage, maybe even in a json object that matches the headers from text.pollinations.ai

What we choose depends a bit if we are accumulating the usage for another reason already in the text service? If not, i'd keep it a standard openai api and accumulate in the enter service.

What do you think @voodoohop ?

voodoohop:
hey @eulervoid i forgot to answer here. so since trailers look difficult lets just extract the usage from the last event

it isnt a solution for the general case yet but we only have one streaming case so let's keep it simple. maybe all streaming apis we connect to can send the same last event.

The **OpenAI streaming protocol** (for both Chat and Responses APIs) already includes **token usage totals in the final SSE event**, under `event: data` with `data: [DONE]` termination.

* At the very end, you get a final chunk like:

  ```
  data: {"id":"...","object":"chat.completion.chunk","usage":{"prompt_tokens":123,"completion_tokens":456,"total_tokens":579},"choices":[]}
  ```

  followed by

  ```
  data: [DONE]
  ```

eulervoid:
Yeah no worries, i found that as well. I'm working on a PR at the moment that refactors the tracking so that it can handle streams. I'm changing the way its done a bit so that the track middleware provides functions to track the response inside the handler instead of just tracking the outgoing response. This is necessary because e.g. in the case of the `/text` endpoint, where the response does not contain the usage data. I also think it makes it more explicit and flexible in general.

PR #4942, opened 2025-11-03
Test and fix streaming (#4795)

eulervoid:
Enable streaming responses with usage and content filter tracking:
- Refactor tracking middleware to support streaming responses
- Add functions to extract usage and content filter results from response streams
- Add `{ stream_options: { include_usage: true }}` to azure requests to get usage information
- Remove the trailers from text.pollinations.ai responses, as they were causing issues
- Add tests for streaming responses

Includes changes from #4941, because i was using that for development. I can rebase this after #4941 is merged to get a cleaner PR.

### DIFF
--- a/enter.pollinations.ai/package-lock.json
+++ b/enter.pollinations.ai/package-lock.json
@@ -20,6 +20,7 @@
                 "clsx": "^2.1.1",
                 "date-fns": "^4.1.0",
                 "drizzle-orm": "^0.44.7",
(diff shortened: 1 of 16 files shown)


eulervoid:
I did a rebase to remove the other commit from here, but i kept the local config in wrangler.toml because it is relevant for running the tests, without using the local config, they won't pass because they run against the deployed text service.

As stated above, my thinking was to do this once the other PR was merged, but you are right, i won't do it like this anymore.

PR #4950, opened 2025-11-03
Add automated project submission workflow

CloudCompile:
This PR adds a GitHub Action workflow to automate processing of project submissions (APPS issues).

The workflow will:
1. Trigger when issues are opened or labeled with APPS
2. Parse the issue form data to extract project information
3. Categorize projects based on the category mapping
4. Add the project to the correct category file following the AGENTS.md format
5. Run the project list generation script to update the README
6. Create a PR with proper attribution and links back to the original issue

This automation will streamline the project submission process and reduce manual work for maintainers.


CloudCompile:
@voodoohop

voodoohop:
@CloudCompile i think we need an LLM in the loop to write the description etc. also to decide if a project has a valid url and needs merge. that's why i think i suggested to just automate it with claude code because it can do those kinds of checks by itself and edit the files and run the script that updates the readme. there should be a prompt in the repo around how to handle project submissions. 

but also possible to use pollinations api. ideally new things should be built on enter. i can add a token to the github secrets if you need.

https://github.com/pollinations/pollinations/blob/master/AGENTS.md

has some prompting for this for claude.

CloudCompile:
oh mb that must've just slipped my mind I'll fix that sorry

redac1ed:
@voodoohop let's use pollinations (preferably gpt-5) since claude can sometimes be kind of expensive

CloudCompile:
but sometimes the character limit gets to be to big if they have a large description

voodoohop:
@CloudCompile with enter there shouldnt be character limits. i think the claude model on enter is probably the best suited. it's haiku 4.5 so it's not that expensive and PRs are not like a chat app. In comparison rather infrequent so I don't think the costs are a worry

CloudCompile:
Ok! Ill work on that!

CloudCompile:
@voodoohop i fixed it could you check it? there is a secret in there called POLLINATIONS_API_KEY just fyi

PR #4959, opened 2025-11-03
Add error fields to tinybird events and include some more context

eulervoid:
Adds error details and additional context to tinybird events:

New fields:

-  `request_path`  - Requested endpoint path
-  `api_key_id`  - API key id
-  `api_key_type`  - Type of API key (secret/publishable)
-  `resolved_model_requested`  - The resolved model name
-  `free_model_requested`  - Boolean indicating if free model was requested
-  `error_name`  - Error name, e.g. `UpstreamError`
-  `error_response_code`  - String representation of error response code e.g. `BAD_REQUEST`
-  `error_source`  - Source of the error (e.g., upstream hostname)
-  `error_message`  - Error message text
-  `error_stack`  - Stack trace
-  `error_details`  - Additional error details (JSON)

Removed field:

-  `cost_type`

Addresses #4872 and #4703

### DIFF
--- a/enter.pollinations.ai/src/middleware/track.ts
+++ b/enter.pollinations.ai/src/middleware/track.ts
@@ -309,7 +309,10 @@ function createTrackingEvent({
         ...responseTracking.cacheData,
 
         modelRequested: requestTracking.modelRequested,
+        resolvedModelRequested: requestTracking.resolvedModelRequested,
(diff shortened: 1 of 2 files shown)


eulervoid:
The image endpoint seems to return a custom response, i'd change that still to also throw an upstream error, with the same context. This way, all error responses follow the same schema.

PR #5034, opened 2025-11-05
Fix Flux steps calculation and concurrent request counting

voodoohop:
## Problem
- Flux images were using too few steps (dropping to 1 step at only 10-14 concurrent requests)
- Concurrent request count was inflated (showing 141-268 instead of actual ~10-20)
- Dead GPU servers with stale heartbeats were still being counted

## Solution
- **Improved step formula**: Maintains 4 steps up to 20 concurrent requests, then gradually drops to 1 at 50+
- **Fixed `countJobs()`**: Now filters out inactive servers (>45s since last heartbeat) before counting
- **Better image quality**: Under normal load, Flux will use 4 steps instead of 1

## Changes
- `availableServers.ts`: Added `filterActiveServers()` call in `countJobs()` to exclude dead servers
- `createAndReturnImages.ts`: Updated step calculation formula for more gradual degradation

## Impact
- 0-20 concurrent: **4 steps** (high quality)
- 30 concurrent: **3 steps**
- 40 concurrent: **2 steps**
- 50+ concurrent: **1 step** (minimum)

Previously was dropping to 1 step at just 10-14 concurrent due to counting dead servers.

### DIFF
--- a/image.pollinations.ai/src/availableServers.ts
+++ b/image.pollinations.ai/src/availableServers.ts
@@ -88,13 +88,15 @@ setInterval(decayErrors, 60 * 1000); // Every 1 minute
 setInterval(() => console.table(serverQueueInfo(SERVERS)), 10000);
 
 /**
- * Returns the total number of jobs for a specific type
+ * Returns the total load (pending + queued jobs) for a specific type
+ * Only counts active servers (with recent heartbeats)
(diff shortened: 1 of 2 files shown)


voodoohop:
Closing this PR - created from wrong branch (main instead of master).

Correct PR targeting master: #5035

PR #5053, opened 2025-11-05
Fix broken sign-in

eulervoid:
Simplify sign-in to fix the error and make it more robust.


voodoohop:
@claude can you review the fix? concisely. i believe it is great but just to follow protocol

PR #5085, opened 2025-11-07
Fix rate limit threshold consistency

voodoohop:
## Changes

- **Fixed threshold inconsistency**: Removed hardcoded `0.001` magic number
  - Allow check: `currentFill > 0`
  - Wait calculation: `pollenNeeded = -currentFill` (to reach 0)
  - Now both use consistent threshold of 0

- **Split error responses**: 
  - Concurrent request: No `waitMs`, retry immediately
  - Rate limit exhausted: Includes `waitMs` and countdown timer

- **Cleaner API**:
  - Made `waitMs` optional (undefined for concurrent requests)
  - Removed unnecessary `waitMs: 0` from allowed response
  - Return actual `remaining` balance in all cases

## Impact

- Fixes variable retry times after successful requests
- More predictable rate limiting behavior
- Clearer error messages for users

## Testing

Verified with parallel image generation test showing consistent countdown timers

### DIFF
--- a/enter.pollinations.ai/src/durable-objects/PollenRateLimiter.ts
+++ b/enter.pollinations.ai/src/durable-objects/PollenRateLimiter.ts
@@ -47,7 +47,7 @@ export class PollenRateLimiter extends DurableObject {
     async checkRateLimit(): Promise<{ 
         allowed: boolean; 
         remaining: number; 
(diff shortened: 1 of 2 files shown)


voodoohop:

PR #5090, opened 2025-11-07
fix: remove 25s timeout from nanobanana Vertex AI requests

voodoohop:
## Fix: Remove Artificial Timeout from Nanobanana

- Removes hardcoded 25-second timeout in `vertexAIClient.ts`
- Vertex AI image generation can legitimately take >25s, especially with reference images
- Let underlying fetch mechanism handle timeouts naturally
- Fixes "Vertex AI request timed out after 25 seconds" errors

**Root Cause:** Artificial timeout was too aggressive for image generation workloads

**Impact:** Nanobanana requests with reference images will no longer fail prematurely


voodoohop:
Closing in favor of #5091 which targets main instead of master

PR #5110, opened 2025-11-08
Improve test infrastructure and fix rate-limiter consumePollen scheme

eulervoid:
- **Provide a callback function to track consumed pollen in the frontend-key rate-limiter**
- **Improve testing infrastructure and fix rate limiter consumePollen scheme**

### DIFF
--- a/enter.pollinations.ai/src/middleware/rateLimit.durable.ts
+++ b/enter.pollinations.ai/src/middleware/rateLimit.durable.ts
@@ -100,7 +100,7 @@ export const frontendKeyRateLimit = createMiddleware<
     }
 
     c.set("frontendKeyRateLimit", {
-        consumePollen: stub.consumePollen,
+        consumePollen: (cost: number) => stub.consumePollen(cost),
     });
 
     await next();


enter.pollinations.ai/test/fixtures.ts
```diff
@@ -126,8 +125,26 @@ export const test = base.extend<Fixtures>({
             },
         });
         if (!createApiKeyResponse.data)
-            throw new Error("Failed to create API key");
+            throw new Error("Failed to create secret API key");
         const apiKey = createApiKeyResponse.data.key;
+        // expect(apiKey.startsWith("sk_")).toBe(true);
```
eulervoid, line 130: I found this check in a test, thats why i added it here, but then commented it out because it was failing. Should the secret api keys always start with `sk_`? If so it needs fixing.

voodoohop:
@eulervoid if you fix the merge conflicts we can merge this

eulervoid:
I rebased this on main, to remove the conflicts, but now the last test is failing, can you have a look @voodoohop ? I suspect its something in PollenRateLimiter, or maybe the expected behaviour just changed and the test is out of date. Also i thought is there maybe a way to test this without witing 15 seconds? Maybe we can manipulate the time somehow between the requests, i think there are tools for this kind of thing.

eulervoid:
I tried to use `vi.useFakeTimers` to speed up the test but unfortunately that does not work with cloudflare workers / durable objects. But maybe we can extract the bucket function and test it in isolation.

voodoohop:
@claude please review the changes. remember to pull again. concisely

PR #5194, opened 2025-11-12
docs: add missing query parameters to /generate/image/{prompt} endpoint

### LINKED ISSUE #5191: [Docs] /generate/image/{prompt} parameters are missing from documentation
[The documentation page for /generate/image/{prompt}](https://enter.pollinations.ai/api/docs#tag/image-generation/get/generate/image/{prompt}) does not include the list of query parameters (e.g., width, height, model, seed, nologo, etc.).

It should show model, width, height, etc but is currently completely missing, with only /generate/image/models showing up.

voodoohop:
Fixes #5191

- Add request.query schema with all parameters (model, width, height, seed, nologo, private, enhance, safe, etc.)
- Add response definitions for image/jpeg and image/png
- Fixes missing parameter documentation in OpenAPI spec

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/enter.pollinations.ai/src/routes/proxy.ts
+++ b/enter.pollinations.ai/src/routes/proxy.ts
@@ -257,6 +257,29 @@ export const proxyRoutes = new Hono<Env>()
                 "",
                 "API keys can be created from your dashboard at enter.pollinations.ai.",
             ].join("\n"),
+            request: {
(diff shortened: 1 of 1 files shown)


voodoohop:
@claude can you double check if this implementation fixes the issue? it seems like its missing something but i could be wrong

voodoohop:

PR #5195, opened 2025-11-12
fix: prevent ORB errors from empty cached images

voodoohop:
- Adds validation to skip empty cached images (content-length: 0)
- Prevents caching empty responses to R2 bucket
- Fixes ORB (Opaque Response Blocking) errors in browsers
- Root cause: corrupted cache entries with image/jpeg content-type but zero bytes
- Solution: Check R2 object size before serving, validate content-length before caching
- Tested: Empty cache entries now trigger cache miss, origin fetch returns valid image
- Already deployed to production (cloudflare-cache worker)

Replaces #5190 (was targeting wrong branch)


voodoohop:

PR #5166, opened 2025-11-11
fix: Preserve upstream HTTP status codes in error responses

voodoohop:
- Preserve original status codes from upstream services instead of wrapping as 500
- Introduce HttpError class for consistent error handling across services
- Pass through actual error responses (400, 403, etc.) to clients
- Improve error logging with status and response body details
- Content policy violations now return 400 (client error) instead of 500
- Affects: enter.pollinations.ai proxy, image.pollinations.ai models (seedream, nanobanana)

**Files Changed:**
- enter.pollinations.ai/src/routes/proxy.ts - Text/chat error handling
- image.pollinations.ai/src/httpError.ts - New HttpError class
- image.pollinations.ai/src/createAndReturnImages.ts - Use HttpError
- image.pollinations.ai/src/models/seedreamModel.ts - Preserve Seedream status codes
- image.pollinations.ai/src/vertexAIImageGenerator.ts - Preserve Gemini status codes

PR #5211, opened 2025-11-12
fix: Reintroduce UpstreamError for consistent error responses

voodoohop:
- Reintroduce `UpstreamError` for structured error responses
- Preserve upstream HTTP status codes (400, 403, 429, etc.)
- Maintain error message extraction from PR #5166
- Update all three proxy endpoints (text GET, image, chat completions)
- Combine benefits: status preservation + consistent error format

**Testing:**
- ✅ Content policy violation: 400 with structured error
- ✅ Rate limit: 429 with retry info
- ✅ Valid requests: 200 with proper response
- ✅ No test updates needed (tests check status codes only)


eulervoid (commented):
Looks good.

PR #5217, opened 2025-11-12
Integrate image cache directly into enter.pollinations.ai

voodoohop:
- Add R2 bucket binding to all environments (dev, local, staging, test)
- Create image-cache middleware with cache-first pattern
- Create image-cache utilities (cache key generation, metadata handling)
- Apply cache middleware BEFORE auth/rate limiting in image route
- Update IMAGE_SERVICE_URL to point directly to EC2 origin
- Add integration tests for cache behavior
- Remove standalone cloudflare-cache worker (no longer needed)

Benefits:
- Cache works without authentication (cache-first)

### DIFF
--- a/enter.pollinations.ai/src/middleware/rateLimit.durable.ts
+++ b/enter.pollinations.ai/src/middleware/rateLimit.durable.ts
@@ -43,13 +43,15 @@ export const frontendKeyRateLimit = createMiddleware<
         return next();
     }
 
-    log.debug("[RATE_LIMIT] Applying rate limit for publishable key: {keyId}", {
-        keyId: apiKey.id,
-    });
-
     // Get composite identifier: pk_{apiKeyId}:ip:{ip}
     const ip = c.req.header("cf-connecting-ip") || "unknown";
     const identifier = `pk_${apiKey.id}:ip:${ip}`;
+    
+    log.debug("[RATE_LIMIT] Applying rate limit for publishable key: {keyId}, IP: {ip}, identifier: {identifier}", {
+        keyId: apiKey.id,
+        ip,
+        identifier,
+    });
 
     // Get Durable Object for this (key + IP) combination
     const id = c.env.POLLEN_RATE_LIMITER.idFromName(identifier);


enter.pollinations.ai/src/utils/image-cache.ts
```diff
@@ -102,28 +89,69 @@ function createHash(str: string): string {
     return Math.abs(hash).toString(16).substring(0, 8);
 }
 
+/**
+ * Remove undefined values from an object
+ */
+function removeUndefined<T extends object>(obj: T): Partial<T> {
```
eulervoid: Remove undefined already exists in enter under `@/util.ts`

enter.pollinations.ai/src/utils/image-cache.ts
```diff
@@ -102,28 +89,69 @@ function createHash(str: string): string {
...
-    c: Context<Env>,
+    c: CacheContext,
 ): Promise<boolean> {
     try {
         // Store the image in R2 using the cache key directly
         const imageBuffer = await c.res.clone().arrayBuffer();
 
         // Get client information from request
-        const clientIp = c.get("connectingIp");
+        const clientIp = c.get("connectingIp") || c.req.header("cf-connecting-ip") || "";
```
eulervoid: I think we can remove the `c.get("connectingIp") if we're not adding it to the context beforehand.

enter.pollinations.ai/src/utils/image-cache.ts
```diff
@@ -102,28 +89,69 @@ function createHash(str: string): string {
...
 
         // Get additional client information
         const userAgent = c.req.header("user-agent") || "";
-        const referer =
-            c.req.header("referer") || c.req.header("referrer") || "";
+        const referer = c.req.header("referer") || c.req.header("referrer") || "";
         const acceptLanguage = c.req.header("accept-language") || "";
-        const requestId = c.req.header("cf-ray") || ""; // Cloudflare Ray ID uniquely identifies the request
+        const requestId = c.get("requestId") || c.req.header("cf-ray") || "";
```
eulervoid: We are always adding a `requestId`, so no need for the fallbacks here.

enter.pollinations.ai/src/utils/image-cache.ts
```diff
@@ -242,7 +202,7 @@ export async function cacheResponse(
 
         return true;
     } catch (error) {
-        console.error("[EXACT] Error caching response:", error);
+        console.error("[CACHE] Error caching response:", error);
```
eulervoid: This should use the logger instead of raw console.log

voodoohop:

PR #5218, opened 2025-11-12
fix: remove rate limiter timeout causing stuck requests

voodoohop:
## Problem
Rate limiter was getting stuck with persistent `retryAfterSeconds` that never decreased, blocking legitimate requests even after waiting.

## Root Cause
- `checkRateLimit()` set a 30s timeout on every allowed request
- If cache hit or `consumePollen()` wasn't called, timeout persisted
- Cache hits return early, bypassing `consumePollen()` in track middleware

## Solution
- Remove timeout from `checkRateLimit()` - only `consumePollen()` sets wait times
- Add `blockConcurrencyWhile` to prevent race conditions during initialization
- Cache hits bypass rate limiting entirely (correct behavior)

## Changes
- Remove `REQUEST_TIMEOUT_MS` constant and timeout logic
- Only set `nextAllowedTime` in `consumePollen()` after actual generation
- Add proper initialization with `blockConcurrencyWhile` per Cloudflare docs

## Testing
- Deployed to staging - logs show correct behavior
- Rate limiting working for publishable keys
- Cache hits return instantly without rate limit interference
- `consumePollen()` called after successful generations

### DIFF
--- a/pollinations.ai/src/components/Feeds/ImageDisplay.jsx
+++ b/pollinations.ai/src/components/Feeds/ImageDisplay.jsx
@@ -24,6 +24,19 @@ export const ImageDisplay = memo(function ImageDisplay({ image }) {
 
     // Show error if image has error property
     if (image?.error) {
+        // Handle both string errors and JSON error objects
+        let errorMessage = image.error;
+        let errorDetails = image.message;
+        
+        // If error is an object, extract the message
+        if (typeof image.error === 'object') {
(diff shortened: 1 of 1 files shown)


voodoohop:

voodoohop:
@claude can you explain to me why consumepollen(0) needs to be called? i thought the cache is earlier in the cahin and the rate limiter would never be hit if its cached

PR #5245, opened 2025-11-13
Fix event processing issues: move batching logic to the front of the processing chain

eulervoid:
Fix event delivery issues:
- Move batching logic to the front of the processing chain, this way only the failing batch is retried
- Add eventId and eventProcessingId to polar and tinybird events, for easier debugging in the future
- Mark events with an error status if they fail to deliver more than 5 times
- Add a test that verifies this works correctly


voodoohop:

PR #5248, opened 2025-11-13
fix: reduce pollen cache TTL from 60s to 5s

voodoohop:
- Reduce cache TTL from 60 seconds to 5 seconds
- Applies to both customer state and customer meters
- Provides more real-time pollen balance updates in dashboard
- Users will see balance changes within 5 seconds instead of 1 minute


voodoohop:

voodoohop:

PR #5284, opened 2025-11-14
fix: wrangler deploy with Vite-generated config

voodoohop:
## Fix Cloudflare Worker Deployment

- Remove environment flag from `wrangler deploy` command
- Vite already bakes environment into generated config during build
- Fixes error: "You have specified the environment 'production', but are using a redirected configuration"

**Root Cause:**
When using `@cloudflare/vite-plugin`, the environment is set during the build step (`CLOUDFLARE_ENV=production vite build`), not during deployment. Wrangler should deploy the pre-configured build output without additional env flags.

**Changes:**
- Split `npm run deploy:production` into separate build and deploy steps
- Use `npx wrangler deploy` without env specification


voodoohop:

eulervoid:
Hm ususally this was just a warning, but yea, the fix should do the trick, lets merge it.

PR #5293, opened 2025-11-14
fix: align cost and price calculations to use actual model

### LINKED ISSUE #5234: Total_cost seem not to be mapped to the used model price, review and fix if confirmed

ElliotEtag:
## Problems

**1. Silent Fallback Bug (Image Service):**
The image service used `.catch("flux")` in Zod validation, which silently replaced invalid models with "flux" without informing enter.pollinations.ai. This caused a mismatch:
- Enter tracked the **requested model** for pricing
- Image service used **"flux"** (fallback) for generation and cost calculation
- Result: Price/cost discrepancy

**2. Hardcoded Model Bug (Turbo/Flux):**
The `callComfyUI` function hardcoded `actualModel: 'flux'` in tracking data, even when "turbo" was used:
- User requests `model=turbo`
- Image service correctly uses turbo
- But reports `x-model-used: flux` to enter
- Result: Enter charges for turbo, calculates cost for flux

**Data Impact:**
- 14.9% of requests affected (9,614 out of 64,330)
- Total discrepancy: $1.76
- Top patterns:
  - undefined → openai → gpt-4.1-nano: 9,021 requests ($1.50)
  - turbo → turbo → flux: 401 requests ($0.072)
  - openai-fast/large variants: 192 requests ($0.184)

## Solutions

**1. Remove Silent Fallback:**
Invalid models now throw proper validation errors instead of silently falling back.

**2. Fix Hardcoded Model:**
Report actual model used (`safeParams.model`) instead of hardcoding 'flux'.

**3. Charge What You Deliver:**
Both cost and price now use the actual model used.

## Changes

**Billing Fix:**
- **`enter.pollinations.ai/src/middleware/track.ts`** - Align price to use actual model
- **`image.pollinations.ai/src/params.ts`** - Remove silent `.catch("flux")` fallback
- **`image.pollinations.ai/src/createAndReturnImages.ts`** - Fix hardcoded 'flux' in turbo/flux tracking
- **`image.pollinations.ai/src/index.ts`** - Move validation inside try-catch, proper 400 error handling

**Maintenance Improvement:**
- **`shared/registry/registry.ts`** - Add `DEFAULT_IMAGE_MODEL` and `DEFAULT_TEXT_MODEL` constants
- **`image.pollinations.ai` & `text.pollinations.ai`** - Import defaults from registry (single source of truth)

**Error Format Alignment:**
- **`text.pollinations.ai/server.js`** - Align error response format with image service

## Result

✅ Invalid models throw errors (no silent fallback)  
✅ Missing models use explicit default from registry  
✅ Turbo/flux tracking fixed (reports actual model)  
✅ **Cost = Price always** (100% of discrepancies resolved)  
✅ Default models centralized (change once, applies everywhere)  
✅ Consistent error format across image and text services

**Impact:** Eliminates $1.76 in billing discrepancies (14.9% of requests)

Fixes #5234

### DIFF
--- a/image.pollinations.ai/src/createAndReturnImages.ts
+++ b/image.pollinations.ai/src/createAndReturnImages.ts
@@ -27,15 +27,11 @@ import { logGptImageError, logGptImagePrompt } from "./utils/gptImageLogger.ts";
 import { callVertexAIGemini } from "./vertexAIImageGenerator.js";
 import { writeExifMetadata } from "./writeExifMetadata.ts";
 import type { ImageParams } from "./params.ts";
-import { withTimeoutSignal } from "./util.ts";
 import type { ProgressManager } from "./progressBar.ts";
 
 // Import model handlers
 import { callSeedreamAPI } from "./models/seedreamModel.ts";
-import {
-    callAzureFluxKontext,
-    type ImageGenerationResult as FluxImageGenerationResult,
-} from "./models/azureFluxKontextModel.js";
+import { callAzureFluxKontext } from "./models/azureFluxKontextModel.js";
 import { incrementModelCounter } from "./modelCounter.ts";
 
 dotenv.config();
@@ -228,7 +224,7 @@ export const callComfyUI = async (
                 buffer: resizedBuffer,
                 ...rest,
                 trackingData: {
-                    actualModel: "flux",
+                    actualModel: safeParams.model,
                     usage: {
                         completionImageTokens: 1,
                         totalTokenCount: 1,
@@ -249,7 +245,7 @@ export const callComfyUI = async (
             buffer: jpegBuffer,
             ...rest,
             trackingData: {
-                actualModel: "flux",
+                actualModel: safeParams.model,
                 usage: {
                     completionImageTokens: 1,
                     totalTokenCount: 1,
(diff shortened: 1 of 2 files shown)


ElliotEtag:
@claude review please

ElliotEtag:
@claude review this please

enter.pollinations.ai/src/middleware/track.ts
```diff
@@ -251,7 +251,7 @@ async function trackResponse(
         );
     const cost = calculateCost(modelUsage.model as ModelId, modelUsage.usage);
     const price = calculatePrice(
-        resolvedModelRequested as ServiceId,
+        modelUsage.model as ServiceId,
```
eulervoid: Not resolving the model will lead to aliases stopping working. So the resolving should just not do the fallback, but we still need to resolve aliases.

eulervoid: Actually this goes deeper.. `modelUsage.model` is a ModelId, not a ServiceId, so this will not work at all and is just not caught by type checking because of the type assertion. Unless there have been more changes to the Registry that i am not aware of. It seems the trend is to go back to treating the services that we provide and the providers as almost the same thing, which i personally think we should try to avoid.

PR #5308, opened 2025-11-15
Remove model defaults from backend services

voodoohop:
## Changes

- **image.pollinations.ai**: Remove 'flux' defaults and fallbacks
- **text.pollinations.ai**: Remove 'openai-fast' defaults  
- Backends now throw errors if model is missing/invalid
- Forces gateway to provide normalized, valid models

## Rationale

Backend services should not apply defaults or normalization - that's the gateway's job. This change:

- ✅ Enforces single source of truth (gateway normalizes models)
- ✅ Makes backend behavior explicit and predictable
- ✅ Prepares for gateway model normalization implementation
- ✅ Eliminates defensive programming that hides issues

## Files Modified

1. **image.pollinations.ai/src/params.ts** - Remove default model from schema
2. **image.pollinations.ai/src/makeParamsSafe.js** - Throw error if model missing/invalid
3. **image.pollinations.ai/src/createAndReturnImages.ts** - Remove fallback in model counter
4. **text.pollinations.ai/utils/parameterValidators.js** - Remove default model
5. **text.pollinations.ai/server.js** - Throw error if model missing

## Testing

Backend services will now fail fast if gateway doesn't provide a valid model. Next step is implementing gateway normalization in `enter.pollinations.ai`.

### DIFF
--- a/enter.pollinations.ai/src/schemas/openai.ts
+++ b/enter.pollinations.ai/src/schemas/openai.ts
@@ -2,6 +2,7 @@
 
 import { z } from "zod";
 import { getTextServices } from "../../../shared/registry/registry.ts";
+import { DEFAULT_TEXT_MODEL } from "../../../shared/registry/text.ts";
 
 const FunctionParametersSchema = z.record(z.string(), z.any());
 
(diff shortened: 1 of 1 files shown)


voodoohop:
## ✅ Added Default Model Constants

**New commits:**

- Add `DEFAULT_TEXT_MODEL` and `DEFAULT_IMAGE_MODEL` constants to registry
- Update `resolveServiceId()` to use constants instead of hardcoded strings
- Add Zod default to text schema using constant
- Centralizes default definitions in one place per service type

**Benefits:**
- DRY - defaults defined once
- Type-safe constants
- No magic strings
- Easy to maintain

**Testing:** Verified locally - all endpoints working correctly with defaults applied at gateway level.

voodoohop:

PR #5335, opened 2025-11-16
fix: remove duplicate auth and fix default model param

voodoohop:
- Remove redundant auth/rate-limit/polar middleware from `/image/*` route
- Global auth middleware (line 105) already covers all routes below it
- Use `DEFAULT_TEXT_MODEL` constant from registry for `/text/:prompt` default


voodoohop:

PR #5383, opened 2025-11-21
fix: Handle image model objects in frontend

voodoohop:
## Fix

- Process `model.name` and `model.description` from API objects
- Match text model processing pattern  
- Fix broken image model dropdown

## Root Cause

API switched from strings to objects but frontend still expected strings

## Files

- `pollinations.ai/src/utils/useModels.js`
- `pollinations.ai/src/hooks/useFetchModels.js`


voodoohop:

PR #5415, opened 2025-11-25
fix: Convert HTTP image URLs to base64 for Vertex AI vision

### LINKED ISSUE #5413: Gemini-large vision: empty mimeType error with OpenAI format
## Issue
- `gemini-large` model returns "Bad request" when using vision capabilities with OpenAI format
- Error in logs: `vertex-ai error: unable to submit request because it has an empty mimeType parameter in fileData`
- Public image URLs fail to process

## Expected
- OpenAI-compatible vision format should work (image_url in content array)

## Questions
- What is the correct payload format for gemini-large vision?
- Does the model require explicit mimeType in the request?
- Should the transform layer auto-detect mimeType from URL/content-type?

voodoohop:
## Problem
- Vertex AI vision fails with "empty mimeType parameter in fileData" error
- HTTP image URLs not directly supported by Vertex AI API

## Solution
- Add `imageUrlToBase64Transform.js` that fetches HTTP URLs → base64 data URLs
- Detects MIME type from Content-Type header or file extension
- Only applies to `vertex-ai` provider models

## Changes
- `transforms/imageUrlToBase64Transform.js` (new)
- `generateTextPortkey.js` - add transform to pipeline

Fixes #5413

### DIFF
--- a/enter.pollinations.ai/test/integration/text.test.ts
+++ b/enter.pollinations.ai/test/integration/text.test.ts
@@ -260,6 +260,48 @@ test(
 );
 
 test(
+    "POST /v1/chat/completions should accept image URL for vision models (Issue #5413)",
+    { timeout: 60000 },
+    async ({ apiKey, mocks }) => {
+        await mocks.enable("polar", "tinybird", "vcr");
+        const response = await SELF.fetch(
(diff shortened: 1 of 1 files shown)


voodoohop:

Issue #4983, opened 2025-11-04
Perplexity model returns invalid finish_reason causing validation error


voodoohop:
## Problem
Perplexity model returning `finish_reason` value that doesn't match OpenAI schema validation

## Error
```
Invalid option: expected one of "stop"|"length"|"tool_calls"|"content_filter"|"function_call"
```

## Details
- **Location**: `enter.pollinations.ai` response validation
- **Error Code**: `INTERNAL_ERROR` (ZodError)
- **Impact**: 500 errors for Perplexity model requests

## Root Cause
- Perplexity API returns non-standard `finish_reason` value
- Response schema validation expects OpenAI-compatible values only
- Need to handle Perplexity-specific finish reasons

## Solution
- Map Perplexity finish reasons to OpenAI schema
- Add validation bypass/transformation for Perplexity responses
- Update Zod schema to accept Perplexity variants

voodoohop:
https://github.com/pollinations/pollinations/pull/4985 closed by

PR #5439, opened 2025-11-27
fix: relax OpenAI schema validation for multi-provider compatibility

voodoohop:
Fixes schema validation errors when using Gemini and other non-OpenAI providers.

**Changes:**
- `finish_reason`: enum → string (accepts any provider value)
- `message.content`: nullable → nullish (accepts undefined)

**Problem:**
- Gemini returns `MAX_TOKENS`, `SAFETY`, `RECITATION` etc. instead of OpenAI values
- Truncated responses can have `undefined` content instead of `null`

**Related:** Closes similar issue to #4983 (Perplexity finish_reason)


voodoohop:

PR #5441, opened 2025-11-27
Clear event buffer, improve batching logic

eulervoid:
- Clear successfully sent events from the D1 buffer after one day
- Only send event batches if there are either a minimum of pending events, or events older than 30 seconds

As this is quite a critical part of the system, i'd like to add some tests before we merge.


voodoohop:
@eulervoid i think this one was superceded by newer PRs. just check if anything is missing and potentially close it

eulervoid:
Merged via a different PR

PR #5444, opened 2025-11-27
cleanup: Remove unused BPAIGen fallback code

voodoohop:
- Remove `bpaigenModel.ts` (dead code - kontext disabled on legacy API)
- Remove `kontextModel.ts` (only used by bpaigenModel)
- Remove unused import from `createAndReturnImages.ts`

Since kontext is now only available via enter.pollinations.ai, this fallback code was never reachable.


voodoohop:

PR #5453, opened 2025-11-28
feat: local balance tracking to reduce Polar API rate limits

voodoohop:
## Problem
- Polar API 429 rate limits on `customerMeters.list()`
- Every request calls Polar (cached 60s per user)

## Solution
- Store meter balances locally in KV (5min TTL)
- Only sync with Polar when cache miss or balance ≤ 0.1
- Decrement balance locally after billable requests

## Changes
- `polar.ts`: Local balance cache with get/set/decrement
- `track.ts`: Call `decrementBalance` after billable requests
- `polar.test.ts`: Balance tracking tests

## Config
- `LOCAL_BALANCE_TTL = 300` (5 min)
- `SYNC_THRESHOLD = 0.1` (sync when nearly depleted)

### DIFF
--- a/enter.pollinations.ai/src/middleware/polar.ts
+++ b/enter.pollinations.ai/src/middleware/polar.ts
@@ -18,9 +18,8 @@ type BalanceCheckResult = {
 const LOCAL_BALANCE_KEY_PREFIX = "polar:local:balance:";
 // TTL for local balance cache (5 minutes)
 const LOCAL_BALANCE_TTL = 300;
-// Threshold to force sync with Polar (negative = allow some overdraft before syncing)
-// -10 means users can go up to 10 pollen negative before we force a Polar sync
-const SYNC_THRESHOLD = -10;
+// Threshold to force sync with Polar (when total balance is low)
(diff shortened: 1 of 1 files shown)


voodoohop:

Juanoto2012:
what is polar api?

eulervoid (commented):
Looks good generally, I'm not completely sure about the very stateful approach, using the event cache to calculate the balance based on a last known balance and a database query summing the values of the events since then seems a bit simpler and less error prone to me, but on the other hand it would probably have worse performance. 

I also find it very strange that we would hit the rate limit even though the requests to get the meters are cached for 1m each time. Maybe polar counts all requests as one, even the event ones? I noticed yesterday that currently sending the cached events does not work correctly sometimes and events get sent in very small batches. So if they count all requests equally, this might be the reason for hitting the rate limit. I implemented a fix in PR #5441.

voodoohop:
> what is polar api?

we are using the payment processor https://polar.sh - checking the user's balance, handling payments....

PR #5454, opened 2025-11-28
fix: enable 4K resolution support for nanobanana-pro

voodoohop:
- Add `imageConfig` with `aspectRatio` and `imageSize` to Vertex AI API calls
- Auto-detect `imageSize` (1K/2K/4K) based on requested pixel count
- Calculate closest standard aspect ratio from width/height
- Increase nanobanana-pro default from 1024 to 2048 pixels

**Root cause**: Gemini 3 Pro Image API supports 4K via `imageSize` param but we weren't sending it


voodoohop:

PR #5460, opened 2025-11-28
fix: error status propagation and Azure max_tokens

voodoohop:
- Fix streaming errors returning HTTP 500 instead of upstream status (400, 401, etc.)
- Use `error.status` before `error.code` in generateTextBasedOnModel
- Fix `errorTypes[statusCode]` → `errorTypes[responseStatus]` lookup
- Include upstream error details in response
- Convert `max_tokens` → `max_completion_tokens` for Azure OpenAI models

**Before:** All streaming errors returned "Internal Server Error" (500)
**After:** Proper status codes with detailed error messages

Supersedes #5458


voodoohop:

PR #5463, opened 2025-11-29
feat: Add DELETE endpoint for cache invalidation

voodoohop:
## Summary

Adds a public API endpoint to delete cached images from the Cloudflare cache worker.

## Changes

- Add `DELETE /delete/prompt/:prompt` endpoint
- Deletes from both R2 bucket and Vectorize index
- Uses same URL structure as original requests (just prepend `/delete/`)
- Returns JSON response with deletion status

## Usage

To delete a cached image, use the same URL you would use to generate it, but:
1. Change the HTTP method to `DELETE`
2. Prepend `/delete/` to the path

**Example:**
```bash
# Original cached URL:
GET https://image.pollinations.ai/prompt/a%20beautiful%20sunset?model=flux&width=1024

# To delete from cache:
DELETE https://image.pollinations.ai/delete/prompt/a%20beautiful%20sunset?model=flux&width=1024
```

## Response

**Success (200):**
```json
{
  "success": true,
  "cacheKey": "_prompt_a_beautiful_sunset_model_flux_width_1024-abc123",
  "r2Deleted": true,
  "vectorDeleted": true
}
```

**Not Found (404):**
```json
{
  "success": false,
  "cacheKey": "_prompt_...",
  "message": "Cache entry not found or already deleted",
  "r2Deleted": false,
  "vectorDeleted": false
}
```

## Files Modified

- `cloudflare-cache/src/index.ts` - Add DELETE route and CORS support
- `cloudflare-cache/src/cache-utils.ts` - Add `deleteCacheEntry()` function
- `cloudflare-cache/src/vector-store.ts` - Add `deleteById()` method

### DIFF
--- a/image.pollinations.ai/cloudflare-cache/src/cache-utils.ts
+++ b/image.pollinations.ai/cloudflare-cache/src/cache-utils.ts
@@ -267,20 +267,32 @@ export async function deleteCacheEntry(
     let vectorDeleted = false;
 
     // Delete from R2
+    // Note: R2 delete() returns void and doesn't throw if key doesn't exist
+    // We first check if the object exists to provide accurate feedback
     try {
-        await env.IMAGE_BUCKET.delete(cacheKey);
-        console.log("[DELETE] Deleted from R2:", cacheKey);
-        r2Deleted = true;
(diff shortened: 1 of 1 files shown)


voodoohop:

Issue #4949, opened 2025-11-03
Clear buffered events after x amount of time


eulervoid:
Currently the event buffer is never cleared, processed events are just marked as processed. We should add a function to the cron handler that deletes processed events that are older that a threshold from the buffer. I would suggest something like 7 days or 1 month. Maybe we start with a month and shorten it if it gets expensive?

eulervoid:
Why is this blocked? Should i go ahead and implement it? Its a realtively small change and we can make the time frame configurable. Since Cloudflare D1 has a default time-travel of 30 days, even if we clear after 1 day, we would theoretically still have access to 30days of event history if we needed.

PR #5473, opened 2025-11-29
feat: clear buffered events after one day (reverts complex balance tracking)

voodoohop:
## Summary
- Clear successfully sent events from D1 buffer after one day
- Improve batching: only send if minimum pending events OR events older than 30 seconds

## Changes
- Includes eulervoid's simpler approach from #5441
- Reverts complex local balance tracking from #5453 (too complex, eulervoid's approach is better)

## Why
The local balance tracking with KV caching and decrement logic was overly complex. This PR uses eulervoid's simpler event buffer clearing approach instead.

Addresses #4949

Co-authored-by: eulervoid <3340175+eulervoid@users.noreply.github.com>

### DIFF
--- a/enter.pollinations.ai/src/events.ts
+++ b/enter.pollinations.ai/src/events.ts
@@ -83,7 +83,7 @@ export async function processEvents(
         }
     }
     // Clear successfully sent events
-    clearExpiredEvents(db);
+    await clearExpiredEvents(db);
 }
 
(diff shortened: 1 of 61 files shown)


voodoohop:
@claude can you review? especially do you think the rate limit errors will be fixed? i think balances have not been going down due to my fix

PR #5476, opened 2025-11-29
fix: add resolve.dedupe for zod to fix CI build

voodoohop:
- Add `resolve.dedupe: ["zod"]` to vite.config.ts
- Fixes CI build failure when importing from shared/registry/model-info.ts


voodoohop:

PR #5483, opened 2025-11-30
fix: Tier panel showing "Unknown Tier" and "0 pollen/day"

voodoohop:
## Problem
- Tier panel displayed "Unknown Tier" and "0 pollen/day" despite active subscription
- `active_tier_name` and `daily_pollen` missing from `/api/tiers/view` response

## Root Cause
- `polar` import is the Hono middleware, not the Polar SDK client
- `polar.products.get()` silently failed → product details never fetched

## Fix
- Use `c.var.polar.client.products.get()` instead of `polar.products.get()`

## Files Changed
- `src/routes/tiers.ts` - 2 occurrences fixed


voodoohop:
@claud can you review concisely

PR #5513, opened 2025-12-01
fix: preserve usage headers in image cache for pollen billing

voodoohop:
**Closing - not needed**

On cache hit, `track.ts` already skips billing (`isBilledUsage: false` when `cacheHit` is true). On cache miss, fresh headers come from origin. No header preservation needed.


voodoohop:

voodoohop:
@eulervoid this fix was unnecessary.

PR #5606, opened 2025-12-03
Fix Telegram CDN MIME detection for I2I

### LINKED ISSUE #5599: Fix "Invalid base64 image_url" error in Seedream/Seedance I2I
- **Report**: User `_dr_misterio_` reports "Invalid base64 image_url" error when using image-to-image with Telegram CDN URLs on Seedream and Seedance.
- **Context**: These URLs work with other models (Nanobanana).
- **Current Code**: `seedreamModel.ts` and `seedanceVideoModel.ts` *do* contain logic to download and convert images to base64, so the feature is not missing but likely malfunctioning.
- **Investigation**:
  - Check if `fetch` handles Telegram CDN headers/redirects correctly.
  - Verify `content-type` extraction (fallback to `image/jpeg` might be incorrect for Telegram images).
  - Check if downloaded content is actually an image (vs HTML/Error page).
  - Verify base64 string formatting and size limits.
- **Action**: Debug the download/conversion pipeline for these specific URLs.

voodoohop:
- Telegram CDN returns `application/octet-stream` instead of `image/jpeg`
- Use `file-type` lib to detect from magic bytes
- Fixes seedream/seedance I2I with Telegram URLs

Closes #5599

### DIFF
--- a/image.pollinations.ai/src/models/seedanceVideoModel.ts
+++ b/image.pollinations.ai/src/models/seedanceVideoModel.ts
@@ -1,7 +1,7 @@
 import debug from "debug";
 import sleep from "await-sleep";
-import { fileTypeFromBuffer } from "file-type";
 import { HttpError } from "../httpError.ts";
+import { downloadImageAsBase64 } from "../utils/imageDownload.ts";
 import type { ImageParams } from "../params.ts";
 import type { ProgressManager } from "../progressBar.ts";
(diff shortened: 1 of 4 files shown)


voodoohop:

voodoohop:
Extended to all models using base64 image input:
- Created shared `utils/imageDownload.ts` utility
- Applied to seedream, seedance, and nanobanana (vertexAI) models
- Eliminates 72 lines of duplicate code

PR #5666, opened 2025-12-04
fix: remove mistral-fast model and fix Bedrock max_tokens

voodoohop:
- Remove `mistral-fast` model (was misleadingly named - actually Llama 3.1 8B, not Mistral)
- Fix Bedrock default `max_tokens` from 16384 to 8192
- Fixes test failures for `chickytutor` model ("maximum tokens exceeds model limit of 8192")


voodoohop:

PR #5670, opened 2025-12-04
fix: Match newlines in image prompt route pattern

### LINKED ISSUE #5669: fix: Image endpoint returns 404 for prompts containing newlines
**Problem:**
- Image generation endpoint returns 404 for prompts containing URL-encoded newlines (`%0A`)
- Route pattern `{.+}` doesn't match newline characters in JavaScript regex

**Root cause:**
- JavaScript regex `.+` doesn't match `\n` by default
- When `%0A` is decoded to `\n`, route matching fails

**Solution:**
- Change route pattern from `{.+}` to `{[\\s\\S]+}` which matches any character including newlines

**Affected users:**
- luckynumbers (storydreaming.com) - prompts with multi-line descriptions

voodoohop:
Fixes #5669

- Change route pattern from `{.+}` to `{[\\s\\S]+}`
- `.+` doesn't match newlines in JS regex
- `[\\s\\S]+` matches any character including `\n`
- Fixes 404 for prompts with URL-encoded newlines (`%0A`)

### DIFF
--- a/enter.pollinations.ai/src/routes/proxy.ts
+++ b/enter.pollinations.ai/src/routes/proxy.ts
@@ -408,7 +408,7 @@ export const proxyRoutes = new Hono<Env>()
             await c.var.auth.requireAuthorization();
             await checkBalance(c.var);
 
-            // Get prompt from validated param (using :prompt{.+} regex pattern)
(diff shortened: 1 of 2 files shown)


voodoohop:

voodoohop:
@claude review the changes please

PR #5845, opened 2025-12-10
fix: trigger app submission workflow when app:review label is added

voodoohop:
- Add `labeled` event type to workflow trigger
- Only run when `app:review` label is specifically added (not other labels)
- Fixes issue where free-form app submissions weren't processed

**Problem**: Users creating free-form issues (not using the template) don't get the `app:review` label automatically, so the PR creation workflow never triggers.

**Solution**: Now when a maintainer manually adds `app:review` label, the workflow triggers automatically.


voodoohop:

PR #5864, opened 2025-12-10
Fix: Claude/Bedrock image URL to base64 conversion

### LINKED ISSUE #5862: Fix: Claude/Bedrock models require base64 image conversion
Claude models via Bedrock fail with HTTP image URLs (500 error) because `imageUrlToBase64Transform` only applies to Vertex AI.

**Problem:**
- `gemini` works with HTTP URLs ✅
- `claude-large` fails with HTTP URLs ❌
- `claude-large` works with base64 ✅

**Fix:** Extend transform to also apply to `provider === "bedrock"`

voodoohop:
Extends `imageUrlToBase64Transform` to also apply to Bedrock provider.

**Changes:**
- Add `provider === "bedrock"` check alongside `vertex-ai`
- Update comments to reflect both providers
- Add test for Claude with image URL

Fixes #5862

### DIFF
--- a/enter.pollinations.ai/test/integration/text.test.ts
+++ b/enter.pollinations.ai/test/integration/text.test.ts
@@ -339,6 +339,48 @@ test(
 );
 
 test(
+    "POST /v1/chat/completions should accept image URL for Claude/Bedrock models (Issue #5862)",
+    { timeout: 60000 },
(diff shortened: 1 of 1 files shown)


voodoohop:

PR #5891, opened 2025-12-11
fix: Rename catch parameter to prevent variable shadowing in gptImageLogger

### LINKED ISSUE #5889: Error logging system broken in gptimage
I was testing different image dimensions with the gptimage model and kept getting this confusing error:

```
{
  "error": "Internal Server Error",
  "message": "logError2 is not a function"
}
```

At first I thought this was related to #5820 (dimension validation), but the error message didn't make sense. Why would a dimension error mention "logError2"? I searched the entire repository for `logError2` and found... nothing. The function doesn't exist anywhere in the codebase.

That seemed really strange, so I started digging through the error handling code to figure out where this phantom "logError2" was coming from. I traced the actual image generation flow and found that errors go through `image.pollinations.ai/src/utils/gptImageLogger.ts`. 

Looking at the code, everything seemed fine at first - all the `logError()` calls looked normal:

```
// Line 13: const logError = debug("pollinations:error");

export async function logGptImageError(...) {
    try {
        // ... logging logic ...
    } catch (logError) {  
        logError("Error logging gptimage error:", logError.message);
    }
}
```

There's only one `const logError` declaration, but then I noticed the catch block is also using `logError` as the parameter name. This is hard to spot, but it means the catch parameter is shadowing the actual `logError` function from line 13. So when line 132 tries to call `logError()`, it's actually trying to call the Error object as a function.

What makes this particularly tricky is that the source code looks perfectly valid - there's no obvious undefined function being called. The bug only appears at runtime after TypeScript compilation. My guess is that the transpiler automatically renamed the conflicting variable to `logError2` to avoid a compilation error, which is why we see that name in the runtime error even though it's not in the source code.

This means any error in gptimage/nanobanana/kontext models shows this "logError2" message instead of the actual error. Users can't see what's really wrong (like "unsupported dimensions" from #5820), and developers can't debug issues properly.

The fix is simple - just rename the catch parameter to avoid the collision:

```
} catch (error) {
    logError("Error logging gptimage error:", error.message);
}
```

elethy:
Fixes "logError2 is not a function" error by renaming catch parameter to avoid shadowing the outer logError function.

Changed catch parameter from `logError` to `error` in `logGptImageError` and `logGptImagePrompt`.

Closes #5889

### DIFF
--- a/image.pollinations.ai/src/utils/gptImageLogger.ts
+++ b/image.pollinations.ai/src/utils/gptImageLogger.ts
@@ -128,8 +128,8 @@ export async function logGptImageError(
         await fsPromises.appendFile(logFile, `${logEntry}\n`);
 
         logOps("Logged gptimage error to", logFile);
-    } catch (logError) {
+    } catch (error) {
         // Non-blocking error handling for logging
-        logError("Error logging gptimage error:", logError.message);
+        logError("Error logging gptimage error:", error.message);
     }
 }


elethy:
Thanks for the quick merge! Happy to contribute to the project. 🙏
Excited to level up to Flower Tier! 🌱

PR #5897, opened 2025-12-11
Add tests for tool calls

eulervoid:
Adds tests for tool calls for all models and updates the snapshots. 

This revealed several issues:
- `deepseek` seems to be extremely slow, leading to the test timing out
- `perplexity` has some issues
- `gemini-search` fails because it does not allow search tools and regular tools in the same request
- the test for image urls is failing, response message content is undefined


hlohaus:
I know a real bug with tools, you can't use tools and seed in the parameters at the same time. @eulervoid

eulervoid:
Hm, I'm actually using the seed parameter in the tests and it seems to work fine. Could you provide an example where it does not work maybe @hlohaus ?

hlohaus:
@eulervoid I’m no longer able to reproduce the issue, and I’m not sure which model was involved.
But each model handles tools differently, and my auto‑generated code doesn’t work with Claude because the arguments aren’t defined in the first tool event.

PR #5919, opened 2025-12-12
Add retry delay to event processing, batch database queries to improve performance

eulervoid:
- Add exponential backoff delay that ramps up from 100ms to 10s over 5 attempts
- Send consecutive queries to D1 as batches to improve performance


voodoohop:
@eulervoid looks good!

PR #5930, opened 2025-12-12
fix: robust JSON parsing for Claude responses

voodoohop:
- Explicitly tell model to return RAW JSON (no markdown fences)
- Add sed to strip code fences as fallback

Fixes issue where Claude wraps JSON in ```json``` blocks


voodoohop:

Issue #5877, opened 2025-12-11
Missing clarification in Flux image API example


losingdivine:
Hey team 👋

I noticed a small inconsistency in the Flux image generation example in the API docs.

In the pricing table, Flux is listed as 0.00012 pollen/image, but in the example request the output parameters (like size and format) aren’t clearly described. This makes it a bit confusing for new developers who want to estimate pollen usage.

🔍 What seems missing

No explanation of how output size affects pollen cost

No example showing width, height, or resolution fields

No note saying whether default size changes pollen cost

💡 Suggested improvements

Add a line explaining how output resolution scales cost

Include a full request example with all fields

Mention the default output size and whether it impacts pollen charge

**If** you want, I can help create a PR updating the docs.
Thanks for the awesome platform! 🌸****

ElliotEtag:
Closing as completed. ✅

The pricing clarifications were added by @voodoohop in December 2025:
- Tooltips on price badges explaining the unit
- Legend below the pricing table: `/image` = flat rate (any resolution), `/M` = per million tokens, `/sec` = per second of video

To confirm: **Image pricing is per-image, not per-pixel.** A 256x256 costs the same as 1024x1024 for the same model.

Thanks for the feedback! 🌸

PR #5977, opened 2025-12-13
fix: clarify image pricing in pricing UI

voodoohop:
- Add pricing type to model tooltips (flat rate vs token-based)
- Add legend explaining /image, /M, /sec pricing units

Addresses #5877


ElliotEtag:
@voodoohop can merge for me

PR #5978, opened 2025-12-13
feat: add model gating by API key permissions

voodoohop:
## Summary
- Adds model-level access control for API keys
- Users can restrict keys to specific models (text/image) during creation
- Requests for non-allowed models return 403 Forbidden

## Changes
- **`/api/api-keys/create`** - Server-side endpoint using `auth.api.createApiKey()` with permissions
- **`requireModelAccess(model)`** - Middleware enforces model permissions
- **`ModelPermissions`** UI component - Expandable model selection using shared registry
- **Auth middleware** - Reads permissions from `verifyApiKey` response

## Architecture
```
User selects models → POST /api/api-keys/create → auth.api.createApiKey({permissions}) → DB
                                                                                          ↓
API request with key → verifyApiKey → permissions from response → requireModelAccess() → Allow/Deny
```

## Key Implementation Details
- Uses `better-auth`'s server-side `createApiKey()` with `permissions` param (server-only field)
- Single atomic call replaces previous two-step flow (client create + permissions endpoint)
- Input sanitization via Zod transform filters `_` prefixed UI markers
- Uses `@shared/registry/*` path alias for model lists

## Backward Compatibility
- Empty/null permissions = all models allowed
- Session-based auth (no API key) = all models allowed

## Testing
- ✅ 3 integration tests for permit/deny/unrestricted scenarios
- ✅ Manual verification: `flux` allowed → 200, `turbo` blocked → 403


thomashmyceli:
@claude can you re-review

voodoohop:

ElliotEtag:
## UI Improvements for API Keys & Layout Fixes

@voodoohop please review these changes:

**API Key Display:**
- Added "Models" column with badge showing "All" or count of allowed models
- Keys now show last 4 chars (`...XXXX`) instead of prefix
- Publishable keys: click to copy full key (display stays compact)
- Backend stores `keySuffix` in metadata for new keys

**Layout Fixes:**
- Auto-sized grid columns (removed hardcoded widths)
- Hidden scrollbars while keeping scroll functionality
- Fixed page-wide horizontal overflow issue
- Applied same scrollbar fix to pricing list

All tested locally and working correctly.

ElliotEtag:
image
image
image

enter.pollinations.ai/src/middleware/model.ts
```diff
@@ -0,0 +1,65 @@
...
+import { HTTPException } from "hono/http-exception";
+import type { EventType } from "@shared/registry/types.ts";
+import { resolveServiceId, type ServiceId } from "@shared/registry/registry.ts";
+import { DEFAULT_TEXT_MODEL } from "@shared/registry/text.ts";
+import { DEFAULT_IMAGE_MODEL } from "@shared/registry/image.ts";
+
+export type ModelVariables = {
+    model: {
+        /** The raw model string from the request (before resolution) */
+        raw: string;
```
eulervoid: Its a minor thing but i'd like to stick with the terminology we use elsewhere for this which would be `requested` and `resolved`, then it matches up with e.g. the events table, the data in tinybird etc.

enter.pollinations.ai/src/routes/api-keys.ts
```diff
@@ -0,0 +1,80 @@
...
+                name,
+                prefix,
+                userId: user.id,
+                metadata: { description, keyType },
+                permissions,
+            },
+        });
+
+        // Store keySuffix (last 4 chars) for all keys, plus plaintextKey for publishable keys
+        if (apiKey.key) {
```
eulervoid: I think the apikey plugin already stores the `start` field for identifying keys, so maybe we can use that instead of storing an extra keySuffix field?

enter.pollinations.ai/src/routes/api-keys.ts
```diff
@@ -0,0 +1,80 @@
...
+ * API key management routes.
+ * Uses better-auth's server API to create keys with permissions in a single call.
+ */
+export const apiKeysRoutes = new Hono<Env>()
+    .use(auth({ allowSessionCookie: true, allowApiKey: false }))
+    /**
+     * Create an API key with optional model permissions.
+     * Uses auth.api.createApiKey() which supports server-only fields like permissions.
+     */
+    .post("/create", validator("json", CreateApiKeySchema), async (c) => {
```
eulervoid: I find it a bit confusing that we now have two routes to create api keys, i'm wondering if maybe there is a way to achieve this without adding a second creation route?

voodoohop:
@eulervoid Refactored per your feedback:

- Removed custom `/create` route
- Now uses better-auth's native `authClient.apiKey.create()` 
- Added `/:id/update` for permissions (server-only field)

Flow: create via better-auth → update permissions if restricted. Single custom endpoint instead of two.

voodoohop:
Also addressed your other review comments:

- ✅ **`raw` → `requested`**: Renamed to match event/tinybird terminology
- ✅ **`keySuffix` removed**: Now using better-auth's built-in `start` field for key identification
- ✅ **Two create routes → one**: Removed custom `/create`, using native `authClient.apiKey.create()` + `/:id/update` for permissions

PR #5987, opened 2025-12-13
fix: Add delay before event delivery to prevent rate limiting (#5909)

### LINKED ISSUE #5909: High error rate on event delivery
There are a lot of events not being delivered. My best guess is that this happens due to rate limiting, so we should add a small delay between delivery attempts to mitigate this.

FabioArieiraBaia:
## Description

This PR adds a small delay before each event delivery attempt to mitigate high error rates caused by rate limiting, addressing issue #5909.

## Problem

There are many events not being delivered successfully. The best guess is that this happens due to rate limiting when making too many requests in quick succession to Polar and Tinybird APIs.

## Solution

Added a 100ms delay before each delivery attempt to both Polar and Tinybird APIs. This small delay helps prevent hitting rate limits while maintaining good throughput.

## Changes

**Modified:** `enter.pollinations.ai/src/events.ts`

1. **Added delay before Polar delivery:**
   - 100ms delay before calling `polar.events.ingest()`
   - Helps prevent rate limiting on Polar API

2. **Added delay before Tinybird delivery:**
   - 100ms delay before calling Tinybird ingest endpoint
   - Helps prevent rate limiting on Tinybird API

## Implementation Details

The delay is minimal (100ms) to balance between:
- Preventing rate limiting errors
- Maintaining good throughput for event processing
- Not significantly impacting overall processing time

This works in conjunction with the existing exponential backoff retry mechanism (lines 95-113) which handles retries after failures.

## Testing

- ✅ Delay is applied before each delivery attempt
- ✅ Works for both Polar and Tinybird deliveries
- ✅ Minimal impact on processing time (100ms per batch)
- ✅ Should reduce rate limiting errors significantly

## Expected Impact

- **Before:** High error rate due to rate limiting when sending events too quickly
- **After:** Reduced error rate with small delay between delivery attempts

## Credits

**Developed by:** Fábio Arieira  
**Website:** https://fabioarieira.com  
**Production Projects:**
- IA-Books: https://iabooks.com.br
- ViralFlow: https://fabioarieira.com/viralflow
- Real Estate Platform: https://fabioarieira.com/imob

Full Stack Developer specializing in backend systems, API integrations, and error handling.

---

Resolves #5909

### DIFF
--- a/enter.pollinations.ai/src/events.ts
+++ b/enter.pollinations.ai/src/events.ts
@@ -351,6 +351,11 @@ async function sendPolarEvents(
             status: "skipped",
             ...deliveryStats,
         };
+    
+    // Add small delay before delivery to prevent rate limiting
+    // This helps mitigate high error rates on event delivery
+    await new Promise((resolve) => setTimeout(resolve, 100));
+    
     let ingested = 0;
     try {
         const response = await polar.events.ingest({
@@ -407,6 +412,11 @@ async function sendTinybirdEvents(
             status: "skipped",
             ...deliveryStats,
         };
+    
+    // Add small delay before delivery to prevent rate limiting
+    // This helps mitigate high error rates on event delivery
+    await new Promise((resolve) => setTimeout(resolve, 100));
+    
     let ingested = 0;
     try {
         const response = await fetch(tinybirdIngestUrl, {


eulervoid:
Hey, we already added a fix for this with exponential backoff, i don't think that adding another 100ms delay will improve the situation further. If you provide some reasoning why this is still needed, i'd be happy to reopen the PR.

Issue #5219, opened 2025-11-12
Investigate negative balances


eulervoid:
Some users have negative balances in polar, we should investigate if this is due to the 1m delay in event processing or if there is more at play. We could also come up with a more robust way to prevent users from going into the negative, e.g. by using the event cache to augment the polar balance when doing the check.

PR #6025, opened 2025-12-15
fix: prevent negative balances with local pending spend tracking

voodoohop:
## Problem
Users are going negative in their Polar balance because:
1. Polar meter balance is cached (60s TTL)
2. Events are batched before sending to Polar
3. Polar takes variable time to process events and update meters

This allows rapid requests to exceed the balance before Polar reflects the spend.

## Solution
- Query D1 events from the last 10 minutes to calculate "recent local spend"
- Subtract this from the cached Polar balance to get an "effective balance"
- Block requests when effective balance ≤ 0
- Display pending spend in dashboard for transparency

## How it works
```
effective_balance = polar_cached_balance - recent_local_spend
```

The 10-minute window is conservative to account for Polar's variable processing time. This may cause brief over-restriction after Polar catches up, but:
- Prevents negative balances
- Self-corrects as events age out
- No new storage needed (uses existing D1 event table)

## Changes
- Add `getRecentSpend()` function to query recent billable events
- Modify `requirePositiveBalance()` to subtract recent spend from meter balances
- Add composite index `idx_event_user_billed_created` for query performance
- Add `/customer/pending-spend` API endpoint
- Display pending spend in dashboard PollenBalance component

## Dashboard Display
Users now see their pending spend next to their balance:
```
0.32 pollen (-0.368 pending)
```

## Migration
After merge, run:
```bash
npm run migrate:staging   # then
npm run migrate:production
```

## Trade-offs
| Benefit | Trade-off |
|---------|-----------|
| Prevents negative balances | May briefly over-restrict after Polar syncs |
| Uses existing data | +1 D1 query per balance check (~5-10ms) |
| Self-healing | 10-min window is conservative |
| Transparent to users | Dashboard shows pending spend |

Addresses #5219

### DIFF
--- a/enter.pollinations.ai/src/client/components/pollen-balance.tsx
+++ b/enter.pollinations.ai/src/client/components/pollen-balance.tsx
@@ -6,11 +6,13 @@ type PollenBalanceProps = {
         pack: number;
     };
     dailyPollen?: number;
+    pendingSpend?: number;
 };
 
 export const PollenBalance: FC<PollenBalanceProps> = ({
     balances,
     dailyPollen = 15,
+    pendingSpend = 0,
 }) => {
     // Use real balances
     const freePollen = balances.tier; // Free pollen from tier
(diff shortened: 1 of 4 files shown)


voodoohop:
@claude can you review

voodoohop:
@claude can you review again?

PR #6046, opened 2025-12-15
feat: auto-manage Polar subscriptions (remove manual activation)

ElliotEtag:
## Summary

Automates Polar subscription management - **users no longer need to manually activate their tier**.

## UX Improvements 🎉

- **Removed activate button** - subscriptions are now automatic
- **Auto-subscription on signup** - new users get SPORE tier immediately
- **Auto-reactivation on cancel** - if user cancels, we re-subscribe them to their D1 tier

## How It Works

```
New user signup → Auto-subscribe to SPORE tier
Tier change in D1 → GitHub Action → Admin endpoint → Polar API
User cancels → Webhook → Auto-reactivate to D1 tier
```

## Features

- **Admin endpoint** `/api/admin/sync-tier` - for GitHub Actions to call after D1 tier updates
- **Webhook handler** - auto-reactivates subscriptions when users cancel
- **Retry logic** - up to 3 attempts with backoff on Polar API failures

## Files Changed

| File | Purpose |
|------|---------|
| `tier-sync.ts` | Core sync logic with retry |
| `routes/admin.ts` | Admin endpoint (ENTER_TOKEN auth) |
| `routes/webhooks.ts` | Polar webhook handler |
| `auth.ts` | Auto-subscription on signup |
| `tier-panel.tsx` | **Removed activate button**, updated UI messages |

## Security

- Admin endpoint secured by `ENTER_TOKEN` (constant-time comparison)
- Webhook signature verified with timestamp validation (replay protection)
- D1 is source of truth - users cannot self-assign higher tiers

## Testing

```bash
# Test admin endpoint
curl -X POST https://enter.pollinations.ai/api/admin/sync-tier \
  -H "Authorization: Bearer $ENTER_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"userId": "USER_ID"}'
```

## Next Steps

- Create GitHub Action to call admin endpoint after D1 tier updates


ElliotEtag:
polly 

Can you review this, please?

ElliotEtag:
polly can you check again?

enter.pollinations.ai/src/routes/webhooks.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    if (!webhookSecret) {
+        log.warn("POLAR_WEBHOOK_SECRET not configured");
+        return c.json({ error: "Webhook not configured" }, 500);
+    }
+
+    const signature = c.req.header("webhook-signature");
+    if (!signature) {
+        log.warn("Missing webhook signature");
+        return c.json({ error: "Missing signature" }, 401);
```
eulervoid: Throw an error here instead to get a standardized error response.

eulervoid (commented):
Hm, looks a bit overly complex to me to be honest, or maybe i just don't understand it? Why do we need to cache the ids in KV, and why do we need to queue these events, i think tier changes should not happen so often that we need to worry about rate limiting issues like with events, but maybe the AI took some inspiration from this code? Generally i think it can be quite simple. 

Tier in D1 gets an onUpdate listener -> updates the Polar Subscription when its updated.

Or am i missing something important?

ElliotEtag:
polly can you review, i simplified a lot thanks to @eulervoid 's review

ElliotEtag:
hey polly , and now, we getting there?

ElliotEtag:
@eulervoid 

can you review again?

enter.pollinations.ai/src/routes/admin.ts
```diff
@@ -0,0 +1,130 @@
...
+
+            const userTier = users[0]?.tier;
+            if (!userTier || !isValidTier(userTier)) {
+                return c.json({ error: "User has invalid tier" }, 400);
+            }
+            targetTier = userTier;
+        }
+
+        // Initialize Polar client
+        if (!c.env.POLAR_ACCESS_TOKEN) {
```
eulervoid: This check is unnecessary as cloudflare ensures via types that all variables are present.

enter.pollinations.ai/src/routes/admin.ts
```diff
@@ -0,0 +1,130 @@
...
+    getTierProductMap,
+    type TierName,
+} from "../tier-sync.ts";
+
+const log = getLogger(["hono", "admin"]);
+
+export const adminRoutes = new Hono<Env>()
+    .use("*", async (c, next) => {
+        // Use ENTER_TOKEN for admin authentication (already in GH secrets)
+        const adminKey = c.env.ENTER_TOKEN;
```
eulervoid, line 21: I would prefer to use the admin plugin that is already installed to verify that the user has permissions (see https://www.better-auth.com/docs/plugins/admin). Manually verifying the token works, but i think its a bit error prone and would need to be duplicated in other routes.

enter.pollinations.ai/src/routes/admin.ts
```diff
@@ -0,0 +1,130 @@
...
+            result |= providedKey.charCodeAt(i) ^ adminKey.charCodeAt(i);
+        }
+        if (result !== 0) {
+            return c.json({ error: "Unauthorized" }, 401);
+        }
+
+        return await next();
+    })
+    .post("/sync-tier", async (c) => {
+        const body = await c.req.json<{ userId: string; tier?: string }>();
```
eulervoid, line 44: No unauthorized access protection for this route?

enter.pollinations.ai/src/routes/admin.ts
```diff
@@ -0,0 +1,130 @@
...
+                targetTier,
+                attempts: result.attempts,
+            });
+        } else {
+            log.error("Tier sync failed for user {userId}: {error}", {
+                userId: body.userId,
+                error: result.error,
+            });
+
+            return c.json(
```
eulervoid: Throw an error here to get a unified error response.

enter.pollinations.ai/src/routes/tiers.ts
```diff
@@ -70,11 +70,12 @@ function getTierFromProductId(
 }
 
 function shouldShowActivateButton(
```
eulervoid: Seems like this function can be removed.

enter.pollinations.ai/src/tier-sync.ts
```diff
@@ -0,0 +1,156 @@
...
+
+export interface TierProductMap {
+    spore: string;
+    seed: string;
+    flower: string;
+    nectar: string;
+    router: string;
+}
+
+export function getTierProductMap(env: {
```
eulervoid: This could be loaded and cached dynamically by getting the ids from polar via their slug. Then we would not need to keep the ids hardcoded in wrangler.toml which removes a lot of friction when spinning up new environments or making changes in polar.

enter.pollinations.ai/src/tier-sync.ts
```diff
@@ -0,0 +1,156 @@
...
+                    attempt,
+                    max: MAX_ATTEMPTS,
+                    userId,
+                    error: lastError,
+                },
+            );
+
+            if (attempt < MAX_ATTEMPTS) {
+                await new Promise((resolve) =>
+                    setTimeout(resolve, RETRY_DELAY_MS * attempt),
```
eulervoid: This could use exponential backoff, there is already a function available for calculating the delay from the number of attempts in `src/util.ts` called `exponentialBackoffDelay`.

eulervoid (commented):
Mostly minor improvement suggestions.

enter.pollinations.ai/src/routes/admin.ts
eulervoid: Ah, i didn't see at first that its a middleware doing the ENTER_TOKEN auth, this comment wrong and can be ignored.

ElliotEtag:
**Re: Admin plugin vs ENTER_TOKEN**

The admin plugin requires a **user session**, but this endpoint is designed for **machine-to-machine auth** (GitHub Actions calling after tier changes in D1).

Flow:
```
User contributes → GH Action updates D1 tier → GH Action calls /api/admin/sync-tier → Polar subscription updates
```

Since GitHub Actions doesn't have a user session, ENTER_TOKEN is the correct approach here. 

If we add more admin routes in the future, we can extract the ENTER_TOKEN auth to a reusable middleware to address the duplication concern.


---
makes sense @eulervoid ?

ElliotEtag:
polly please review. take into consideration all the comments

ElliotEtag:
@polly review again please

ElliotEtag:
@polly review again please

ElliotEtag:
@eulervoid how does it look now?

enter.pollinations.ai/src/routes/polar.ts
```diff
@@ -17,6 +17,53 @@ export const productSlugs = [
...
+): Promise<TierProductMap> {
+    if (cachedTierProducts && Date.now() - tierCacheTime < TIER_CACHE_TTL) {
+        return cachedTierProducts;
+    }
+    const result = await polar.products.list({
+        limit: 100,
+        metadata: { slug: [...tierSlugs] },
+    });
+    const map: Partial<TierProductMap> = {};
+    for (const product of result.result.items) {
```
eulervoid: This check is unnecessary as you are already loading just the products with matching slugs.

enter.pollinations.ai/src/routes/polar.ts
```diff
@@ -17,6 +17,53 @@ export const productSlugs = [
...
+
+export function isValidTier(tier: string): tier is TierName {
+    return ["spore", "seed", "flower", "nectar", "router"].includes(tier);
+}
+
+let cachedTierProducts: TierProductMap | null = null;
+let tierCacheTime = 0;
+const TIER_CACHE_TTL = 5 * 60 * 1000; // 5 minutes
+
+export async function getTierProductMapCached(
```
eulervoid: Write a normal async function, and use the `cached` function from util, to cache the result in KV. In memory caching will not work reliably in cloudflare workers because each request might run in a different isolate.

Let me know if you want me to do the quick update for this.

enter.pollinations.ai/src/routes/webhooks.ts
```diff
@@ -0,0 +1,261 @@
...
+        };
+        product?: {
+            id: string;
+            name?: string;
+            metadata?: Record<string, unknown>;
+        };
+    };
+}
+
+async function verifyWebhookSignature(
```
eulervoid: Polar provides an inbuilt function to validate incoming webooks with their sdk, which should really simplify this code. See [Polar Webhook Docs](https://polar.sh/docs/integrate/webhooks/delivery)

enter.pollinations.ai/src/auth.ts
```diff
@@ -184,6 +197,34 @@ function onAfterUserCreate(polar: Polar) {
...
+                        await polar.subscriptions.create({
+                            productId: defaultTierProductId,
+                            customerId: existingCustomer.id,
+                        });
+                        ctx.context.logger.info(
+                            `Created default tier subscription for user ${user.id}`,
+                        );
+                    }
+                } catch (subError) {
+                    // Log but don't fail user creation if subscription fails
```
eulervoid: The code to update the tier should be extracted into its own function to make the intent clearer.

eulervoid (commented):
Its getting there, but there are still some things i would clean up.
1. Use the polar sdk to verify the incoming webhook this should make the code much simpler and make sure we do it correctly
2. Cache the tier map in KV, there is a simple utility function `cached` in `src/util.ts` that lets you easily cache any async function result in KV
3. I'm a bit uneasy about creating the subscription in the `onAfterUserCreate` hook. Its the right place for this, but we need to take care not to mess with the other things, as it could lead to polar and d1 getting out of sync which would be annoying to fix. Thats why i'd like to encapsulate the code in a function to keep the original hook nice and small.

eulervoid (commented):
Nice looking good. Let me do some last minor fixes myself really quick, the i thing its ready.

ElliotEtag:
Whenever you're ready, just merge it. Let me know so I can check for new SPORE subs that everything goes well

PR #6076, opened 2025-12-16
fix: add promptCachedTokens to gemini-large model

### LINKED ISSUE #6016: Failed to get conversion rate for usage type: promptCachedTokens
Some requests show this error in the logs, which indicates an issue with the registry. My guess is that there was a rename somewhere, but some part of the code did not get updated properly.

voodoohop:
- Adds `promptCachedTokens: perMillion(0.2)` to `gemini-large` model
- Uses 10% of input price ratio (same as `gemini` and `gemini-search`)
- Fixes "Failed to get conversion rate for usage type: promptCachedTokens" error

Fixes #6016

### DIFF
--- a/shared/registry/text.ts
+++ b/shared/registry/text.ts
@@ -365,6 +365,7 @@ export const TEXT_SERVICES = {
             {
                 date: COST_START_DATE,
                 promptTextTokens: perMillion(2.0),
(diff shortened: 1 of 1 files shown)


voodoohop (commented):
@eulervoid Ready for review

PR #6082, opened 2025-12-16
feat: Add apiKeyExtended plugin for HTTP permissions updates

voodoohop:
- Add `apiKeyExtended` plugin that wraps better-auth's `apiKey` plugin
- Add `/api/auth/api-key/update-permissions` endpoint
- Calls base `updateApiKey` internally as server-side call to bypass `SERVER_ONLY` check
- Update client to use new endpoint for setting API key permissions

## Problem
better-auth's native `updateApiKey` endpoint marks `permissions` as `SERVER_ONLY`, meaning it can only be set via `auth.api.updateApiKey()` and not from HTTP requests.

## Solution
Create an extended plugin that adds a new endpoint which calls the base endpoint internally (as a server-side call), bypassing the restriction while maintaining better-auth's internal logic.

Based on approach suggested by @eulervoid

### DIFF
--- a/enter.pollinations.ai/src/client/routes/index.tsx
+++ b/enter.pollinations.ai/src/client/routes/index.tsx
@@ -123,14 +123,16 @@ function RouteComponent() {
             formState.allowedModels !== null &&
             formState.allowedModels !== undefined
         ) {
+            // Use our extended better-auth endpoint that allows permissions from HTTP
             const updateResponse = await fetch(
-                `/api/api-keys/${apiKey.id}/update`,
+                "/api/auth/api-key/update-permissions",
                 {
                     method: "POST",
                     headers: { "Content-Type": "application/json" },
                     credentials: "include",
                     body: JSON.stringify({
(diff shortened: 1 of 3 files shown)


enter.pollinations.ai/src/client/routes/index.tsx
```diff
@@ -123,14 +123,16 @@ function RouteComponent() {
             formState.allowedModels !== null &&
             formState.allowedModels !== undefined
         ) {
+            // Use our extended better-auth endpoint that allows permissions from HTTP
             const updateResponse = await fetch(
-                `/api/api-keys/${apiKey.id}/update`,
+                "/api/auth/api-key/update-permissions",
```
eulervoid, line 128: Not super important but we could use the better-auth client here if we infer the type. This should be relatively simple and return a type-safe response. Can be addressed in another PR though.

eulervoid (commented):
Looks very nice and clean

PR #6092, opened 2025-12-16
fix: Dashboard API Key display bugs (Prefix and Copy)

### LINKED ISSUE #6087: fix: Dashboard API Key display bugs (Prefix and Copy)
Reported by: `elliot8701`

There are two bugs/improvements needed in the API key section of the dashboard:

1.  **Remove `PLLN_` prefix**: We should remove the prepended `PLLN_` from the display so the beginning of the actual key is visible. This is more practical for users.
2.  **Fix Clickable Publishable Key**: The publishable key is currently not clickable, making it impossible to copy. This needs to be fixed.

---
**Author:** `elliot8701`

*Created via Discord*

voodoohop:
- Remove `plln_` prefix from new keys (now `pk_`/`sk_`)
- Restore `plaintextKey` storage for publishable keys via `auth.apiKey.update()`
- Show key prefix instead of suffix in dashboard display
- Configure 10 chars for `start` field via `startingCharactersConfig`

Fixes #6087

### DIFF
--- a/enter.pollinations.ai/src/auth.ts
+++ b/enter.pollinations.ai/src/auth.ts
@@ -27,6 +27,9 @@ export function createAuth(env: Cloudflare.Env) {
         enableMetadata: true,
         defaultPrefix: PUBLISHABLE_KEY_PREFIX,
         defaultKeyLength: 16, // Minimum key length for validation (matches custom generator)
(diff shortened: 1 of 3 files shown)


voodoohop:

PR #6096, opened 2025-12-16
Enable observablility with structured logging

eulervoid:
Enables cloudflare observability for the production environment, using structured logging.
- Output log lines as json in production (cloudflare parses and indexes this natively)
- Use logger category consistently to provide context (instead of manually adding e.g. `[PROXY]` to the log message)
- Add a script that parses json log lines from stdin and outputs formatted text logs
    - Useful for doing `wrangler tail --env production | tsx format-logs.ts`)
    - Still needs updating as wrangler seems to change its output format when piped to another process


voodoohop:

Issue #6083, opened 2025-12-16
fix: Add pre-request estimated cost tracking to prevent parallel request abuse


voodoohop:
## Problem

PR #6025 added local pending spend tracking to prevent negative balances, but it only tracks **completed** events. Users can still abuse the system by making many parallel requests:

1. User has 1 pollen balance
2. User fires 100 parallel video requests simultaneously
3. All requests pass balance check (balance shows 1 pollen for all)
4. All requests complete and record their cost
5. User ends up with massive negative balance (-$600 in extreme cases)

## Root Cause

The current `getRecentSpend()` function only sums `totalPrice` from **completed** events:

```typescript
const result = await db
    .select({
        total: sql<number>`COALESCE(SUM(${event.totalPrice}), 0)`,
    })
    .from(event)
    .where(
        and(
            eq(event.userId, userId),
            eq(event.isBilledUsage, true),
            gte(event.createdAt, windowStart),
        ),
    );
```

This doesn't account for requests that are **in-flight** but haven't completed yet.

### Phase 1: Pre-request Estimated Cost Tracking

1. **Get average cost per model from Tinybird**
   - Create endpoint: `GET /api/analytics/avg-cost-per-model`
   - Returns average `totalPrice` per model from historical data
   - Cache this data (refresh every hour)

2. **Add "pending" event type to D1**
   - New event status: `pending_estimate` (before request starts)
   - Store `estimatedCost` field based on model average

3. **Track pending requests**
   - **On request start**: Insert pending event with `estimatedCost`
   - **On request complete**: Update event with actual `totalPrice`
   - **On request error**: Remove pending event or mark as failed

4. **Update balance check**
   ```typescript
   const getEffectiveSpend = async (userId: string): Promise<number> => {
       const result = await db
           .select({
               total: sql<number>`COALESCE(SUM(
                   CASE 
                       WHEN event_status = 'pending_estimate' THEN estimated_cost
                       ELSE total_price
                   END
               ), 0)`,
           })
           .from(event)
           .where(
               and(
                   eq(event.userId, userId),
                   eq(event.isBilledUsage, true),
                   gte(event.createdAt, windowStart),
               ),
           );
       return result[0]?.total || 0;
   };
   ```

### Phase 2: Model-specific Minimum Balance (Optional)

For expensive models, require a minimum balance before allowing requests:

```typescript
const MODEL_MIN_BALANCE = {
    'veo': 0.15,        // Video model
    'seedance': 1.0,    // Video model
[...]

PR #6097, opened 2025-12-16
feat: pre-request estimated cost tracking

voodoohop:
- Add `pending_estimate` event status for in-flight requests
- Add `estimatedCost` column to event table
- Insert estimate event BEFORE request, UPDATE after completion
- Include estimated costs in balance calculation (`getRecentSpend`)
- Filter `pending_estimate` from Polar/Tinybird delivery
- Self-healing: stale estimates expire after 10min window

Addresses #6083


enter.pollinations.ai/src/middleware/track.ts
```diff
@@ -120,6 +122,67 @@ export const track = (eventType: EventType) =>
...
+                    userGithubUsername: c.var.auth.user?.githubUsername,
+                    apiKeyId: c.var.auth.apiKey?.id,
+                    apiKeyType: c.var.auth.apiKey?.metadata
+                        ?.keyType as ApiKeyType,
+                    apiKeyName: c.var.auth.apiKey?.name,
+                    modelRequested: requestTracking.modelRequested,
+                    resolvedModelRequested:
+                        requestTracking.resolvedModelRequested,
+                    modelProviderUsed: requestTracking.modelProvider,
+                    isBilledUsage: true, // Assume billable, will be updated
```
eulervoid: I think the default should be false here.

voodoohop:

eulervoid (commented):
I like the general idea, but i'm not sure about updating the events, in my head they were immutable. But i guess we already update them when they change status and its fine, but i'd want fields like `totalCost` to be undefined while the event is incomplete to make it really clear that this is whats going on, and to make more modular functions and types. Maybe i could have a go at this and try to sort things a bit?

Issue #5879, opened 2025-12-11
[Feature Request] JSON Schema Enforcement for Chat Completions (Structured Output)


Narvarthus:
## Summary
Add support for guaranteed JSON schema compliance in chat completion responses, similar to OpenAI's `response_format` parameter.

## Use Case
Many applications need LLMs to return structured data that can be reliably parsed:

- **Chatbots**: Extract intents, entities, and actions from user input
- **Data extraction**: Pull structured fields from unstructured text
- **Agent workflows**: Tool calling with predictable response shapes
- **Content generation**: Articles with title, body, tags, metadata

Currently, developers include schemas in prompts and hope the model complies. This works ~85% of the time, but failures require expensive retry logic and error handling.

## Proposed API

POST /v1/chat/completions
{
  "model": "openai",
  "messages": [
    {"role": "system", "content": "Extract product info from the text"},
    {"role": "user", "content": "The new iPhone 16 Pro costs $999 and has 256GB storage"}
  ],
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "product_info",
      "strict": true,
      "schema": {
        "type": "object",
        "properties": {
          "product_name": {"type": "string"},
          "price": {"type": "number"},
          "specs": {"type": "object"}
        },
        "required": ["product_name", "price"]
      }
    }
  }
}

## Benefits

Reliability: 100% parseable responses, no regex/string parsing hacks
Simplified Code: Remove try/catch retry loops and validation logic
Agent Workflows: Critical for tool-calling architectures where downstream code expects specific shapes
Reduced Costs: No retry loops = fewer API calls

## Alternatives Considered

Prompt engineering: Works mostly, but not guaranteed
Post-processing with another LLM call: Expensive and slow
Client-side schema validation + retry: Adds latency and complexity

## Additional Context

OpenAI's implementation: https://platform.openai.com/docs/guides/structured-outputs
Anthropic supports this via tool_use
This is becoming table-stakes for LLM APIs in 2024/2025

## Priority

High — This is a key blocker for building reliable AI-native applications.

ElliotEtag:
Closing as stale — no fresh reproduction in 4+ months and a prior comment asked for current details. Please open a new issue with the current endpoint, model, request, response body, and timestamp if this is still happening.

ElliotEtag:
[...]

PR #6130, opened 2025-12-17
Proper JSON Schema

fricker2025-star:
**Core Implementation Files:**
- src/middleware/json-schema.ts - Main JSON schema enforcement middleware
- src/middleware/json-schema-validation.ts - Response validation middleware
- src/routes/proxy.ts - Integration into the chat completion pipeline Key Features Implemented:

**1. JSON Schema Detection** : The middleware detects response_format requests with:
   
   - type: "json_object" - Forces valid JSON object responses
   - type: "json_schema" - Validates against specific JSON schemas

**2. System Prompt Injection** : Automatically injects system prompts that force the LLM to:
   
   - Respond with valid JSON only
   - Strictly adhere to the provided JSON schema
   - Avoid hallucinating extra text like "Here is your JSON:"
   - Include all required properties with correct types

3**. Response Validation** : Validates both streaming and non-streaming responses against:
   
   - Basic JSON object structure
   - Complex JSON schemas with properties, types, enums, patterns
   - Numeric constraints (minimum/maximum values)
   - String length constraints
   - Required property validation
4. **Error Handling** : Provides detailed validation errors when responses don't match the schema Example Usage:
```
{
  "model": "gpt-4",
  "messages": [{"role": "user", 
  "content": "Generate a user 
  profile"}],
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "user_profile",
      "schema": {
        "type": "object",
        "properties": {
          "name": {"type": 
          "string"},
          "age": {"type": "number", 
          "minimum": 0},
          "email": {"type": 
          "string", "pattern": "^
          [^@]+@[^@]+$"}
        },
        "required": ["name", 
        "email"]
      }
    }
  }
}
```

#5879

### DIFF
--- a/enter.pollinations.ai/src/middleware/json-schema-validation.ts
+++ b/enter.pollinations.ai/src/middleware/json-schema-validation.ts
@@ -116,8 +116,13 @@ async function handleNonStreamingValidation(
 /**
  * Handle validation for streaming responses
  * 
- * THE FIX: Properly buffers incomplete lines to prevent JSON parse errors
- * when TCP packets split in the middle of SSE data lines.
+ * FIXES APPLIED:
+ * 1. Line buffering: Prevents JSON parse errors when TCP packets split SSE data lines
+ * 2. Multi-byte character handling: Reuses TextDecoder with stream:true for emojis/Unicode
+ * 
+ * NOTE: Validation happens at the END of the stream (when finish_reason appears).
+ * This means invalid JSON is still streamed to the client, but an error is appended.
+ * This trade-off prioritizes streaming speed/UX over preventing invalid data transmission.
  */
 async function handleStreamingValidation(
     c: any,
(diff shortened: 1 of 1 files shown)


voodoohop:
Thanks for working on #5879! 🌸

This is a legitimate feature request and the implementation looks substantial. A few notes:

**Positives:**
- References an actual issue (#5879)
- Has a proper PR description
- The code is well-documented

**Concerns:**
- This is a large feature (300+ lines of new middleware) - needs careful review
- The unrelated `api-key.tsx` change should be in a separate PR
- Need to verify this works with streaming responses

Leaving open for maintainer review. @eulervoid might want to take a look at this since it touches `enter.pollinations.ai`.

eulervoid:
Thanks for the PR, i generally think it would be cool to support this, but i'm not sure if a middleware in enter is the right place for that code. It seems its a text service concern, if not actually something that i would expect from the portkey gateway.

I had a quick look and it seems that portkey actually does support this for a lot of models. So maybe its just a matter of communicating which models support this and which don't?

I would suspect that implementing this ourselves would involve a lot of small details, edge cases and bugs, especially with streaming responses across all the different models, so if we can use any off-the-shelf solution we should.

fricker2025-star:
Im looking at what i can fix for you guys now @eulervoid

fricker2025-star:
I see json schema didnt need to be added Closing PR

PR #6134, opened 2025-12-17
negative balance is no more

fricker2025-star:


voodoohop:
Thanks for your interest in contributing! 🌸

This PR appears to be a subset of #6140 which was already closed. The atomic balance deduction and micro-pollen precision changes are the same.

Please see the feedback on #6140 - major architectural changes like adding a balance column and changing pricing precision need to be discussed in an issue first.

Closing as duplicate of #6140.

PR #6135, opened 2025-12-17
Comprehensive CSP for all

fricker2025-star:


voodoohop:
Thanks for the CSP middleware contribution! 🌸

The code looks reasonable, but could you please add a PR description explaining:
1. What problem this solves (is there a security issue we should know about?)
2. Why CSP is needed specifically for `/api/docs/*` routes
3. Any testing you've done to verify it doesn't break the Scalar docs

Also, the unrelated change to `api-key.tsx` (`undefined` → `""`) should be in a separate PR.

Leaving open for now pending description.

mohityadav8:
Hi @fricker2025-star ! 👋
Following up on this PR — a maintainer requested some clarification about the purpose of the CSP changes and testing details.

Could you update the description and also rebase to resolve the current conflicts?
After that it should be ready for review 🙂

PR #6139, opened 2025-12-17
Refactor tier update and frontend code

eulervoid:
Refactors the tier routes and frontend:
- Remove duplicate code, add more structure
- Removes unused code/routes
- Extract helper functions in separate utils files 
- Fix long standing frontend `t-o-j-s-o-n` bug (was caused by trying to serialize the auth client)
- Simplify frontend auth setup by not using RouterContext to pass clients, but import directly

### DIFF
--- a/enter.pollinations.ai/src/client/components/tier-panel.tsx
+++ b/enter.pollinations.ai/src/client/components/tier-panel.tsx
@@ -1,10 +1,6 @@
 import { useEffect, useState, type FC } from "react";
 import { TierExplanation } from "./tier-explanation";
 import { TierName, TierStatus } from "../../utils/polar.ts";
-import type {
-    TiersRoutes,
-    TierSubscriptionStatus,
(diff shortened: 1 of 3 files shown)


enter.pollinations.ai/src/client/components/tier-panel.tsx
```diff
@@ -24,7 +20,7 @@ const TierContainer: FC<{ children: React.ReactNode }> = ({ children }) => (
 );
 
 const BetaNotice = () => (
-    <div className="bg-linear-to-r from-gray-100 to-slate-100 rounded-xl p-4 border border-gray-300 mt-3">
+    <div className="bg-gradient-to-r from-gray-100 to-slate-100 rounded-xl p-4 border border-gray-300 mt-3">
```
eulervoid, line 23: Why do you change this back? My tailwind lsp tells me this is deprecated, is it wrong?

PR #6161, opened 2025-12-18
Fix GPT-5 temperature forcing for Azure OpenAI models

voodoohop:
- Force `temperature=1` for GPT-5 series models (Azure only supports this value)
- Check resolved model name (`options.model`) instead of `requestedModel`
- Add integration test verifying `temperature: 0.7` requests succeed

Fixes 400 error: `"temperature" does not support 0.7 with this model`


voodoohop:

PR #6162, opened 2025-12-18
Updated tier and client cleanup refactor

eulervoid:
#### Fixed version of reverted #6139 

- **Refactor after tier changes: remove unused code, better structure, dynamic tier to productId mapping, fix frontend to-j-s-o-n error**
- **Fix TypeScript error, typos, and unused imports in tier refactor**
- **fix: Use iso string dates in TierSubscriptionDetailsSchema to fix api docs**

The issue that was breaking the docs was that i used a `z.date()` in `TierSubscriptionDetailsSchema`, which could not be tranformed to json schema. I don't think this PR was involved in any issues with `gen.pollinations.ai` as it does not contain any routing changes @voodoohop.


voodoohop:

voodoohop (commented):
@eulervoid why is there only one product id in the wrangler.toml for each env?

e.g.
```
POLAR_PRODUCT_TIER_SPORE = "19fa1660-a90c-453d-8132-4d228cc7db39"
```

but no other product ids for the other tiers

eulervoid:
The other ones get loaded from polar via the metadata slug field. I left this one in because it gets used as the default tier and i didn't want to mess with that. It should be removed too in another pr.

I'll do some of claudes suggestions, then we can merge.

PR #6166, opened 2025-12-18
Fix/csp security clean

fricker2025-star:
fix/billing-usage-tracking-clean - This is about fixing a billing loophole where streaming requests could be exploited. The key changes are:
     - Added enhanced billing protection for streaming responses
     - Implemented fallback billing using content length estimation when actual usage data isn't available
     - Ensured billing always happens even if streams are interrupted or clients disconnect mid-stream
     - Added minimal fallback charges to prevent infinite free usage exploits


mohityadav8:
Thanks for the fix 👍

The PR currently has merge conflicts with `main`.
Could you rebase/update the branch so reviewers can properly evaluate it?
@fricker2025-star

PR #6201, opened 2025-12-19
Further optimize D1 performance

eulervoid:
I looked at the most impactful queries via: 
`npx wrangler d1 insights DB --sort-by=reads --sort-type=sum --limit=10 --env=production`

And made some fixes for the most impactful queries:
- Add another index on (event_status, created_at) which is often queried
- Don't count the pending events when deciding if a batch is ready to process, instead just check if there are at least `minBatchSize` events by using a subquery with `OFFSET ${minBatchSize - 1}` and checking if the row exists
- Add a small constant delay between successful batches to avoid rate limiting


voodoohop:

PR #6210, opened 2025-12-19
fix: handle unencoded nested URLs in image parameter

### LINKED ISSUE #6208: gen.pollinations.ai: nested image URLs cause NaN validation errors
## Problem
When using image-to-image with nested URLs in the `image` parameter, the API returns validation errors:

```json
{"fieldErrors":{"width":["Invalid input: expected number, received NaN"],"image":["Invalid input: expected string, received array"]}}
```

## Root Cause
- Nested image URLs contain unencoded `&` and `?` characters
- Query parser interprets them as top-level parameters
- Results in multiple conflicting values for `width`, `height`, `seed`, `image`

## Example
```
?image=https://gen.pollinations.ai/image/prompt?model=nanobanana&width=2048...
```
The `&width=2048` inside the nested URL is parsed as a top-level param.

## Fix Options
1. **Documentation**: Clarify that nested URLs must be fully URL-encoded
2. **API improvement**: Better error message explaining the encoding requirement
3. **Smart parsing**: Detect and handle nested URL patterns

## Affected Endpoint
- `gen.pollinations.ai/image/*` → `enter.pollinations.ai/api/generate/image/*`

Itachi-1824:
When users pass a nested URL in the `image=` param (like another gen.pollinations.ai URL), the `&` and `?` in it mess up query parsing.

**Fix:**
- Gateway now auto-detects and encodes nested URLs by looking for duplicate params
- Added a clear error message if it still fails: "Put image= param last in your URL, or URL-encode it"

Works for most cases now without users having to manually encode anything.

Resolves #6208

### DIFF
--- a/gen.pollinations.ai/src/index.ts
+++ b/gen.pollinations.ai/src/index.ts
@@ -22,11 +22,12 @@ interface Env {
 // Strategy: collect all params, find duplicates, keep outer ones and encode nested URL
 function fixNestedImageUrls(rawUrl: string): string {
     const imageMatch = rawUrl.match(/([?&])(image=)(https?:\/\/)/i);
-    if (!imageMatch) return rawUrl;
+    if (!imageMatch || imageMatch.index === undefined) return rawUrl;
 
-    const nestedUrlStart = imageMatch.index! + imageMatch[1].length + imageMatch[2].length;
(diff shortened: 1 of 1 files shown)


Itachi-1824:
polly can you review it thoroughly & concisely

Itachi-1824:
polly looks good now?

PR #6248, opened 2025-12-20
feat: add request deduplication middleware

voodoohop:
## Summary
- Adds request deduplication middleware to prevent duplicate concurrent requests
- Wired into all generation endpoints in enter.pollinations.ai

## Changes
- Fix Cloudflare Workers I/O limitation (serialize response data instead of sharing Response objects)
- Wire middleware into proxy routes after auth/rate-limit/polar
- Add `X-Cache: HIT` + `X-Cache-Type: DEDUP` headers for deduplicated responses
- Add test suite with realistic 5ms delays between requests

## How it works
- Uses method + URL + body hash as deduplication key
- When identical requests arrive concurrently, only first executes
- Others wait and receive same response with `X-Cache: HIT` header
- Auto-cleanup when request completes

## Tests
- ✅ Concurrent identical requests deduplicated (verified via X-Cache header)
- ✅ Different requests processed independently
- ✅ Sequential requests work after cleanup

## Use cases
- User reloads during long-running generation (Veo video ~90s)
- Prevents duplicate billing for same request
- Cache stampede prevention

### DIFF
--- a/enter.pollinations.ai/src/middleware/requestDeduplication.ts
+++ b/enter.pollinations.ai/src/middleware/requestDeduplication.ts
@@ -143,8 +143,13 @@ async function createRequestKey(c: {
             if (body) {
                 parts.push(body);
             }
-        } catch {
+        } catch (error) {
             // If body reading fails, just use URL (graceful degradation)
+            // This can happen with malformed requests or stream issues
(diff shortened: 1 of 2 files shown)


voodoohop:

PR #6276, opened 2025-12-21
[Feat] Attach a Feed Route to the hello.pollinations.ai 

Circuit-Overtime:
## Attach a Feed Section to the Pollinations AI Dashboard at hello.pollinations.ai/

- Attaches a FEED Section 
- Allows user to directly watch / download / visit community feed 
- Lightweight and fits theme 

> Updated to the latest version of the npm packages

### DIFF
--- a/hello.pollinations.ai/src/ui/pages/Feed.tsx
+++ b/hello.pollinations.ai/src/ui/pages/Feed.tsx
@@ -172,7 +172,7 @@ function Feed() {
                                         loading="lazy"
                                         className="w-full h-full object-cover group-hover:scale-105 transition-transform duration-300"
                                     />
-                                    <div className="absolute inset-0 bg-gradient-to-t from-black/80 via-transparent to-transparent opacity-0 group-hover:opacity-100 transition-opacity duration-300 flex flex-col justify-end p-4">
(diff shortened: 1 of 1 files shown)


Circuit-Overtime:
This is how it looks like -- 

image

image


> Preview of the page on desktop view

ElliotEtag:
Hey feed was already implemented. Have you seen?

ElliotEtag:
Let’s try to keep focus on issues that are validated, there is so much to do!

Circuit-Overtime:
> Hey feed was already implemented. Have you seen?

Oops I might have missed it!! Can you please share me the URL. This was with me for a long time had forgotten to raise a PR

ElliotEtag:
image

Circuit-Overtime:
@ElliotEtag I like that! But technically we would miss the watch button even in that case cause of the UI .. Also there's some advantage to the one I have added 

- Users can watch / download / get details on the prompt + other details behind it 
- They can stop and watch feed if they wish to 
- Adds a decentralized one 

> If this is overkill I would just ask If I can implement a button atl-east to switch to the feed or self it would be a bit diff to spot that on the hello page as part of the UI (just my thought ok rest all awesome Elliot)

ez-vivek:
Having a `/feed` directory on hello.pollinations is preferable because the `/play` endpoint is primarily intended for a different purpose—serving as a dashboard for users to generate content. This distinction was overlooked in the legacy dashboard implementation, FEED having its own directory would be better.

PR #6297, opened 2025-12-21
feat: pre-request cost tracking to prevent parallel request abuse

voodoohop:
## Problem

Users can accumulate large negative balances by firing many parallel requests before any costs are deducted. The current system only tracks costs after request completion, creating a window for abuse.

## Solution

Insert a `pending_estimate` event **before** the request starts, using historical model statistics to estimate the cost. This estimated cost is included in `getPendingSpend()` calculation, preventing users from overspending.

## Changes

- Add `pending_estimate` event status for in-flight requests
- Add `estimatedPrice` field to event schema
- Create `model-stats.ts` utility to fetch avg prices from Tinybird (cached 1hr in KV)
- Insert `pending_estimate` event BEFORE `await next()` in track middleware
- Update `pending_estimate` event with actual data after completion
- Modify `getPendingSpend()` to sum `estimatedPrice` for pending_estimate events
- Filter `pending_estimate` events from Polar/Tinybird delivery
- Self-healing: stale estimates expire after 10min window

## Flow

```
1. Request arrives → Insert pending_estimate with estimatedPrice
2. getPendingSpend() includes estimatedPrice in balance check
3. Request completes → Update event with actual cost, clear estimatedPrice
4. Stale pending_estimate events auto-expire after 10 minutes
```

Addresses #6083

### DIFF
--- a/enter.pollinations.ai/test/integration/text.test.ts
+++ b/enter.pollinations.ai/test/integration/text.test.ts
@@ -120,6 +120,10 @@ describe("POST /generate/v1/chat/completions (authenticated)", async () => {
                 expect(event.tokenCountCompletionText).toBeGreaterThan(0);
                 expect(event.totalCost).toBeGreaterThan(0);
                 expect(event.totalPrice).toBeGreaterThanOrEqual(0);
+                // Regression test: selectedMeterSlug must be captured AFTER next()
+                // If this is null, balanceTracking was captured before the balance check middleware ran
+                expect(event.selectedMeterSlug).toBeDefined();
+                expect(event.selectedMeterSlug).not.toBeNull();
             });
         },
     );
@@ -172,6 +176,9 @@ describe("POST /generate/v1/chat/completions (streaming)", async () => {
                 expect(event.tokenCountCompletionText).toBeGreaterThan(0);
                 expect(event.totalCost).toBeGreaterThan(0);
                 expect(event.totalPrice).toBeGreaterThanOrEqual(0);
+                // Regression test: selectedMeterSlug must be captured AFTER next()
+                expect(event.selectedMeterSlug).toBeDefined();
+                expect(event.selectedMeterSlug).not.toBeNull();
             });
         },
     );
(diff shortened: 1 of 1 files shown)


voodoohop:
@claude can you review concisely especially i light of recent changes?

eulervoid:
I was updating this a bit to create separate functions for creating and updating the events and i'm a bit worried about the fact that when we insert an estimate event, we have to fill the event with dummy values for a lot of fields. The correct type for an extimate event would look something like this:

```ts
export type EstimateGenerationEvent = Omit<
    InsertGenerationEvent,
    // Response fields - not available until request completes
    | "endTime"
    | "responseTime"
    | "responseStatus"
    | "modelUsed"
    // Usage fields - actual token counts unknown until completion
    | "tokenCountPromptText"
    | "tokenCountPromptAudio"
    | "tokenCountPromptCached"
    | "tokenCountPromptImage"
    | "tokenCountCompletionText"
    | "tokenCountCompletionReasoning"
    | "tokenCountCompletionAudio"
    | "tokenCountCompletionImage"
    | "tokenCountCompletionVideoSeconds"
    | "tokenCountCompletionVideoTokens"
    // Cost fields - calculated after completion
    | "totalCost"
    | "totalPrice"
    // Cache fields - response not cached yet
    | "cacheHit"
    | "cacheType"
    | "cacheSemanticSimilarity"
    | "cacheSemanticThreshold"
    | "cacheKey"
    // Moderation fields - response not processed yet
    | "moderationPromptHateSeverity"
    | "moderationPromptSelfHarmSeverity"
    | "moderationPromptSexualSeverity"
    | "moderationPromptViolenceSeverity"
    | "moderationPromptJailbreakDetected"
    | "moderationCompletionHateSeverity"
    | "moderationCompletionSelfHarmSeverity"
    | "moderationCompletionSexualSeverity"
    | "moderationCompletionViolenceSeverity"
    | "moderationCompletionProtectedMaterialCodeDetected"
    | "moderationCompletionProtectedMaterialTextDetected"
    // Error fields - no error yet
    | "errorResponseCode"
    | "errorSource"
    | "errorMessage"
>;
```
But a lot of these values have a `NOT NULL` constraint in our current schema (mainly endTime, tokenCount* and totalPrice/totalCost). I'm thinking how we should handle this. With the counts and cost/price, i guess it could be 0 by default, but at least the endTime should not get some dummy value but we should remove the not null constraint to clearly indicate that this request has not been completed yet. What do you think @voodoohop ?

eulervoid:
I added some updates and merged main into this branch to fix the conflicts. Its ready to merge i'd say.

voodoohop:
@claude there was quite a serious bug before my last commit in that there was a problem with the sequence of events that was causing a meter slug to be null and the balance not applied. can you just verify if it was fixed in a nice way?

PR #6319, opened 2025-12-21
perf: Optimize Polar subscription lookup

voodoohop:
- Optimize `userUpdateTier` command in manage-polar.ts
- Instead of iterating through ALL subscriptions checking each email, first lookup customer by email (fast), then filter subscriptions by customerId (also fast)
- Reduces lookup time from potentially minutes to under a second

**Before:** Paginated through all subscriptions (100 at a time), checking each customer's email
**After:** 2 targeted API calls - customer lookup by email, then subscriptions by customerId

Tested and verified working.


voodoohop:

voodoohop:
Addressed review feedback:
- Changed `limit: 10` → `limit: 1` (users typically have 1 active subscription)
- Added `active: true` filter to the API call instead of filtering in JS
- Removed duplicate `status !== "active"` check (now handled by API filter)

PR #6387, opened 2025-12-23
fix: Return registry pricing fields directly in /models API

### LINKED ISSUE #6385: fix: Sync and correct pricing for ALL models (Text, Image, Video, Audio)
**Description**
The pricing returned by the API (e.g., `gen.pollinations.ai/image/models`) for `nanobanana-pro` and potentially other models appears to be incorrect/outdated compared to the actual costs reflected on the UI (`enter.pollinations.ai`).

For example, `nanobanana-pro` is reported by the API as `0.00012` (cheap), while it is actually one of the most expensive models.

**Task**
- Audit and sync pricing for **ALL** models (Text, Image, Video, Audio).
- Ensure the API endpoints return the correct, up-to-date pricing that matches the billing/UI logic.
- Verify `shared/registry/image.ts` and other registry files are accurate.

**Reported by**
_dr_misterio_ via Discord

voodoohop:
Fixes #6385

- Return registry field names directly (`completionImageTokens`, `promptTextTokens`, etc.) instead of renamed fields (`image_price`, `input_token_price`)
- Filter out zero/undefined values automatically
- Removes ambiguity between per-token vs per-image pricing (e.g., `nanobanana-pro` now clearly shows `completionImageTokens` is per-token, not per-image)

**Before:**
```json
{ "image_price": 0.00012 }
```

**After:**
```json
{ "completionImageTokens": 0.00012 }
```


voodoohop:

PR #6437, opened 2025-12-24
Cleaned up mono-repo root for stray files

Circuit-Overtime:
## Changes made 

- Removed some unused .md files 
- Removed lefthook.yml since no pre commit validation is being done we have agentic system for that via PRs
- Removed the cursor/claude/contrib config 
> We are doing it from the pollinations.ai/community page and the editor configs shall be .gitignore ed, also we don't use the `all-contributors-cli` 
- Edited the Readme to remove blank contributor section

@voodoohop please check this once it would keep the monorepo way cleaner

### DIFF
--- a/.sops.yaml
+++ b/.sops.yaml
@@ -1,0 +1,4 @@
+creation_rules:
+    - path_regex: (\.env$|\.encrypted\.env$|\.vars.json$|env\.json$)
+      age: >-
+        age1dd8ammekey2t9tzvvznwwx9m35q7jyxvzaydwqjawy99hgmdp3ms9556l5
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md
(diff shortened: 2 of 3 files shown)


ElliotEtag:
hmm i see many files deleted that would i think absolutely break all the service (check nix: https://nix.dev/manual/nix/2.26/command-ref/new-cli/nix3-develop)

also PROJECT.md can be removed, now it lives in apps/APPS.md so that'd be redondent.

let's imporve and get another review

Circuit-Overtime:
Okk elliot! Also are we using nix? I'll then just add it back and remove projects that's it. I had asked thom about pre commit he said he doesn't know and i didn't find any usage of it in the repo itself, and rest editor configs I don't think we need to keep! I'll make another commit here in this branch itself and then you can review again ok 💯

Circuit-Overtime:
@ElliotEtag check once now please

voodoohop:

PR #6443, opened 2025-12-25
feat: add reference image URL input for image-to-image

voodoohop:
- Add multi-image URL input with thumbnails and remove buttons
- Use pipe separator for multiple images (API format)
- Show real API error messages instead of generic ones
- Remove nologo parameter and unused imports


voodoohop:

PR #6455, opened 2025-12-25
feat: Fix Gemini thinking mode (enable/disable logic)

### LINKED ISSUE #6450: [Bug]: Thinking parameter ignored in Chat Completions API for Gemini reasoning models
🐛 Description

I have encountered a critical issue with the API endpoint https://gen.pollinations.ai/v1/chat/completions when using reasoning-capable models (specifically gemini-3-flash-preview).

**The thinking parameter in the request payload appears to be completely ignored by the backend:**

- Setting thinking.type to "disabled" does not stop the model from generating reasoning tokens.

- Setting thinking.budget_tokens (e.g., to 100 or 1000) does not limit the reasoning output.

- Setting reasoning_effort: "low" (OpenAI standard) does not result in thinking_level: "minimal" behavior.

**As a result, the model likely defaults to thinking_level: "high" (default for Gemini 3), causing massive over-computation.**

🔄 Steps to Reproduce
Send a POST request to the chat completions endpoint with the following payload structure:

```
curl -X POST "https://gen.pollinations.ai/v1/chat/completions" \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer YOUR_SECRET_TOKEN" \
  -d '{
    "model": "gemini",
    "messages": [
      {
        "role": "user",
        "content": "Explain the theory of relativity in simple terms but with mathematical proofs."
      }
    ],
    "thinking": {
      "type": "disabled",
      "budget_tokens": 100
    },
    "reasoning_effort": "low",
    "temperature": 0.6,
    "stream": false
  }'
```

📉 Expected Behavior

- Mapping: The backend should detect the intent to limit thinking and map it to the Vertex AI parameter thinking_level="minimal" (or "low"), which is the only valid way to constrain Gemini 3.0 Flash.

- Latency: The response should be fast (standard generation speed, <5s).

- Outcome: Reasoning tokens should be minimal (<1000) or close to zero.

📈 Actual Behavior
The model enters a deep reasoning mode regardless of the settings.

- Latency: ~40-50 seconds (instead of expected <5s).

- Reasoning Tokens: Extremely high counts (e.g., 4000+ or 7000+).

- Cost: Significantly higher due to unwanted hidden tokens.

📋 Logs / Evidence
Here is a log from a recent transaction where I attempted to limit/disable thinking. 

[Timestamp-Model-InputTokens-OutputTextTokens-OutputReasoningTokens-CostUSD-Timems.csv](https://github.com/user-attachments/files/24339863/Timestamp-Model-InputTokens-OutputTextTokens-OutputReasoningTokens-CostUSD-Timems.csv)

> _Critical Note: output_reasoning_tokens reached 7050, despite the thinking parameter being explicitly disabled/limited in the request._

🕵️ Analysis

**The issue likely stems from an incorrect parameter mapping in the Vertex AI / Gemini adapter for the 3.0 model family..**

According to recent Google Vertex AI documentation:
- Gemini 3.0 deprecates thinking_budget: Unlike Gemini 2.5, Gemini 3.0 models ignore the token budget parameter. Passing a budget of 1 or 1000 has no effect.
- New Control Parameter: Gemini 3.0 requires the thinking_level parameter (values: "low", "high", or "minimal" for Flash).
- Hypothesis: The Pollinations backend is likely receiving my request (either budget_tokens or reasoning_effort: low) and translating it into a thinking_budget integer. Since Gemini 3.0 ignores this integer, it falls back to its default behavior, which is thinking_level="high", causing the excessive token generation.

**Suggested Fix:**

**The backend logic needs to be updated for gemini-3-* models to map reasoning_effort: low (or thinking.type: disabled) to thinking_level="minimal" or thinking_level="low" instead of sending a token budget.**

andarcanum:
**Summary**

  I have implemented a fix to ensure the backend correctly respects the thinking parameter sent by clients, specifically
  targeting Google Gemini models.

  1. What was done:
   * Updated `parameterValidators.js`: Added support for parsing the OpenAI/Anthropic-style thinking object (e.g.,
     thinking: { type: "disabled" }). This is now normalized into an internal thinking_budget integer.
   * Created `createGeminiThinkingTransform.ts`: A new transformation middleware that translates the internal
     thinking_budget into the specific thinking_config required by Google's Vertex AI API.
   * Implemented Model-Specific Strategies: Since different Gemini versions handle "thinking" differently, I implemented
     tailored strategies for disabling it:
       * Gemini 2.5: Uses thinking_budget: 0 (Explicitly disables thinking).
       * Gemini 3 Flash: Uses thinking_level: "MINIMAL" (Minimizes latency, closest to "off").
       * Gemini 3 Pro: Uses thinking_level: "LOW" (Most economical setting, as full disable is not supported).
   * Updated `availableModels.ts`: Integrated the new transform into the pipeline for all relevant Gemini models
     (gemini, gemini-fast, gemini-large, gemini-search), assigning the correct strategy to each.

  2. Why this will work:

 Previously:

 **The server ignored the client's request to disable thinking because it didn't recognize the thinking**

  Now:

    1. The client's intent (disabled) is captured.
    2.  It is translated into the exact API parameters (include_thoughts: false, thinking_budget: 0, etc.) documented by
    Google for reducing reasoning costs and latency.

 3. How it works:
  When a request comes in with "thinking": { "type": "disabled" }:

-       Validation Layer: Converts this to thinking_budget: 0.

-       Transformation Layer: The createGeminiThinkingTransform sees the 0 budget and the model type (e.g., "v3-flash").

-       API Request Construction: It constructs the final payload for Google, e.g., for Gemini 3 Flash:

        "thinking_config": {
         "include_thoughts": false,
          "thinking_level": "MINIMAL"
        }

-       Result: The model executes with minimal or no reasoning, saving tokens and time.

4. Testing:
  I created and executed a dedicated test script (test-thinking-fix.js) that verified:
   * Correct parsing of the input thinking object (enabled/disabled states).
   * Correct generation of the thinking_config output for all three strategies (v2.5, v3-flash, v3-pro).
   * Confirmed that disabled input results in the correct "off" signals for each model version.
### Related Issue
Closes #6450


voodoohop:
Thanks for the contribution @andreykolesnikov! 

I've created an alternative implementation in #6472 that uses **Portkey-compatible format** instead of raw `thinking_config`. Key differences:

| Aspect | This PR | #6472 |
|--------|---------|-------|
| Output format | Raw `thinking_config` | Portkey `thinking` object |
| `include_thoughts` | Used | Removed (not valid Vertex AI param) |
| Gemini 3 handling | `thinking_level` | `reasoning_effort` (Portkey maps this) |
| Tests | None | 3 integration tests |

Per [Portkey docs](https://portkey.ai/docs/integrations/llms/gemini), the gateway expects:
```python
thinking={ "type": "enabled", "budget_tokens": 0 }  # to disable
reasoning_effort="none"  # alternative for Gemini 3
```

Portkey then translates these to Vertex AI's native format internally. Creating `thinking_config` directly may be ignored.

Feel free to review #6472 - I've credited you as co-author since the core idea came from your PR!

voodoohop:
I've also created a PR directly to your branch that you can merge: https://github.com/andreykolesnikov/pollinations/pull/1

Once you merge it, your PR here will automatically update with the fixes + tests! 🎉

voodoohop:
Hey @andreykolesnikov! 👋

First off, **huge thanks** for identifying this issue and putting together the initial fix! Your analysis of the Gemini thinking modes (v2.5 vs v3-flash vs v3-pro) was spot-on and really helped us understand the problem. 🙌

We've merged #6472 which uses the Portkey-compatible format (as discussed in the comments above). The key insight was that Portkey expects `reasoning_effort` and `thinking: { type, budget_tokens }` rather than raw `thinking_config` - the gateway handles the translation to Vertex AI's native format internally.

**You're credited as co-author on the merged commit** since the core idea and model-specific strategies came from your work here! 🎉

Closing this in favor of the merged PR, but seriously - thanks for the contribution and for being so responsive to feedback. Hope to see more PRs from you! ✨

PR #6464, opened 2025-12-26
feat: filter /models endpoints by API key permissions

voodoohop:
- Add optional auth to `/v1/models`, `/image/models`, `/text/models`
- Filter returned models by `apiKey.permissions.models` when API key is provided
- Change `/v1/models` from proxy to use shared registry directly
- Update `GetModelsResponseSchema` to match OpenAI format


voodoohop:

voodoohop:
@claude /text/models endpoint (line 219) has try/catch but getTextModelsInfo() is synchronous — the error handling works but is unnecessary overhead can you fix this

PR #6511, opened 2025-12-28
fix: Streaming responses not working (buffered instead of real-time)

voodoohop:
Fixes streaming responses being buffered instead of delivered in real-time.

**What users experienced:** When using `stream: true`, responses arrived all at once instead of token-by-token.

**Cause:** Request deduplication middleware was buffering the entire response before returning it.

**Fix:** Skip deduplication for streaming requests so tokens flow immediately.


voodoohop:
@claude can you review diligently

voodoohop:
@claude what do you think now?

PR #6623, opened 2025-12-30
fix: increase Polar cache TTL to reduce API rate limits

voodoohop:
Reduces Polar API rate limit errors (429s) by caching customer meters longer.

**Changes:**
- Cache TTL: 8 min → 25 min
- Pending spend window: 10 min → 30 min (must exceed cache TTL for conservative balance estimates)

Works with existing `staleOnError` fallback from previous session.


voodoohop:

PR #6627, opened 2025-12-30
feat: Add local pollen balance tracking in D1

voodoohop:
- Add `pollenBalance` column to user table
- Add webhook handlers for `benefit_grant.cycled` and `order.paid` to increment balance
- Modify `requirePositiveBalance` to read from D1 instead of Polar API
- Add balance decrement after billable requests complete
- Lazy initialization for existing users (fetch from Polar once on first request)

**Why**: Reduces Polar API calls on hot path, avoids rate limits (300 req/min)

### DIFF
--- a/enter.pollinations.ai/src/routes/webhooks.ts
+++ b/enter.pollinations.ai/src/routes/webhooks.ts
@@ -259,9 +259,14 @@ async function handleOrderPaid(
 
     const units = benefit?.properties?.units;
     if (!units || units <= 0) {
-        log.debug("order.paid for user {userId} has no pollen credits", {
-            userId: externalId,
-        });
+        log.warn(
+            "order.paid for user {userId} has no pollen credits (product: {productId}, benefits: {benefitCount})",
+            {
+                userId: externalId,
+                productId: order.productId,
+                benefitCount: product.benefits?.length ?? 0,
+            },
+        );
         return;
     }
 


voodoohop:
@claude can you review diligently? we have to push such a substantial change to production very quickly because our users are experiencing a lot of problems due to rate limits. so take extra care to make sure everything will work as expected. it is important to not allow the users to go significantly negative as that could cost us. but the priority is that users dont receive errors.

voodoohop:
@claude can you review again after the changes i made? we purposefully didnt address tracking estimated spend for now...

voodoohop:
## Local Testing Complete ✅

All flows verified working locally:

| Test | Result |
|------|--------|
| Lazy init from Polar | ✅ Balance fetched and stored on first request |
| Local balance check | ✅ No Polar API call on subsequent requests |
| Balance decrement | ✅ Decrements after each billable request |
| Blocking at ≤ 0 | ✅ Returns 403 "Insufficient pollen balance" |
| 503 on Polar failure | ✅ Fails safely if Polar unavailable during lazy init |

### Test Run (1 pollen account)
- Started: 1.0 pollen
- After 8 images: -0.0752 (went negative on 8th)
- Request 9: **403 blocked** ✅

### Key Benefits
- Eliminates Polar API calls on hot path (balance checks)
- Self-healing: fewer lazy inits over time = less Polar load
- Webhooks will increment balance on purchases/renewals

Ready to merge.

voodoohop:
@claude can you just check this last point:




Now let me check the Polar SDK types to verify the webhook payload structure:

Let me check the BenefitGrantWebhook type to see the `properties` structure:

Let me check the `BenefitGrantMeterCreditWebhook` type to see the exact structure:

Now let me search for Polar docs on webhooks:









Let me search for more specific documentation on the `benefit_grant.cycled` event:







Based on my research of the local code and Polar docs, here's my analysis:

## Webhook Balance Increment Verification

### `benefit_grant.cycled` Handler

**Our code** (`@/Users/thomash/Documents/GitHub/pollinations/enter.pollinations.ai/src/routes/webhooks.ts:197-234`):
```typescript
const properties = payload.data.properties as {
    lastCreditedUnits?: number;
};
const units = properties.lastCreditedUnits;
```

**Polar SDK type** (`BenefitGrantMeterCreditProperties`):
```typescript
export type BenefitGrantMeterCreditProperties = {
  lastCreditedMeterId?: string | undefined;
  lastCreditedUnits?: number | undefined;  // ✅ Correct field name
  lastCreditedAt?: string | undefined;
};
```

**Polar docs confirm:**
- `benefit_grant.cycled` fires at subscription renewal
- For subscription products: "credited at the beginning of every subscription cycle period"
- `lastCreditedUnits` contains the amount credited

✅ **Our implementation is correct.**

### `order.paid` Handler

**Our code** (`@/Users/thomash/Documents/GitHub/pollinations/enter.pollinations.ai/src/routes/webhooks.ts:236-285`):
- Looks for `meter_credit` benefit on the product
- Extracts `units` from `benefit.properties.units`

**Polar docs confirm:**
- For one-time products: "credited once at the time of purchase"
- `order.paid` fires when payment completes

✅ **Our implementation is correct.**

### Missing: `benefit_grant.created` for initial subscription

Looking at the docs, I noticed that `benefit_grant.cycled` fires on **renewal**, but for the **initial grant** when a subscription is first created, Polar sends `benefit_grant.created`. Let me check if we handle that:

**Found an issue!** We're missing the `benefit_grant.created` handler. According to Polar docs:

- `benefit_grant.created` - fires when a benefit is **first granted** (initial subscription)
- `benefit_grant.cycled` - fires on subscription **renewal**

For new users, the initial pollen grant comes via `benefit_grant.created`, not `benefit_grant.cycled`. Let me add this handler:

Let me check the `benefit_grant.created` payload type to understand its structure:

The `benefit_grant.created` uses the same `BenefitGrantWebhook` data type as `benefit_grant.cycled`. However, looking more closely at the SDK types, I see that `benefit_grant.created` may not have `lastCreditedUnits` populated yet (it's the initial grant, not a cycle).

Let me check the Polar docs more carefully to understand when credits are actually applied:

Looking at the Polar docs, I see that for subscription renewals:
1. `order.created` → `order.paid` fires
2. `benefit_grant.cycled` fires when credits are applied

For **initial subscription creation**, the flow is different. Let me check if `benefit_grant.created` has the credits or if they come via `order.paid`:

Looking at the docs, when a customer purchases a product with a Meter Credits benefit:
- **Subscription products**: Credits are granted at the beginning of each cycle via `benefit_grant.cycled`
- **One-time products**: Credits are granted at purchase via `order.paid`

The key question is: does `benefit_grant.cycled` fire for the **initial** subscription cycle, or only on renewals?

Let me check the SDK webhooks file to see what events are validated:

Based on my research, I see that `benefit_grant.cycled` is specifically for **subscription renewals**, while `benefit_grant.created` fires for the **initial grant**. However, looking at the SDK types more carefully:

- `benefit_grant.created` uses `BenefitGrantWebhook` which is a union type
- For meter credits, it would be `BenefitGrantMeterCreditWebhook` with `properties` containing `lastCreditedUnits`

But wait - for the **initial** grant, `lastCreditedUnits` might not be populated since it's the "last credited" (implying a previous cycle). Let me think about this more carefully...

Actually, looking at our current implementation:
1. **New user signs up** → User created in D1 with `pollenBalance = null`
2. **First API request** → Lazy init fetches balance from Polar meters (which already has the credits from the initial grant)
3. **Subscription renewal** → `benefit_grant.cycled` fires → We increment local balance

The lazy init covers the initial grant case! When a new user makes their first request, we fetch their current balance from Polar meters, which already includes any credits from their subscription.

**However**, there's still a potential issue: if a user's subscription renews and they haven't made any requests yet (so lazy init hasn't happened), the `benefit_grant.cycled` webhook would try to increment a `null` balance. Let me check our handler:

Looking at line 226:
```typescript
pollenBalance: sql`COALESCE(${userTable.pollenBalance}, 0) + ${units}`,
```

✅ **This handles it correctly!** `COALESCE(null, 0) + units` = `units`

## Summary of Webhook Verification

| Scenario | Webhook | Our Handler | Status |
|----------|---------|-------------|--------|
| Subscription renewal | `benefit_grant.cycled` | ✅ Increments with `lastCreditedUnits` | Correct |
| One-time purchase | `order.paid` | ✅ Finds `meter_credit` benefit, extracts `units` | Correct |
| Initial subscription grant | Lazy init from Polar | ✅ Fetches current balance on first request | Correct |
| Null balance handling | Both handlers | ✅ Uses `COALESCE(balance, 0) + units` | Correct |

**The implementation is correct.** The lazy initialization covers the initial grant case, and `benefit_grant.cycled` handles renewals.

## Webhook Verification Complete ✅

**Reviewed code and Polar docs. Implementation is correct:**

| Scenario | Handler | How it works |
|----------|---------|--------------|
| **Subscription renewal** | `benefit_grant.cycled` | Extracts `lastCreditedUnits` from properties, increments balance |
| **One-time purchase** | `order.paid` | Finds `meter_credit` benefit on product, extracts `units` |
| **Initial subscription** | Lazy init | First API request fetches current balance from Polar meters (already includes initial credits) |
| **Null balance** | Both handlers | `COALESCE(pollenBalance, 0) + units` handles null gracefully |

**Key findings from Polar docs:**
- `benefit_grant.cycled` fires on subscription **renewal** with `lastCreditedUnits`
- For **initial** subscription grants, credits are already in Polar meters when user makes first request
- Our lazy init fetches from Polar meters, so initial credits are captured

**SDK types verified:**
- `BenefitGrantMeterCreditProperties.lastCreditedUnits` - correct field name ✅
- `WebhookBenefitGrantCycledPayload.data.properties` - correct path ✅

**No changes needed.** Ready to merge: https://github.com/pollinations/pollinations/pull/6627

PR #6632, opened 2025-12-30
fix: reduce Polar cache TTL and pending spend window

voodoohop:
Now that local D1 balance tracking is in place (PR #6627), we can reduce the conservative cache TTL and pending spend window values.

**Changes:**
- Cache TTL: 25 min → 2 min (Polar API only used for lazy init now)
- Pending spend window: 30 min → 3 min (D1 balance updates in real-time)
- Remove `staleOnError` option (no longer needed - local balance is source of truth)

**Why:** With local balance tracking, the Polar API cache is only used during lazy initialization for new users. The long TTL and staleOnError were workarounds for rate limiting that are no longer needed.


voodoohop:
@claude can you review. only mention things that need fixing

PR #6633, opened 2025-12-30
[PATCH]: URL parameter ordering for I2I image generation

Circuit-Overtime:
- Build query string manually with ordered params instead of dict
- Ensure image= parameter is LAST in query string (API requirement)
- Single-encode reference URLs to avoid double-encoding
- Maintains consistent parameter order for public URL generation

Fixes 400 BAD_REQUEST validation errors on I2I requests


voodoohop:
@Circuit-Overtime image parameter is not required to be last

Circuit-Overtime:
@voodoohop Aah no no I'm just experimenting with a setup! This is not the intention of the PR yet, I'm holding off cause I'm a bit busy.. it's actually to put together some workflows, I'll do some stress testing in this branch, that's the reason it's a draft PR

PR #6663, opened 2025-12-31
chore: remove pending spend from dashboard display

voodoohop:
- Remove pending-spend API call from dashboard loader
- Remove pendingSpend prop from PollenBalance component
- Simplify dashboard - D1 balance is already accurate after decrement


voodoohop:
@claude can you review concisely? talk more about the problems than whats good

PR #6672, opened 2025-12-31
fix: Gemini 3 Flash fast mode (thinkingBudget:0)

### LINKED ISSUE #6652: gemini 3 flash fast mode not working
The reasoning effort options none or minimal do not work. This is not a pollination bug because it is not handled equally in the official API. We just need to loosen up the fact that you forcibly blocked the option. The complete fast mode (no thinking) of gemini 3 flash is

thinkingConfig={ 
    thinkingBudget:0 
}

It only goes into full fast mode. I don't think you need to stop this. In fact, the Gemini 3 series can still have Thinking Tokens set at the same time, and in the case of Gemini 3 Pro, it works properly even with Thinking Token 128. I hope they don't force the option to be blocked.

At the bottom, I attach the code you are forcibly blocking.

```
export function createGeminiThinkingTransform(
    modelType: GeminiModelType = "v2.5",
) {
    return (messages: unknown[], options: Record<string, unknown>) => {
        const updatedOptions = { ...options };

        // Check if user explicitly requested thinking to be disabled
        // This can come from:
        // 1. thinking_budget: 0 (direct)
        // 2. thinking: { type: "disabled" } (Anthropic-style, parsed by parameterValidators)
        const thinkingBudget = updatedOptions.thinking_budget as
            | number
            | undefined;

        // Only modify if thinking_budget is explicitly set
        if (thinkingBudget !== undefined) {
            const isDisabled = thinkingBudget === 0;

            log(
                `Configuring thinking for ${modelType}. Budget: ${thinkingBudget}, Disabled: ${isDisabled}`,
            );

            if (isDisabled) {
                // User wants to disable thinking - use Portkey-compatible format
                switch (modelType) {
                    case "v3-flash":
                        // Gemini 3 Flash: Use reasoning_effort="none" (maps to MINIMAL)
                        // Note: MINIMAL still requires thought signatures
                        updatedOptions.reasoning_effort = "none";
                        break;
                    case "v3-pro":
                        // Gemini 3 Pro: Can't fully disable, use "low" for minimal thinking
                        updatedOptions.reasoning_effort = "low";
                        break;
                    case "v2.5":
                    default:
                        // Gemini 2.5: Use thinking object with budget_tokens: 0
                        // Portkey docs: { "type": "enabled", "budget_tokens": 0 } disables thinking
                        updatedOptions.thinking = {
                            type: "enabled",
                            budget_tokens: 0,
                        };
                        break;
                }
            } else {
                // User wants thinking enabled with specific budget
                // Only set for v2.5 models that support budget_tokens
                if (modelType === "v2.5") {
                    updatedOptions.thinking = {
                        type: "enabled",
                        budget_tokens: thinkingBudget,
                    };
                } else {
                    // For v3 models, map budget to reasoning_effort levels
                    // This is a rough approximation since v3 uses levels not tokens
                    if (thinkingBudget <= 1024) {
                        updatedOptions.reasoning_effort = "low";
                    } else if (thinkingBudget <= 4096) {
                        updatedOptions.reasoning_effort = "medium";
                    } else {
                        updatedOptions.reasoning_effort = "high";
                    }
                }
            }

            // Clean up the internal thinking_budget parameter
            // (it's not a valid OpenAI/Portkey parameter)
            delete updatedOptions.thinking_budget;

            log("Final options:", {
                thinking: updatedOptions.thinking,
                reasoning_effort: updatedOptions.reasoning_effort,
            });
        }

        return { messages, options: updatedOptions };
    };
}
```

voodoohop:
- Use `thinking.budget_tokens:0` for v3-flash (same as v2.5)
- `reasoning_effort:none` only maps to minimal thinking, not full fast mode
- Simplifies code by consolidating v3-flash with v2.5 case

Fixes #6652

### DIFF
--- a/text.pollinations.ai/transforms/createGeminiThinkingTransform.ts
+++ b/text.pollinations.ai/transforms/createGeminiThinkingTransform.ts
@@ -45,20 +45,17 @@ export function createGeminiThinkingTransform(
 
             if (isDisabled) {
                 // User wants to disable thinking - use Portkey-compatible format
+                // Portkey docs: { "type": "enabled", "budget_tokens": 0 } disables thinking
+                // This works for both Gemini 2.5 and Gemini 3 models
                 switch (modelType) {
-                    case "v3-flash":
-                        // Gemini 3 Flash: Use reasoning_effort="none" (maps to MINIMAL)
-                        // Note: MINIMAL still requires thought signatures
-                        updatedOptions.reasoning_effort = "none";
-                        break;
                     case "v3-pro":
                         // Gemini 3 Pro: Can't fully disable, use "low" for minimal thinking
                         updatedOptions.reasoning_effort = "low";
                         break;
+                    case "v3-flash":
                     case "v2.5":
                     default:
-                        // Gemini 2.5: Use thinking object with budget_tokens: 0
-                        // Portkey docs: { "type": "enabled", "budget_tokens": 0 } disables thinking
+                        // Gemini 2.5 and 3 Flash: Use thinking object with budget_tokens: 0
                         updatedOptions.thinking = {
                             type: "enabled",
                             budget_tokens: 0,


snz2:
It doesn't work. Middleware or intermediate gateway seems to ignore the corresponding options. Native API works in fast mode exactly.

snz2:
I tested it after 7 hours and it now works. With Think 0 tokens. But there seems to be some delay in Gemini calls on this platform. When you call the native API, it is called almost immediately. There may be circumstances in this part. No problem now.

PR #6680, opened 2025-12-31
fix: lazy init also checks Polar when D1 balance is zero

voodoohop:
- Previously lazy init only triggered when balances were NULL
- Now also triggers when both balances are 0
- Only updates D1 if Polar has higher balance (won't overwrite spent balance)
- Fixes issue where users with 0 D1 balance but pollen in Polar see 0.00

**Root cause**: User `yonetkege-pixel` had D1 balance of 0 but Polar showed 11 pollen. The lazy init didn't trigger because balances were 0 (not NULL).

### DIFF
--- a/.github/ISSUE_TEMPLATE/balance-problem.yml
+++ b/.github/ISSUE_TEMPLATE/balance-problem.yml
@@ -1,0 +1,21 @@
+name: Balance Problem
+description: Pollen not showing after purchase
+title: "Balance: "
+labels: ["BUG"]
+body:
+  - type: input
+    id: email
+    attributes:
+      label: Email
+      description: The email you used to sign up
+      placeholder: "your@email.com"
+    validations:
+      required: true
+
+  - type: textarea
+    id: description
+    attributes:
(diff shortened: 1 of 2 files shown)


voodoohop:
@claude can you review concisely? focus on problems

voodoohop:
## Review Summary

**Changes in this PR:**
1. Lazy init now triggers when `either balance is NULL` OR `both are 0` (was: only when both NULL)
2. Added debug logging for all edge cases

**Review points addressed:**
- ✅ **Edge case logging**: Added `log.debug` when `hasBothZero` but Polar also returns 0
- ✅ **Exception handling**: Not a regression - same behavior as before
- ✅ **Cache staleness**: 2min TTL is acceptable tradeoff

**Verified with production data:**
- 3 users from #6686 had this exact issue (D1=0, Polar=20)
- Manually fixed by resetting to NULL
- Once merged, this won't require manual intervention

voodoohop:
@claude can you review again? ok to go? anything that could be simpler?

PR #6733, opened 2026-01-02
feat: Add NOWPayments crypto integration with cryptoBalance

voodoohop:
## Summary
Adds NOWPayments as alternative crypto payment method with separate `cryptoBalance` tracking.

## Changes
- Add NOWPayments invoice creation endpoint (`/api/nowpayments/invoice/:pack`)
- Add crypto webhook handler with HMAC-SHA512 signature verification
- Add `cryptoBalance` field to D1 schema (separate from fiat `packBalance`)
- Add $1 crypto-only payment option
- Add fiat/crypto payment toggle in dashboard UI
- Update balance display to combine pack + crypto as "paid" pollen

## How it works
- **Crypto payments** → credit `cryptoBalance`
- **Fiat payments** → credit `packBalance`
- **Total** = `tierBalance + packBalance + cryptoBalance`

## Next steps after merge
- Run `npm run migrate:staging` / `npm run migrate:production`
- Configure NOWPayments IPN webhook URL in dashboard

### DIFF
--- a/enter.pollinations.ai/src/routes/nowpayments.ts
+++ b/enter.pollinations.ai/src/routes/nowpayments.ts
@@ -101,11 +101,11 @@ export const nowpaymentsRoutes = new Hono<Env>()
             const priceAmount = PACK_AMOUNTS[pack];
             const description = PACK_DESCRIPTIONS[pack];
 
-            // Build success/cancel URLs
+            // Build success/cancel URLs (use request origin for dev)
             const appUrl =
                 c.env.ENVIRONMENT === "production"
                     ? "https://enter.pollinations.ai"
-                    : "http://localhost:5173";
+                    : new URL(c.req.url).origin;
             const successUrl = `${appUrl}/?payment=success&method=crypto`;
             const cancelUrl = `${appUrl}/?payment=cancelled&method=crypto`;
 
@@ -117,8 +117,6 @@ export const nowpaymentsRoutes = new Hono<Env>()
 
             log.info("Creating invoice", {
                 baseUrl,
-                apiKeyLength: apiKey.length,
-                apiKeyPreview: `${apiKey.slice(0, 4)}...${apiKey.slice(-4)}`,
                 pack,
             });
 
--- a/enter.pollinations.ai/src/routes/webhooks-crypto.ts
+++ b/enter.pollinations.ai/src/routes/webhooks-crypto.ts
@@ -216,6 +216,29 @@ export const webhooksCryptoRoutes = new Hono<Env>().post(
             return c.json({ received: true, processed: false });
         }
 
+        // For partially_paid, only credit if user paid at least 90% of the expected amount
+        if (payload.payment_status === "partially_paid") {
+            const actuallyPaidFiat = payload.actually_paid_fiat ?? 0;
+            const expectedAmount = payload.price_amount;
+            const paidRatio = actuallyPaidFiat / expectedAmount;
+            if (paidRatio < 0.9) {
+                log.warn(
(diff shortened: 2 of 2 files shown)


voodoohop:

voodoohop:
## Alternative: Use Polar Events API Instead of `cryptoBalance`

Polar's Events API supports **negative values to grant credits** ([docs](https://polar.sh/docs/guides/grant-meter-credits-before-purchase)). This could simplify the implementation.

### Proposed Change

Replace D1 `cryptoBalance` update with Polar Events API call:

```typescript
// Current (PR #6733):
await db.update(userTable)
    .set({ cryptoBalance: sql`COALESCE(...) + ${pollenAmount}` })
    .where(eq(userTable.id, userId));

// Alternative:
await polar.events.ingest({
    events: [{
        name: "pack_usage",  // Must match pack meter filter
        externalCustomerId: userId,
        metadata: { units: -pollenAmount },  // Negative = grant
    }],
});
```

### Comparison

| Aspect | Current PR (`cryptoBalance`) | Polar Events API |
|--------|------------------------------|------------------|
| **D1 schema** | New `crypto_balance` column + migration | No changes |
| **Balance tracking** | 3 fields: tier + pack + crypto | 2 fields: tier + pack |
| **Source of truth** | Split (Polar + D1) | Unified (Polar) |
| **`getBalance()` changes** | Sum all 3 balances | No changes |
| **UI changes** | Show crypto separately | Credits appear in pack |
| **Polar dashboard** | Crypto invisible | All credits visible |
| **Latency** | D1 write (~5ms) | Polar API (~100-200ms) |
| **Failure mode** | D1 transaction | Need retry logic |
| **Rollback complexity** | Drop column | None |

### Pros of Polar Events API
- **Simpler**: No new D1 field, no migration, no balance aggregation changes
- **Single source of truth**: All credits in Polar
- **Better visibility**: Crypto grants appear in Polar dashboard
- **Fungible credits**: Pack balance = fiat + crypto combined

### Cons of Polar Events API
- **Latency**: API call slower than D1 write
- **Dependency**: Requires Polar API availability
- **Meter name**: Need to verify exact filter name for pack meter

### Recommendation
If crypto/fiat distinction in UI isn't required, Polar Events API is cleaner. If you need to show "crypto balance" separately, current PR approach works.

RomirJ:
sticking with the local d1 cryptoBalance instead of the polar events api definitely saves on latency, but the deduction logic (tier -> crypto -> pack) introduces a tricky distributed consistency problem.

since tierBalance likely comes from polar (external) and cryptoBalance is d1 (local), you can't wrap the deduction in a single atomic database transaction. if a user blasts concurrent generation requests, you might successfully read available balance from both sources simultaneously before the deductions commit, effectively double-spending the crypto credits.

i’m working on hikaflow to analyze exactly these kinds of distributed state risks and "split-brain" ledger issues that standard linters miss.

given the merge conflicts and the financial logic here, it might be worth running a quick consistency analysis on the branch before resolving: app.hikaflow.com

are you handling the concurrency via an optimistic lock version on the user row, or just relying on the d1 write speed to beat the race condition?

RomirJ:
removing the fiat value checks for partially_paid status (84ce78b) seems risky.

if the gateway callback confirms a partial payment (e.g. user underpaid by 99%), does your logic scale the credited cryptoBalance proportionally, or does it trigger the full pack credit? without strict amount validation, an attacker could intentionally underpay (send dust) to trigger the partially_paid webhook and receive full credits.

i'm building hikaflow to map out these specific payment state/value reconciliation gaps that standard tests often miss.

might be worth verifying the credit calculation logic on that partial path: app.hikaflow.com

are you calculating the pro-rated pollen amount based on actually_paid, or just treating partial success as full success?

voodoohop:
polly can you review? focus on the negative if there are any

PR #6735, opened 2026-01-02
feat: use Polar Events API for crypto credits

voodoohop:
## Summary
Refactors crypto payment integration to use Polar's Events API instead of a separate `cryptoBalance` field in D1.

## Changes
- Replace D1 `cryptoBalance` update with `polar.events.ingest` using negative units to grant credits
- Remove `cryptoBalance` from `getBalance`, `requirePositiveBalance` in middleware
- Remove `cryptoBalance` from API response and client UI
- Crypto credits now flow through Polar pack meter for unified balance tracking

## Why
- **Centralized balance tracking**: All pollen balances managed in Polar
- **Simpler architecture**: No separate D1 column for crypto purchases
- **Consistent with fiat flow**: Crypto credits behave like pack purchases
- **Better auditability**: All credits visible in Polar dashboard

## Technical Details
Uses Polar's Events API with negative `units` to grant credits:
```typescript
await polar.events.ingest({
    events: [{
        name: "crypto_credit",
        externalCustomerId: userId,
        metadata: {
            units: -pollenAmount, // Negative = grant credits
            pack,
            paymentId: String(payload.payment_id),
            paymentMethod: "nowpayments",
        },
    }],
});
```

## Note
This PR targets `feat/nowpayments-crypto-balance` (PR #6733) and should be merged into that branch first.

### DIFF
--- a/enter.pollinations.ai/src/client/components/pollen-balance.tsx
+++ b/enter.pollinations.ai/src/client/components/pollen-balance.tsx
@@ -3,15 +3,13 @@ import type { FC } from "react";
 type PollenBalanceProps = {
     tierBalance: number;
     packBalance: number;
-    cryptoBalance: number;
 };
 
 export const PollenBalance: FC<PollenBalanceProps> = ({
     tierBalance,
     packBalance,
-    cryptoBalance,
 }) => {
-    const paidBalance = packBalance + cryptoBalance;
+    const paidBalance = packBalance;
     const totalPollen = Math.max(0, tierBalance + paidBalance);
     const freePercentage =
         totalPollen > 0 ? (tierBalance / totalPollen) * 100 : 0;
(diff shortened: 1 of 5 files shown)


voodoohop:
## ⚠️ Trade-off: Crypto Orders Not Visible in Polar Dashboard

**Con:** Polar's Events API grants meter credits but does **not** create "orders" in the Polar dashboard. Crypto purchases won't appear in the Orders section alongside fiat purchases.

**Why this limitation exists:**
- Polar has no API to create orders programmatically without checkout
- Checkout requires user interaction (redirect → complete flow)
- Events API is designed for usage tracking, not order creation

**What we get:**
- ✅ Credits show in customer's meter balance
- ✅ Event history with metadata (`paymentMethod: "nowpayments"`, `pack`, `paymentId`)
- ✅ Single source of truth for customer balances
- ❌ No unified "Orders" view for crypto + fiat

**Alternatives considered:**
1. **Keep separate `cryptoBalance` in D1** - Works but splits balance tracking
2. **$0 product + 100% discount checkout** - Requires user interaction, defeats purpose
3. **Track crypto orders in separate D1 table** - Could add for internal reporting

**Recommendation:** The Events API approach is still valid if unified balance tracking is the priority. For order visibility, we'd need to maintain a separate crypto orders table or request this feature from Polar.

PR #6744, opened 2026-01-02
fix: skip cache when seed=-1 for random generation

voodoohop:
- Skip text cache when `seed=-1` in POST body or query params
- Skip image cache when `seed=-1` in query params
- Supports SillyTavern convention where `-1` means "random seed"

Fixes issue reported in pollen-beta channel where users got same results with `seed=-1`.


voodoohop:

PR #6746, opened 2026-01-03
fix: seed=-1 generates random images instead of identical ones

voodoohop:
## Summary
- Fix `seed=-1` to generate random images instead of identical ones
- Clean up unused files

## Changes
- **`image.pollinations.ai/src/params.ts`**: Fix `sanitizedSeed` to convert `-1` to random seed
- **`enter.pollinations.ai/test/integration/image.test.ts`**: Add test for seed=-1 cache bypass
- **Deleted**: `cacheGeneratedImages.ts` (unused)
- **Deleted**: `makeParamsSafe.js` (unused)
- **Removed**: Unused import from `index.ts`

## Problem
When users passed `seed=-1` (convention for "random"), the image service was passing `-1` directly to the generation API, resulting in identical images every time.

## Solution
Convert `seed=-1` to a random seed in the Zod schema preprocessor (`params.ts`), which is the central place where all image parameters are validated.


voodoohop:

PR #6764, opened 2026-01-03
fix: Text generation seed=-1 should generate random output

### LINKED ISSUE #6759: fix: Text generation seed=-1 should be random (parity with image gen)
**Problem:**
User reported that `seed=-1` in text generation produces identical/deterministic outputs, unlike image generation where it now randomizes (fixed in PR #6746).

**Analysis:**
- PR #6746 only applied the fix to `image.pollinations.ai`.
- `text.pollinations.ai` validates `seed` as a simple integer in `utils/parameterValidators.js` and passes it directly to model providers.
- Sending `seed: -1` to providers (like OpenAI) makes them use that specific static number as the seed, resulting in deterministic output.

**Solution:**
Update the text generation logic (likely in `text.pollinations.ai/utils/parameterValidators.js` or `requestUtils.js`) to:
1. Detect if `seed === -1`.
2. Replace it with a random integer (e.g., `Math.floor(Math.random() * 2**32)`).
3. OR remove the `seed` parameter entirely to let the provider handle randomness (though explicit random seed is usually safer for consistency).

**References:**
- Related PR: #6746 (Image gen fix)
- Reported by user: `aegis_plus` (Discord)

---
**Author:** `aegis_plus` (UID: `414108324649500673`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1456995767813144586/1456996119468052564)

voodoohop:
## Summary
- Fix `seed=-1` to generate random text instead of deterministic output
- Parity with image generation (PR #6746)

## Changes
- Add `validateSeed()` function in `text.pollinations.ai/utils/parameterValidators.js`
- Converts `seed=-1` to random seed (`Math.floor(Math.random() * 1000000)`)
- Use `validateSeed()` instead of `validateInt()` for seed validation

## Problem
When users pass `seed=-1` (convention for "random"), the text service was passing `-1` directly to model providers, resulting in deterministic output.

## References
- Fixes #6759
- Related: #6746 (image gen fix)

### DIFF
--- a/text.pollinations.ai/utils/parameterValidators.js
+++ b/text.pollinations.ai/utils/parameterValidators.js
@@ -27,6 +27,21 @@ export const validateInt = (value) => {
     return parsed;
 };
 
+// Maximum seed value supported by most LLM/image providers (32-bit integer)
+const MAX_SEED_VALUE = 4294967296; // 2^32
+
+/**
+ * Processes seed value, converting -1 to a random seed (parity with image generation)
+ * @param {*} value - Value to validate
+ * @returns {number|undefined} Processed seed or undefined
+ */
+export const processSeed = (value) => {
+    const seed = validateInt(value);
+    if (seed === undefined) return undefined;
+    // seed=-1 means "random" - generate a random seed
+    return seed === -1 ? Math.floor(Math.random() * MAX_SEED_VALUE) : seed;
+};
+
 /**
  * Validates boolean values including string representations
  * @param {*} value - Value to validate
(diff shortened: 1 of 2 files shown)


text.pollinations.ai/utils/parameterValidators.js
```diff
@@ -27,6 +27,18 @@ export const validateInt = (value) => {
...
+/**
+ * Processes seed value, converting -1 to a random seed (parity with image generation)
+ * @param {*} value - Value to validate
+ * @returns {number|undefined} Processed seed or undefined
+ */
+export const processSeed = (value) => {
+    const seed = validateInt(value);
+    if (seed === undefined) return undefined;
+    // seed=-1 means "random" - generate a random seed
+    return seed === -1 ? Math.floor(Math.random() * 1000000) : seed;
```
Simplereally: I believe most LLM and Image providers support standard 32-bit integers, and would support `4294967296` (2^32)  to maximize entropy. Won't `1000000` limit us to 1m of randomness?

(same for the recent update to `images` code)

voodoohop:

PR #6800, opened 2026-01-04
[FEAT]: Longcat Image model experimentation

Circuit-Overtime:
# LongCat Multi-GPU Inference Experiment

## Setup
- Python 3.10 venv, PyTorch 2.x, Diffusers latest.
- GPUs: 2 × Tesla V100 32 GB.
- Model cache: `model_cache`.
- Torch dtype: `bfloat16`.
- Attention slicing enabled: `pipe.enable_attention_slicing("max")`.

---

### 1. Baseline
- Run full pipeline on single GPU.
- Resolution: 768×768.
- Observed OOM at higher resolutions (≥768×1344).
- Tried lowering inference steps (6–25) → memory still high.

### 2. Attempted two-GPU pipeline
- Idea: Split model components across GPUs:
  - GPU0: `text_encoder + transformer` (diffusion)
  - GPU1: `vae` (decoder)
- Challenges:
  - LongCat automatically decodes latents → cannot intercept output easily.
  - Moving intermediate latents required patching pipeline.
  - Raw attempt caused shape mismatch errors (`conv_in` expected 16 channels, got 1).

### 3. Latent-level workaround
- Subclassed `LongCatImagePipeline` to return **latents instead of pixels**.
- Diffusion stage computed on GPU0 → latents moved to GPU1 → VAE decode.
- Post-processing on GPU1 → converted to PIL image.
- Result: Multi-GPU decoding works, but requires manual patching.

### 4. Memory optimization
- Used `bfloat16` for all modules.
- Enabled attention slicing.
- Could not reduce image resolution below 768×768 for desired quality.
- LoRA or adapters considered **for training/fine-tuning**, not for inference memory reduction.

---

## Conclusion
- Full LongCat cannot easily fit extremely large images on a single V100 without slicing.
- Two-GPU latent split is feasible by **patching pipeline to bypass auto-decoding**.
- True tensor/model parallelism requires **DeepSpeed / Megatron-LM**, which is non-trivial.
- LoRA helps with **fine-tuning**, but not inference memory.

### DIFF
--- a/image.pollinations.ai/longcat/latent_only.py
+++ b/image.pollinations.ai/longcat/latent_only.py
@@ -1,0 +1,47 @@
+import torch
+import torch.nn as nn
+from diffusers import LongCatImagePipeline
+
+
+def main():
+    device = torch.device("cuda:0")
+
+    pipe = LongCatImagePipeline.from_pretrained(
+        "meituan-longcat/LongCat-Image",
+        torch_dtype=torch.float16,
+    ).to(device)
+
+    # Memory optimizations (no INT8, no compile)
+    pipe.enable_xformers_memory_efficient_attention()
+    pipe.enable_attention_slicing("max")
+    pipe.set_progress_bar_config(disable=True)
+
+    with torch.inference_mode():
+        out = pipe(
+            prompt="two cats sitting on a bench in a park",
+            height=512,
+            width=512,
+            num_inference_steps=20,
+            guidance_scale=3.0,
+            enable_cfg_renorm=True,
+            output_type="latent",   # stop at latents
+        )
+
+    latents = out.images
+
+    # Print raw latent information
+    print("Latents produced successfully")
+    print(f"Shape        : {latents.shape}")
+    print(f"Dtype        : {latents.dtype}")
+    print(f"Device       : {latents.device}")
+    print(f"Min value    : {latents.min().item():.6f}")
+    print(f"Max value    : {latents.max().item():.6f}")
+    print(f"Mean         : {latents.mean().item():.6f}")
+    print(f"Std          : {latents.std().item():.6f}")
+
+    print("\nLatent slice [0, :5, :5]:")
+    print(latents[0, :5, :5])
+
+
+if __name__ == "__main__":
+    main()


Circuit-Overtime:
## Ok here's what I have got from experimenting with V100 

> Quantized Inference Failed and Sometimes Produced Blank Outputs


### Hardware and Precision Constraints (Root Cause)
The NVIDIA V100 (Volta, SM70) does **not support bfloat16** and has **no FP8** or modern Tensor Core pathways used by recent Diffusers defaults. Running LongCat with `torch_dtype=bfloat16` caused silent precision fallback and mixed execution paths, leading to severe slowdowns and numerical instability. This alone made timing targets unreachable and introduced non-deterministic behavior.

### Incorrect INT8 Quantization Assumptions
Initial attempts assumed the existence of `bitsandbytes.nn.Int8Linear`, which is **not part of the released bitsandbytes API**. This caused partial or failed module replacement. In several cases, only a subset of `nn.Linear` layers were quantized, leaving the transformer in an inconsistent mixed-precision state. This resulted in invalid activations propagating through attention blocks, sometimes collapsing latents to near-zero values.

### Unsafe Module Replacement Logic
Early quantization code attempted to replace nested `nn.Linear` layers using shallow `setattr` calls on the root module. This silently failed for deeply nested submodules inside the LongCat transformer. The result was a **structurally corrupted model graph**: some layers ran INT8, others FP16, and some were never executed as intended. This is a direct cause of occasional **blank or near-black images**, even in single-GPU inference.

### Scheduler–Pipeline Incompatibility
LongCatImagePipeline internally assumes schedulers that support **custom sigma schedules**. Switching to `DPMSolverMultistepScheduler` with `use_karras_sigmas=True` violates this assumption. The pipeline then enters an invalid timestep setup path, which either raises a hard exception or, in some runs, produces degenerate denoising trajectories that converge to flat latents.

### CFG Renormalization and Quantization Interaction
`enable_cfg_renorm=True` amplifies numerical error under INT8 + FP16 mixed precision. On V100, this occasionally caused classifier-free guidance to overshoot, pushing latents outside the VAE’s valid range. The downstream VAE decoder then produced uniform or near-uniform pixel outputs, perceived as blank images.

### VAE Handling and Latent Scaling Sensitivity
When decoding latents produced by a quantized transformer, even small scaling or shift mismatches (especially under unstable denoising) caused the VAE to saturate. This effect was magnified when the transformer output was numerically degraded, again resulting in blank images without explicit runtime errors.

### Final Resolution
The failures were resolved by:
- Enforcing **FP16-only execution** (no BF16) on V100
- Using **`bitsandbytes.nn.Linear8bitLt`** with correct recursive replacement
- Ensuring **complete and consistent quantization** of all transformer linear layers
- Disabling **Karras sigmas** (or switching schedulers) for LongCat compatibility
- Disabling **CFG renormalization** under INT8
- Preserving correct latent scaling before VAE decode

@ElliotEtag  @voodoohop These changes restored numerical stability, eliminated blank outputs, and enabled reliable sub-20-second 512×512 inference on V100. 

--- 

## This is why the `t2i_infer.py` works:- 

This implementation works because it stays within a **numerically safe and memory-conservative execution regime**, even though it relies on fragile internal assumptions.

### Key Reasons It Produces Correct Images

1. **bfloat16 Improves Numerical Stability**
   - Using `torch_dtype=torch.bfloat16` preserves FP32 exponent range.
   - This prevents latent underflow during CFG subtraction on V100.
   - Earlier FP16-based versions collapsed latents to near-zero, producing blank images.

2. **CFG Renormalization Is Enabled**
   - `enable_cfg_renorm=True` prevents conditional–unconditional drift.
   - This stabilizes latent magnitudes, making manual decoding viable.

3. **Manual Latent Unpacking Matches LongCat (for 512×512)**
   - LongCat currently emits patch-packed latents:
     ```
     (B, N_patches, 4 × C_vae)
     ```
   - At 512×512, patches form a square grid and map cleanly to 2×2 latent tiles.
   - The hardcoded unpacking logic happens to align with the current model layout.

4. **VAE Decoding Is Offloaded to a Second GPU**
   - Diffusion (UNet + text encoder) runs on `cuda:0`.
   - VAE decoding runs on `cuda:1`.
   - This avoids peak-memory overlap and prevents OOM on V100.

---

### Why This Version Is Significantly Slower

1. **Cross-GPU Latent Transfers**
   - Latents are moved from `cuda:0` → `cuda:1`, incurring PCIe/NVLink overhead.
   - This transfer happens at full latent resolution and dominates decode time.

2. **Manual Reshape and Decode Path**
   - Latent unpacking, scaling, and shifting are done in Python-side tensor ops.
   - This bypasses fused, optimized paths inside the pipeline.

3. **Attention Slicing Trades Speed for Memory**
   - `enable_attention_slicing("max")` lowers peak memory usage.
   - This increases kernel launches and reduces throughput.

4. **No Scheduler or Step-Count Optimization**
   - Uses a conservative step count (20) and default scheduler.
   - Faster solvers and lower step counts are not leveraged.

The behavior is **resolution- and model-version dependent** and should be treated as a safe-but-fragile workaround rather than a production-robust inference path.

Circuit-Overtime:
- Longcat on int8
- TwinAI Inference on l40s
- ltx-2 video 
- upscaler pipeline

PR #6832, opened 2026-01-04
feat: re-enable flux as separate model on io.net

voodoohop:
- Add flux back as separate model in shared/registry/image.ts
- Add flux config to image.pollinations.ai/src/models.ts
- Update MODELS.md documentation (flux + zimage under io.net)
- Remove flux from zimage aliases
- Rename callComfyUI to callSelfHostedServer
- Default image model remains zimage


voodoohop:

PR #6875, opened 2026-01-06
feat: update audio capability icons

Itachi-1824:
- Use 🎙️ (studio mic) for audio input instead of 👂 (ear)
- Add 🔊 (speaker) for audio output capability
- Update legend to match


voodoohop:

PR #6889, opened 2026-01-06
fix: display videos correctly in Play page and add model tooltips

voodoohop:
- Render `<video>` element for models with video output (e.g., Veo 3.1)
- Add `autoPlay`, `loop`, `muted` attributes for seamless video playback
- Show model description on hover via native tooltip


voodoohop:

PR #6893, opened 2026-01-06
fix: gptimage model ignores seed parameter (#6892)

### LINKED ISSUE #6892: bug: `gptimage` (GPT Image 1 Mini) ignores `seed` parameter
User reports that the model `gptimage` (GPT Image 1 Mini) ignores the `seed` parameter.

**Behavior:**
- Changing the `seed` value does not change the generated image.
- The output appears to be identical regardless of the seed provided (or stuck on a fixed seed).

**Expected:**
- Different seeds should produce different variations.
- Same seed should produce deterministic output (if supported).

**Context:**
- Model: `gptimage` (GPT Image 1 Mini)
- Reported by: `marcosfrgames08`

---
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1457246799994687619/1458216076323192999)

voodoohop:
Fixes #6892

**Root Cause:**
The `gptimage` model had a cache rule that forced all requests to use `seed=42` for cache key generation. This caused all gptimage requests with the same prompt to return the same cached image, regardless of the seed parameter provided.

**Changes:**
- Remove gptimage-specific cache rule in `enter.pollinations.ai/src/utils/image-cache.ts`

**Notes:**
- The Azure GPT Image API doesn't support the seed parameter for deterministic generation
- However, users expect different requests to produce different images
- Each unique request now generates a fresh image instead of returning cached results

### DIFF
--- a/enter.pollinations.ai/src/utils/image-cache.ts
+++ b/enter.pollinations.ai/src/utils/image-cache.ts
@@ -9,40 +9,13 @@ import { removeUnset } from "@/util.ts";
 import { Logger } from "@logtape/logtape";
 
 /**
- * Apply model-specific caching rules to the URL
- * @param {URL} url - The URL object to transform
- * @returns {URL} - The transformed URL object
- */
-function applyModelSpecificRules(url: URL): URL {
-    // Get the model parameter
-    const model = url.searchParams.get("model");
-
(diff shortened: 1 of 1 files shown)


voodoohop:

PR #6915, opened 2026-01-07
Add whisper model to registry and update APPS.md stars

voodoohop:
- Add whisper-1 model to text registry with audio input modality
- Update APPS.md with refreshed GitHub stars
- Add app-update-stars.js script for star count updates


ElliotEtag:
@voodoohop we have this too https://www.assemblyai.com/

PR #6971, opened 2026-01-09
feat: use metadata.pollenBudget for API key budget tracking

voodoohop:
## Summary
- Store `pollenBudget` in metadata field instead of better-auth's `remaining` field
- Use SQLite JSON functions for atomic budget decrements
- Combine Expires and Budget columns into single "Limits" column in UI
- Fix table grid layout (7 columns, not 8)

## Why direct DB updates instead of better-auth?

⚠️ **This is a work in progress** - we tried using `auth.api.updateApiKey()` but encountered issues:

1. **Caching layer complexity** - better-auth uses secondary storage (Cloudflare KV) as cache with `fallbackToDatabase: true`. Direct DB updates cause stale cache.

2. **Context availability** - `authClient` from `c.var.auth.client` wasn't reliably available in tracking middleware context.

3. **Atomic updates** - SQLite JSON functions (`json_set`, `json_extract`) allow atomic decrements without read-modify-write, which is more reliable for concurrent requests.

The direct DB approach works because:
- `requireKeyBudget` check reads fresh from DB (not cache) for budget validation
- Atomic SQLite JSON update prevents race conditions
- Simpler and more reliable

## Files changed
- `src/middleware/track.ts` - Budget deduction using SQLite JSON functions
- `src/middleware/auth.ts` - Read budget from metadata
- `src/routes/api-keys.ts` - Store budget in metadata
- `src/client/components/api-key.tsx` - UI updates, combined Limits column

### DIFF
--- a/enter.pollinations.ai/src/client/components/api-key.tsx
+++ b/enter.pollinations.ai/src/client/components/api-key.tsx
@@ -129,10 +129,10 @@ const LimitsBadge: FC<{
     if (hasBudget) {
         const budgetStr =
             pollenBudget <= 0
-                ? "0🌸"
+                ? "0"
                 : pollenBudget < 1
-                  ? `${pollenBudget.toFixed(2)}🌸`
-                  : `${Math.floor(pollenBudget)}🌸`;
+                  ? pollenBudget.toFixed(2)
+                  : `${Math.floor(pollenBudget)}`;
         parts.push(budgetStr);
     }
 


voodoohop:
Added follow-up commit to remove the pollen emoji (🌸) from budget numbers in the API keys table for cleaner display.

**Testing locally confirmed:**
- Budget deduction working via SQLite JSON functions
- User balance and API key budget both decremented correctly
- Cache MISS requests trigger proper deduction

**Known limitation:** Some existing API keys have corrupted metadata (double-serialized JSON) from a previous bug - these keys won't have budget tracking until metadata is fixed or new keys are created.

voodoohop:
Closing in favor of #7294 which uses D1 `pollen_balance` column directly instead of metadata. The metadata approach had caching issues with better-auth's secondary storage.

PR #7005, opened 2026-01-10
fix: consolidate model-stats cache to single source

voodoohop:
## Problem

Dashboard pricing shows incorrect "1 pollen ≈" values on production vs local because two files were using the same KV cache key `model-stats` with different formats:

| File | Purpose | Format Stored |
|------|---------|---------------|
| `routes/model-stats.ts` | Dashboard API | Raw Tinybird `{ data: [...] }` |
| `utils/model-stats.ts` | Pre-request cost estimation | Wrapped `{ value: {...}, ttl }` via `cached()` |

When `track.ts` middleware ran first (on any API request), it populated the cache with the wrapped format. Then `/api/model-stats` returned that wrapped format which the client couldn't parse.

## Solution

- Reuse `getModelStats()` from `utils/model-stats.ts` instead of duplicate caching
- Transform output to format client expects: `{ data: [{ model, avg_cost_usd }] }`
- Single 1h TTL cache shared with pre-request cost estimation

## Changes

- Remove duplicate Tinybird URL and caching logic from `routes/model-stats.ts`
- Import and use `getModelStats` from `utils/model-stats.ts`

## Post-deploy

Clear production KV cache key `model-stats` to remove stale data:
```bash
npx wrangler kv key delete model-stats --namespace-id=eb4ea5149a5a413db53902bf2b8c1d95 --remote
```

### DIFF
--- a/enter.pollinations.ai/src/utils/model-stats.ts
+++ b/enter.pollinations.ai/src/utils/model-stats.ts
@@ -1,10 +1,12 @@
+import { cached } from "@/cache";
 import type { Logger } from "@logtape/logtape";
 
-const TINYBIRD_URL =
+const TINYBIRD_MODEL_STATS_URL =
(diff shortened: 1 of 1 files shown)


voodoohop:

voodoohop:

Issue #7003, opened 2026-01-10
nova-fast - does not work


jkeks:
second day **nova-fast** model - does not work, is it RIP ?

Circuit-Overtime:
please check the #7009 fixes this issue?

PR #7009, opened 2026-01-10
[PATCH]: Sanitized model names + alias to stop 4xx errors on gen.pollinations.ai

Circuit-Overtime:
## The Problem- 

- The issue #7004  and #6977 and #7003 mentions the problem with working on the `nova-micro` model by amazon 
    1. User requests: "nova-micro"
    2. OLD TEXT.TS SCHEMA (BEFORE FIX):
       Only allowed: openai, openai-fast, nova-fast, glm, minimax
       REJECTS nova-micro = 400 Bad Request
    
    3. IF nova-fast HAD passed validation:
       Gateway: Accepts "nova-fast" 
       Backend: availableModels.ts doesn't have aliases field
       Lookup fails because:
          - availableModels[].name is "nova-fast"
          - availableModels[].aliases is UNDEFINED
          - Code checks: m.aliases?.includes(requestData.model)
          - Result: 404 Not Found
    4. ROOT CAUSE: availableModels.ts is missing alias metadata

- Secondary unrelated bug at the `availableModels.ts` at line 192 we have `resolveServiceId` function which takes in only 1 param but were being passed 2

## Fixes Made:- 

- Added logic to accept both the serviceID and the alias name for the models while validating it for gen.*
- Fixed the extra param problem
- Added the logic to accept model alisas array for ModelRegistry
- Added comprehensive alias for the text.* for all 23 models

### DIFF
--- a/shared/registry/text.ts
+++ b/shared/registry/text.ts
@@ -366,8 +366,8 @@ export const TEXT_SERVICES = {
         search: true,
         isSpecialized: false,
     },
-    "kimi-k2-thinking": {
-        aliases: ["kimi-k2", "kimi-thinking"],
+    "kimi": {
+        aliases: ["kimi-k2", "kimi-reasoning", "kimi-k2-thinking"],
         modelId: "moonshotai/kimi-k2-thinking-maas",
         provider: "google",
         cost: [


voodoohop:
Hey @circuit-overtime! 👋 Thanks for digging into this!

Good catch on the text schema fix - that's the real issue. The aliases in `availableModels.ts` aren't needed though since they're already defined in `shared/registry/text.ts` and `resolveServiceId()` handles the lookup. I know it's a bit confusing how it's structured - some legacy code makes the flow harder to follow 😅

I've pushed a minimal fix to your branch that:
- ✅ Keeps the text.ts schema fix (the actual fix)
- ✅ Fixes the `resolveServiceId()` param (good catch!)
- ❌ Removes the duplicate aliases in availableModels.ts

Also fixed the test file signature while investigating - the tests were passing but using an outdated 2-param call.

Let me know if you have questions!

Circuit-Overtime:
@voodoohop Welcome Thom, I have checked through the codes it's all good. Can you please push your commit (`Also fixed the test file signature while investigating - the tests were passing but using an outdated 2-param call.`)

voodoohop:
@Circuit-Overtime check it and feel free to merge if you think it's ok

Circuit-Overtime:
@voodoohop yes all good 💯, merging now Thom 👍

PR #7018, opened 2026-01-10
[PATCH]: Fix the `discord-post-weekly-news` trigger 

Circuit-Overtime:
## Change made:- 

- only YYYY-MM-DD.md formatted files will trigger the weekly news workflow now 
- or else editing any other .md like LINKS.md would falsely trigger the workflow


Circuit-Overtime:
polly check this once

PR #7032, opened 2026-01-11
feat: re-enable flux as separate model on io.net

voodoohop:
- Re-add `flux` as separate model in registry (zimage remains default)
- Fix Dockerfile: add curl, set SM 8.9 targets for RTX 4090, use ninja
- Fix nunchaku import path in server.py
- Standardize port to 8765
- Rename `callComfyUI` → `callSelfHostedServer`


voodoohop:

PR #7039, opened 2026-01-11
fix: update claude-code-action inputs for app-review workflow

ElliotEtag:
- `direct_prompt` → `prompt` (valid input name)
- `allowed_tools` → `claude_args` with `--allowedTools` flag
- `timeout_minutes` → `--max-turns 30`
- Add `allowed_non_write_users` for external submitters

Fixes the workflow error where external users without write permissions couldn't trigger app reviews.


ElliotEtag:
polly does this PR fixes https://github.com/pollinations/pollinations/actions/runs/20895486485 ?

PR #7040, opened 2026-01-11
fix: add label_trigger to app-review workflow

ElliotEtag:
- Add `label_trigger: TIER-APP` so the action recognizes issues with this label as valid triggers
- The action defaults to looking for `@claude` trigger which was causing 'No trigger was met' errors


ElliotEtag:
https://github.com/pollinations/pollinations/actions/runs/20895486485/job/60033875985

Our previous fix for this workflow issue hasn't worked. Here's another fix. What do you think? polly

PR #7072, opened 2026-01-12
[REFACTOR]: App Category is now just AI dependent

Circuit-Overtime:
Yooo!! This makes a small change to the tier app submission issue template by removing the "App Category" dropdown field. This simplifies the submission process for users. This is in accordance to what @voodoohop suggested


Circuit-Overtime:
polly check if any other changes are needed in order to push this through

PR #7087, opened 2026-01-12
feat: replace claude code action with lightweight python agent

Itachi-1824:
- Replaces claude-code-action with simple Python script (~260 lines)
- Uses Pollinations API directly with SKILL.md as system prompt
- AI generates emoji, category, description, and error responses
- GitHub App token for all operations
- Issue author as co-author on commits

Triggers:
- Issue opened with `TIER-APP`
- Issue edited with `TIER-APP-INCOMPLETE`  
- Comments on `TIER-APP-INCOMPLETE`

Faster, simpler, full control.

### DIFF
--- a/.github/docs/APP-SUBMISSION.md
+++ b/.github/docs/APP-SUBMISSION.md
@@ -25,17 +25,18 @@ Community apps built with Pollinations.AI can be submitted for inclusion on [pol
 
 ## Workflow: app-review-submission.yml
 
-Uses Claude Code via router (deepseek model) with pre-validation.
+Lightweight Python agent using Pollinations API directly with `SKILL.md` as system prompt.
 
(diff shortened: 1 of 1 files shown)


Circuit-Overtime:
@Itachi-1824 I have added a project name sanitation layer, if any name comes in as a shell command (prompt injection) it would have been directly sent to the shell `umm like this app"; rm -rf /; echo "`. So a validation function is added. 

The rest of the logics and the file looks ok but the thing is if Thom decides to merge #7072 then the line 
`"category": sanitize_string(extract(r"### App Category\s*\n(.+?)(?:\n|$)"), 20),` will resolve out an error since category would no longer come from the issue template. We gotta ask him once about the final decision on this. Rest all good brother nice implementation man.

PR #7089, opened 2026-01-12
feat: add Claude Bedrock→Vertex AI fallback + fix cost tracking

voodoohop:
## Summary
- Adds Portkey fallback mechanism for Claude models
- AWS Bedrock as primary provider, Google Vertex AI as fallback
- Automatic failover if Bedrock fails
- Fixes cost tracking for fallback models

### text.pollinations.ai
- `modelConfigs.ts`: Add `claude-sonnet-4-5-fallback` and `claude-opus-4-5-fallback` configs
- `portkeyUtils.js`: Handle fallback configs with `x-portkey-config` header
- `availableModels.ts`: Switch `claude` and `claude-large` to use fallback configs
- `package.json`: Add Node.js 20.x constraint (tsx ESM bug in 22+)

### enter.pollinations.ai
- `track.ts`: Use service's canonical `modelId` for cost calculation instead of provider-returned model ID

## Why the cost fix?
Provider-specific model IDs (e.g. `us.anthropic.claude-sonnet-4-5-20250929-v1:0` from Bedrock) are not in `MODEL_REGISTRY`. The fix uses the service's canonical modelId (e.g. `claude-sonnet-4-5-20250929`) which is always present in the registry.

### DIFF
--- a/text.pollinations.ai/portkeyUtils.js
+++ b/text.pollinations.ai/portkeyUtils.js
@@ -60,51 +60,25 @@ export function extractApiVersion(endpoint) {
 }
 
 /**
- * Convert hyphenated keys to snake_case for Portkey config JSON
- * e.g., "aws-access-key-id" -> "aws_access_key_id"
- * @param {string} key - Hyphenated key
- * @returns {string} - Snake_case key
- */
-function toSnakeCase(key) {
-    return key.replace(/-/g, "_");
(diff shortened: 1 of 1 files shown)


voodoohop:
polly can you review concisely?

voodoohop:
polly can you review again?

PR #7112, opened 2026-01-12
feat: add SPORE tier check before app submission

ElliotEtag:
- Adds tier check during app submission validation
- Users with SPORE tier cannot submit apps
- Requires at least SEED tier (automatically granted based on GitHub activity)
- Fetches tier alongside registration check from D1 database

### DIFF
--- a/.github/docs/TIER-SYSTEM.md
+++ b/.github/docs/TIER-SYSTEM.md
@@ -163,8 +163,9 @@ flowchart TD
 
 | Test | Query/Check | Pass | Fail |
 |------|-------------|------|------|
-| **Registration** | `SELECT id, tier FROM user WHERE github_username = '${author}'` | User exists | → INCOMPLETE |
-| **Tier Check** | `tier !== 'spore'` | Seed or higher | → REJECTED + closed |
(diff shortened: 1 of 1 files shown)


ElliotEtag:
polly can you check pleasE?

ElliotEtag:
polly can you check again pleasE?
also the docs!

be meticulous please

ElliotEtag:
polly can you check again please i think i fixed what you highlighted

Itachi-1824 (changes requested):
hey! looked through the changes, couple things:

the string matching in the python agent (`"spore tier" in e.lower()`) feels a bit risky - if someone changes the wording in the ts file later this will just silently stop working. maybe pass an error code or something more structured?

also wondering about the null tier case - if a user has no tier set in the db, they'd still be able to submit since `null?.toLowerCase()` won't match "spore". not sure if that's what we want?

one more thing - mixing the indent reformatting with the actual logic changes made this harder to review than it needed to be. not a blocker but maybe split those next time?

the actual feature looks good though, and nice catch adding the username sanitization

ElliotEtag:
polly Did I address Itachi's concerns about the null tier?

PR #7230, opened 2026-01-13
feat: remove D1 event buffer, send directly to Tinybird

voodoohop:
- Remove `storeEvents()`, `updateEvent()`, `processEvents()`, `sendPolarEvents()`, `getPendingSpend()` from events.ts
- Replace buffer logic in track.ts with direct `sendToTinybird()` call
- Remove cron scheduled handler from index.ts and wrangler.toml
- Remove `/customer/pending-spend` endpoint from polar.ts
- Add migration to drop `event` table
- Rewrite events.test.ts for simplified functionality


voodoohop:
polly can you review? focus on the negatives if there are any

voodoohop:
polly can you revie again please?

PR #7237, opened 2026-01-13
fix: add READ token to polar_event datasource

voodoohop:
- Add `polar_event_read` READ token for querying the datasource
- Already deployed to Tinybird via `tb --cloud deploy`
- Enables SQL queries on polar_event data

### DIFF
--- a/enter.pollinations.ai/src/routes/webhooks.ts
+++ b/enter.pollinations.ai/src/routes/webhooks.ts
@@ -42,20 +42,31 @@ async function sendPolarEventToTinybird(
         return;
     }
 
-    // Extract fields from payload
+    // Extract fields from payload for analytics indexing
+    // Note: data_id is intentionally generic - it represents the primary entity ID
+    // (subscription_id, order_id, benefit_grant_id, etc.) depending on event_type.
+    // For detailed queries, use JSONExtract on the payload column.
     const p = payload as {
         type?: string;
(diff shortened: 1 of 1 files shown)


voodoohop:
## Review Feedback Addressed

**1. Sorting key mismatch** ⏳
- Valid point. Current sorting key (`timestamp, event_type, user_id`) optimizes time-series queries
- Adding `customer_id`, `product_id`, `data_id` would require schema migration
- For now, these fields enable filtering without JSON parsing; full scans acceptable given low event volume
- Can revisit if query patterns show need for indexing

**2. Incomplete type coverage** ✅ Fixed
- Added extraction from `subscription.productId` and `order.productId`
- Added fallback for `customerId` field at data level

**3. data_id semantics** ✅ Documented
- Added comment explaining `data_id` is intentionally generic (subscription_id, order_id, benefit_grant_id depending on event_type)
- Full payload stored for detailed queries via `JSONExtract`

**4. No backfill strategy** ✅ Already handled
- FORWARD_QUERY deployed to Tinybird backfills existing rows with empty strings
- Historical data queryable, just won't have extracted fields

**5. Duplicated field extraction** ℹ️ By design
- Webhook handler extracts for business logic (tier sync, balance updates)
- Tinybird extraction is for analytics indexing
- Different purposes, acceptable to have parallel paths

PR #7321, opened 2026-01-15
fix Cloudflare 524 timeout limitation for reasoning models

### LINKED ISSUE #7280: Currently, it is not possible to use models that take a long time, such as openai xhigh.
I am also concerned about the rules for when this symptom appears.

1. If the user's token is consumed, the user will not receive a response and the credit will be reduced.
2. If you consume the credits of the pollinations service itself, isn't it your loss?

{"error":"{"success":false,"error":{"message":"error code: 524","code":"UNKNOWN_ERROR","timestamp":"2026-01-13T11:57:50.124Z"},"status":524}"}

When are you going to fix it? In fact, openai-large's xhigh is unusable. But what I'm curious about is that if I send the same request to native API (openai.com), it doesn't take that long.
1. There is also a problem with the net response time taking a long time. If you make the same request to Native, you will receive a response within 1 minute.
2. It seems like you have set a limit on response time. After just 130 seconds, error 524 occurs.

#7211 I am re-inquiring because this answer did not work. Thank you for your understanding.

ElliotEtag:
## Analysis of Issue #7280

This is a **legitimate infrastructure limitation**, not a bug. Here's what's happening:

### Root Cause: Cloudflare 524 Timeout

**Error 524** = Cloudflare terminates the connection when the origin doesn't respond within ~100-120 seconds.

**Request flow:**
```
Client → Cloudflare CDN → gen.pollinations.ai (Worker) → enter.pollinations.ai → text.pollinations.ai (EC2) → Azure OpenAI
```

**The problem:**
- OpenAI reasoning models with `reasoning_effort: "high"` or higher (the user mentions "xhigh") can take **2-5+ minutes** to complete
- Cloudflare's proxy layer has a **hard 100 second timeout** that cannot be configured on Workers plans
- The request gets killed at Cloudflare's edge before the model finishes reasoning

### User's Concerns Are Valid

1. **Credits consumed but no response** - The request starts processing at Azure/OpenAI, tokens are consumed, but Cloudflare kills the connection before completion. This is a real problem.

2. **Works on OpenAI directly** - Correct, because direct API calls to OpenAI don't go through Cloudflare's proxy.

3. **130 seconds then error** - This matches Cloudflare's 524 timeout behavior (100s + some buffer).

### Possible Solutions

| Solution | Complexity | Notes |
|----------|------------|-------|
| **1. Use streaming** | Low | Stream keeps connection alive; user gets partial results. Already supported. |
| **2. Async job pattern** | High | Submit job → poll for status → retrieve result. Bypasses timeout entirely. |
| **3. Cloudflare Enterprise** | $$$ | Allows custom origin timeouts (up to 600s). |
| **4. Warn users / rate-limit xhigh** | Low | Document limitation; prevent accidental charges. |
| **5. Direct EC2 endpoint** | Medium | Bypass Cloudflare for long-running models (loses CDN benefits). |

### Recommendation

The **immediate fix** is to ensure users always use **streaming (`stream: true`)** for reasoning models. Streaming keeps the connection alive with incremental data.

For a proper fix, consider implementing an **async job queue pattern** for long-running models - this is how OpenAI's own API handles very long completions internally.

---

Fixes #7280

### DIFF
--- a/.github/docs/KNOWN-ISSUES.md
+++ b/.github/docs/KNOWN-ISSUES.md
@@ -1,0 +1,35 @@
+# Known Issues & Limitations
+
+## Cloudflare 524 Timeout for Long-Running Models
+
+**Affected models:** `openai-large` with high `reasoning_effort` (e.g., `xhigh`)
+
+**Issue:** Requests to reasoning models with high effort settings may fail with error 524 after ~100-130 seconds.
+
+**Root cause:** Cloudflare's proxy layer has a hard 100-second timeout that cannot be configured on Workers plans. OpenAI reasoning models with high effort can take 2-5+ minutes to complete.
+
+**Request flow:**
+```
+Client → Cloudflare CDN → gen.pollinations.ai → enter.pollinations.ai → text.pollinations.ai → Azure OpenAI
+```
+
+### Workaround
+
+**Use streaming** (`stream: true`) for reasoning models. Streaming keeps the connection alive with incremental data:
+
+```bash
+curl 'https://gen.pollinations.ai/v1/chat/completions' \
+  -H 'Authorization: Bearer YOUR_API_KEY' \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "model": "openai-large",
+    "messages": [{"role": "user", "content": "Your prompt"}],
+    "stream": true,
+    "reasoning_effort": "high"
+  }' \
+  --no-buffer
+```
+
+### Related
+
+- Issue: [#7280](https://github.com/pollinations/pollinations/issues/7280)


snz2:
Even setting stream true does not solve this problem. The result is the same. However, I saw that there was a difference in the quality of answers between using Azure Openai and writing directly on Openai. Azure has an excessive safe policy, so there are problems such as blocking certain keywords even for questions that do not have any particular problems. I think it handles wrapped prompts. Since Openai's policy is basically censorship-heavy anyway, I wonder if there is a need to apply a double policy. If pollinations.ai does not have a service contract with Azure, I wonder if there is any need to use Azure.

snz2:
When will this problem be resolved?

I am not urging you to do this, but I would appreciate it if you could let me know if it will take a long time.
Because when used as an agent, all operations result in error 524. It's difficult to use.

Apart from this, I would appreciate it if you could also look at the problem of not thinking in the Claude model I posted. The issue has been opened.

thank you

PR #7329, opened 2026-01-15
Preselect profile and balance permissions on login

voodoohop:
Adds default permissions to authorize URL.

**Changes:**
- Adds `permissions=profile,balance` query parameter to authorize URL
- Users see profile and balance permissions pre-selected when logging in
- Improves UX - no need to manually check these boxes

**Benefits:**
- Smoother login flow
- Balance and profile info loads immediately after auth
- Reduces friction for new users

---
*Description condensed by Claude for readability*


voodoohop:
polly can you review focus on negative if there is something

PR #7351, opened 2026-01-16
fix: simplify gen.pollinations.ai proxy

voodoohop:
- Add `/account/*` route for profile/balance/usage endpoints
- Remove `/openai` route (deprecated)
- Remove URL normalization logic (users should encode URLs properly)
- Remove `fixNestedImageUrls` function (~80 lines deleted)
- Fix type errors with `env.ENTER.fetch()`

**Result:** True thin proxy - just URL rewriting and forwarding.


voodoohop:

PR #7361, opened 2026-01-16
fix: CSV download for usage data

ElliotEtag:
- Fix download button to use fetch with credentials instead of anchor tag
- Correct API endpoint path to /api/account/usage  
- Add TINYBIRD_READ_TOKEN to dev secrets for Tinybird pipe access

Fixes CSV download button that was not working due to:
1. Missing session cookies (anchor tag doesn't send credentials)
2. Missing TINYBIRD_READ_TOKEN environment variable


voodoohop:
polly cna you review? focus on the negative

PR #7372, opened 2026-01-16
fix(image): propagate Azure API error status codes correctly

voodoohop:
- Fix Azure 400 errors being returned as 500
- Pass Azure error response directly (light proxy)

**Before:** `throw new Error("Azure GPT Image API error: 400 - ...")`  
**After:** `throw new HttpError(errorText, response.status)`

~570 errors/day now correctly return 400 with Azure's rejection reason.


voodoohop:
polly please review focusing on the negatve

PR #7386, opened 2026-01-17
[PATCH]: Removal of redundant workflow

Circuit-Overtime:
## Problem Faced:-  

The [profile update](https://github.com/pollinations/pollinations/actions/runs/21089708383) workflow was constantly failing everyday. I checked that the [Org Repo](https://github.com/pollinations/.github/blob/main/profile/README.md) uses the shields.io badges which are dynamic and they don't need any other workflow to work, so these two files are no longer needed to be run as a cron by the CI, but if we ever wish to use numbers than the shields badges then the best policy is to run this workflow in the pollinations/.github itself as we were doing here pollinations/pollinations

## Changes Made:- 

- Removed `.github/scripts/profile-update-stats.js`
- Removed `.github/workflows/profile-update-stats.yml` 

@ElliotEtag Elliot please take a look at it and merge accordingly


ElliotEtag:
polly , please tell me what these workflows were doing.

ElliotEtag:
can merge @Circuit-Overtime 
some biome errors though to fix prior to that ;))

Circuit-Overtime:
@ElliotEtag The biome error was due to an intentional code design which I figured out now --  It was a CI failure where **Biome exited with an error when only `.github/**` files were changed**.

**Root cause**
- Biome excludes `.github/**` from its eligible file set by default.
- The workflow passed changed files explicitly to `biome check`.
- When all changed files were under `.github/`, Biome received only ignored paths and exited with a non-zero error.

**Fix**
- Updated the workflow/config to avoid passing `.github/**` files to Biome (or explicitly include them, depending on the change).
- Prevents Biome from running with an empty post-filter file set.

> No functional code changes this only stabilizes CI behavior.

PR #7462, opened 2026-01-19
fix: disable crypto payments temporarily

ElliotEtag:
- Remove Fiat/Crypto payment toggle from UI
- Disable nowpayments invoice endpoint (returns 503)
- Keep webhook handler for in-flight payments
- Preserve cryptoBalance for existing users


ElliotEtag:
polly please review

PR #7537, opened 2026-01-21
feat(enter): add manage api key dialog with enable toggle and budget

fisventurous:
the following changes adds a "manage" dialog for api keys, users can enable/disable keys, edit budget, and modify allowed models without deleting.

## changes
- add EditApiKeyDialog component with enable toggle, budget input, model selection
- add gear button next to each api key in the list
- update api-keys.ts backend to support enabled field
- disabled keys show with reduced opacity and "(disabled)" label

## because
- users currently can't disable a key without deleting it (losing history)
- no way to fix a key with wrong budget after creation
- this is a simple additive, existing delete mechanism unchanged

## files
- src/client/components/edit-api-key-dialog.tsx (new)
- src/client/components/api-key.tsx
- src/client/routes/index.tsx
- src/routes/api-keys.ts

### DIFF
--- a/enter.pollinations.ai/src/client/components/api-key.tsx
+++ b/enter.pollinations.ai/src/client/components/api-key.tsx
@@ -355,7 +355,9 @@ export const ApiKeyList: FC<ApiKeyManagerProps> = ({
                                         {/* Row 2: Stats and Permissions */}
                                         <div className="flex flex-wrap items-center gap-4 text-xs">
                                             <span title="Created">
-                                                <span className="text-gray-400">Created: </span>
+                                                <span className="text-gray-400">
+                                                    Created:{" "}
+                                                </span>
                                                 <span className="text-gray-500">
                                                     {formatDistanceToNowStrict(
                                                         apiKey.createdAt,
@@ -367,7 +369,9 @@ export const ApiKeyList: FC<ApiKeyManagerProps> = ({
                                                 </span>
                                             </span>
                                             <span title="Last used">
-                                                <span className="text-gray-400">Used: </span>
+                                                <span className="text-gray-400">
+                                                    Used:{" "}
+                                                </span>
                                                 <span className="text-gray-500">
                                                     {apiKey.lastRequest
                                                         ? formatDistanceToNowStrict(
@@ -389,7 +393,9 @@ export const ApiKeyList: FC<ApiKeyManagerProps> = ({
                                                 }
                                             />
                                             <span className="flex items-center gap-1">
-                                                <span className="text-gray-400">Models:</span>
+                                                <span className="text-gray-400">
+                                                    Models:
+                                                </span>
                                                 <ModelsBadge
                                                     permissions={
                                                         apiKey.permissions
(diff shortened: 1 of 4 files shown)


voodoohop:
Fixed the critical issues found during review:

## Bugs Fixed

### 1. **Authentication Middleware** (`src/middleware/auth.ts`)
- Added check for `enabled` field when verifying API keys
- Disabled keys (`enabled: false`) now properly reject authentication
- Previously disabled keys could still make API calls

### 2. **Permissions Parsing** (`src/routes/api-keys.ts`)
- Fixed JSON parsing of `existingKey.permissions` 
- Was treating JSON string as object, causing "expected array, received string" errors
- Now properly parses permissions before passing to better-auth

### 3. **Error Display** (`src/client/components/edit-api-key-dialog.tsx`)
- Added error state and display in the UI
- Users now see error messages when updates fail
- Previously errors were only logged to console

The update functionality should now work correctly. Test by:
1. Opening the manage dialog for an API key
2. Toggle enabled/disabled state
3. Modify budget or model permissions
4. Click Save

Disabled keys will no longer authenticate for API requests.

ElliotEtag:
## ⚠️ Known Issue: better-auth verifyApiKey Race Condition

During investigation of API key permission updates being overwritten, we found a race condition in better-auth's `verifyApiKey` function.

### Root Cause

Looking at `better-auth/dist/plugins/api-key/routes/verify-api-key.mjs:114-132`:

```javascript
const updated = {
    ...apiKey,        // ⚠️ Spreads ALL fields including permissions
    ...update,        // rate limit fields
    remaining,
    lastRefillAt,
    updatedAt: new Date()
};
// Then writes ENTIRE object back to DB
```

### The Race Condition

1. Your update writes new permissions at time T1
2. Another request's `verifyApiKey` reads at T0 (before your update)
3. That verify writes back at T2, overwriting your T1 changes

### The Real Bug

better-auth should use **partial updates** (only `remaining`, `lastRefillAt`, `requestCount`, `updatedAt`) instead of full object replacement.

### Short-term Workaround

Wait ~1 minute after updating permissions before using the key.

### Long-term Fix Options

1. **File an issue/PR with better-auth** to use partial updates
2. **Override the adapter's update method** to only update rate-limit fields during verify
3. **Use a custom verifyApiKey** that doesn't write back permissions

This doesn't block the current PR but is worth noting for future work on API key management.

ElliotEtag:
## ✅ Race Condition Fixed

Added a patch to `better-auth` that fixes the `verifyApiKey` race condition.

### The Fix
`verifyApiKey` now only updates rate-limit fields (`remaining`, `lastRefillAt`, `requestCount`, `updatedAt`) instead of the entire API key object. Permissions are never touched during verify operations.

### Files Added
- `patches/better-auth+1.4.17.patch` — the fix
- `package.json` — added `postinstall: patch-package` for automatic application

### Result
User updates permissions → concurrent API requests (100+/sec) → permissions persist ✅

The patch auto-applies on `npm install`. Can be removed once better-auth releases an upstream fix.

ElliotEtag:
polly can you review systematically and deeply?

ElliotEtag:
polly, what do you think now?

PR #6320, opened 2025-12-22
refactor: apps/APPS.md as single source of truth

ElliotEtag:
## Summary

Consolidates all app listings into a single `apps/APPS.md` markdown file as the source of truth.

### Source of Truth
- **`apps/APPS.md`** - Single markdown table with 318 apps
- Columns: Emoji | Name | Description | Language | Category | GitHub | Repo | Stars | Discord | Other | Submitted
- Sorted by date (newest first)

### Automated Workflow
- `tier-app-submission.yml` parses issues with AI
- Extracts all fields from issue template
- Fetches GitHub stars automatically
- AI generates emoji and reformulates description (~80 chars)
- Prepends new row to `apps/APPS.md`
- Updates `README.md` with last 10 apps

### hello.pollinations.ai
- New `src/lib/parseApps.ts` parses `apps/APPS.md` at build time
- `predev`/`prebuild` scripts copy file from `apps/` directory
- Copied file is gitignored (source of truth remains `apps/APPS.md`)

### Maintenance Script
- New `tier-apps-check-links.js` - Checks all app URLs for broken links
- Generates report with `--report` flag
- Can filter by category with `--category=<name>`

### Scripts
- `pollinator-agent/project-list-scripts/*` (9 files)
- `.github/scripts/app_list_update_stars.js`

### Workflows
- `app-list-update-entries.yml`
- `app-list-update-stars.yml`

### Documentation
- `.claude/skills/app-submissions/SKILL.md`

### Issue Template
- Removed "Additional Information" field
- Added "Other Contact" field
- Fields: Name, Description, URL, GitHub Repo, Discord, Category, Language, Other Contact

### Column Cleanup (APPS.md)
- Renamed: Author → GitHub, Date → Submitted
- Added: Other (for non-GitHub contacts)
- Reordered: Language moved after Description
- Cleaned emojis (moved from names to Emoji column)
- Converted country flags to ISO language codes
- Added 128 missing repo URLs from old JS files

### Documentation
- `AGENTS.md` - Updated with new system instructions
- `README.md` - Shows last 10 apps only
- `.github/workflows/README.md` - Updated flow diagrams

### PR Labeling Workflow
- `tier-label-external-pr.yml` - Improved logic:
  - `pr:external` → Only for non-org members
  - `tier:review` → Only for users with no merged PRs

---

⚠️ **Note:** `pollinations.ai/` is untouched - website being replaced soon.

PR #7533, opened 2026-01-21
feat: Add issue tracking columns to APPS.md

ElliotEtag:
- Adds `Submitted_Date`, `Issue_URL`, `Approved_Date` columns to APPS.md
- Backfills 472 apps (426 matched with original GitHub issues, 90% match rate)
- Updates all parsing scripts for 15-column format
- Updates SKILL.md documentation

**New column structure:**
```
| ... | Submitted_Date | Issue_URL | Approved_Date |
```

- **Submitted_Date**: Issue creation date (when user submitted)
- **Issue_URL**: Link to original GitHub issue  
- **Approved_Date**: PR merge date (renamed from old Submitted column)

PR #7539, opened 2026-01-21
feat: add weekly star update workflow for apps

fisventurous:
## summary
adds a scheduled workflow to auto-update github stars for apps in `APPS.md`. also flags deleted repos.

## changes
- add `app-update-stars.js` script that fetches star counts from github api
- add `app-update-stars.yml` workflow running weekly (sunday 4am utc)
- supports `--dry-run` and `--verbose` flags for testing

noticed the old `app-list-update-stars.yml` was removed in #6320 during the APPS.md refactor. figured it might have been accidental or just lost in the shuffle, so this brings it back in a cleaner form that works with the new system (that is if you still want it!).

### DIFF
--- a/.github/scripts/app-update-stars.js
+++ b/.github/scripts/app-update-stars.js
@@ -251,7 +251,7 @@ async function main() {
         }
     }
 
-    return stats.deleted > 0 ? 1 : 0;
+    return 0;
 }
 
 main()


fisventurous:
noticed PR #7533 just landed with the new APPS.md structure. just updated the script in [ae2fec8](https://github.com/pollinations/pollinations/pull/7539/commits/ae2fec8a000dc5418255fc0f57e46cda0307d6f5) to be column name aware so it handles both the old (13-col) and new (15-col) formats dynamically. shouldn't have any conflicts hopefully

voodoohop:
@fisventurous Nice PR! One issue:

**Line 254**: `return stats.deleted > 0 ? 1 : 0;`

This exits with code 1 if any repo is deleted, which will **fail the workflow and skip the commit step** - so star updates won't be saved.

**Fix**: Change to `return 0;` - deleted repos are already logged as warnings, no need to fail the workflow.

PR #7578, opened 2026-01-22
feat: add time of pollen refill to /balance api endpoint

XHyperDEVX:
This PR adds that the /account/balance endpoint not only shows the balance but also the time in seconds until the next pollen refill (every 24 hours)


XHyperDEVX:
ready to merge

XHyperDEVX:
Obsolete, as it is now always refilled at 00:00 UTC.

PR #7606, opened 2026-01-23
[PATCH]: Migrated the .ts to .js type files for workflow scripts 

Circuit-Overtime:
## Changes Made:- 

- rename .github/scripts/{app-check-duplicate.ts => app-check-duplicate.js}
- rename .github/scripts/{app-validate-submission.ts => app-validate-submission.js}

## Reason:- 

- .js easier to compile on the git actions serverless 
- the scripts/ dir contained .ts and .js both, we are not aiming for type safety in this scenario (not needed for straight forward scripts)
- shift to one consistent .js file type for the scripts/ dir


Circuit-Overtime:
polly check if all the workflows are not correctly using the .js file type with node as run time v8 engine

Circuit-Overtime:
polly check if this is good to merge

PR #7610, opened 2026-01-23
fix(text): add default max_tokens for Claude fallback configs

voodoohop:
## Summary
- Adds `defaultOptions: { max_tokens: 200000 }` to Claude fallback configs

## Problem
Vertex AI Claude models require `max_tokens` parameter. Without it, requests fail when falling back from AWS Bedrock to Vertex AI.

## Solution
- Added default to `claude-sonnet-4-5-fallback` and `claude-opus-4-5-fallback`
- User-provided values override the default via existing `modelResolver.js` logic

## Testing
- Verified config loads correctly with `npx tsx`
- Confirmed user override works (user value takes precedence)


voodoohop:
polly can you review? focus on the negative

PR #7617, opened 2026-01-24
fix(enter): improve auth error messages to differentiate invalid key vs missing auth

voodoohop:
## Summary

Improves authentication error messages to clearly differentiate between two distinct failure cases:

1. **Invalid API key provided** → `"Invalid API key. Please check your key and try again."`
2. **No authentication provided** → `"Authentication required. Please provide an API key via Authorization header (Bearer token) or ?key= query parameter."`

## Problem

Previously, both cases returned the same generic message:
> "You need to authenticate by providing a session cookie or Authorization header (Bearer token)."

This was confusing because:
- Users with invalid keys thought they hadn't provided auth at all
- The "session cookie" mention is an internal-only feature, irrelevant to API users

### `src/middleware/auth.ts`
- When API key verification fails (`!keyResult.valid`), now throws a specific 401 error instead of returning `null`

### `src/error.ts`  
- Updated default 401 message to be clearer and remove internal implementation details

## Testing

```bash
# Invalid API key - now shows specific error
curl "https://enter.pollinations.ai/api/generate/image/test?model=flux" \
  -H "Authorization: Bearer invalid_key_here"
# → "Invalid API key. Please check your key and try again."

# No auth provided - shows generic auth required message  
curl "https://enter.pollinations.ai/api/generate/image/test?model=flux"
# → "Authentication required. Please provide an API key..."
```

### DIFF
--- a/enter.pollinations.ai/src/client/components/account-permissions-input.tsx
+++ b/enter.pollinations.ai/src/client/components/account-permissions-input.tsx
@@ -16,7 +16,7 @@ const ACCOUNT_PERMISSIONS = [
     {
         id: "balance",
         label: "Balance",
-        description: "Read pollen balance via API",
+        description: "Read key's budget (if set) or account balance",
     },
     {
         id: "usage",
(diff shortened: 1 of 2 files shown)


voodoohop:
polly please review

PR #7629, opened 2026-01-24
fix(enter): return 429 for rate limit errors instead of 401

voodoohop:
## Problem

The `better-auth` library has a bug where rate limiting returns HTTP 401 (Unauthorized) instead of HTTP 429 (Too Many Requests). This causes users to see "Invalid API key" errors when they're actually being rate limited.

## Root Cause

In `better-auth`'s `verify-api-key.mjs`:
```javascript
if (success === false) throw new APIError("UNAUTHORIZED", {
    message: message ?? void 0,
    code: "RATE_LIMITED",  // ← correct code but wrong HTTP status
    details: { tryAgainIn }
});
```

## Solution

Detect the `RATE_LIMITED` error code and return the correct 429 status:

```typescript
if (keyResult.error?.code === "RATE_LIMITED") {
    throw new HTTPException(429, {
        message: "Rate limit exceeded. Please try again later.",
    });
}
```

## Changes

- **1 file changed**: `enter.pollinations.ai/src/middleware/auth.ts`
- **6 lines added**: Simple check before the existing 401 error

This is a minimal workaround until better-auth fixes the upstream bug.

### DIFF
--- a/enter.pollinations.ai/drizzle/0016_migrate-temporary-keytype-to-secret.sql
+++ b/enter.pollinations.ai/drizzle/0016_migrate-temporary-keytype-to-secret.sql
@@ -1,0 +1,8 @@
+-- Custom SQL migration file, put your code below! --
+-- Migrate keyType: "temporary" to "secret" in metadata JSON
+-- The "temporary" type was redundant - we now only have "secret" and "publishable"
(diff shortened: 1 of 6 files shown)


voodoohop:
polly can you review

PR #7631, opened 2026-01-24
fix(text): handle image URL conversion for Claude fallback configs

### LINKED ISSUE #7516: regression: HTTP 500 on model='claude' with remote image URLs (Cloudinary)
**Reported by:** `mano.resells.` (Discord)

**Description:**
User reports that `model="claude"` and `claude-large` frequently return **HTTP 500** errors when the input message includes a remote image URL (specifically citing Cloudinary links). 

**Context & Potential Cause:**
This appears to be a regression or related to **Issue #5862** ("Fix: Claude/Bedrock models require base64 image conversion"), which was closed in Dec 2025. 
- That issue identified that Claude models (via Bedrock) failed with HTTP 500 on remote URLs because they required base64 conversion, which was missing.
- It's possible the fix was reverted, or the provider configuration changed (e.g., check `provider === "bedrock"` might not be matching if the underlying provider tag changed, or if we switched to Anthropic direct which also might need handling).

**Steps to Reproduce:**
1. Send a chat completion request to `gen.pollinations.ai`.
2. Use `model: "claude"` or `"claude-large"`.
3. Include an image content block with a remote URL (e.g., Cloudinary).
4. **Result:** HTTP 500 Internal Server Error.

**Expected Behavior:**
The backend should automatically fetch and convert the remote URL to the format required by the upstream provider (Base64 if needed), similar to how `gemini` handles it.

---
**Author:** `mano.resells.` (UID: `1147416764179816498`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/889573359111774329/1463298107247955979)

voodoohop:
## Summary
Fixes #7516 - Claude models returning HTTP 500 with remote image URLs (Cloudinary, etc.)

## Root Cause
The `imageUrlToBase64Transform` was checking `options.modelConfig.provider` to determine if base64 conversion was needed. However, the **fallback configs** for `claude` and `claude-large` models use a nested structure with a `targets` array instead of a top-level `provider` field:

```javascript
// claude-sonnet-4-5-fallback config structure:
{
    strategy: { mode: "fallback" },
    targets: [
        { provider: "bedrock", ... },   // ← provider is nested here
        { provider: "vertex-ai", ... }
    ]
}
// NO top-level "provider" field!
```

The transform was checking `config.provider` which is `undefined` for fallback configs, so the base64 conversion **never happened** for Claude models with image URLs.

## Changes
- Check for providers in `config.targets` array for fallback configs
- Apply base64 conversion when any target uses `vertex-ai` or `bedrock`
- Update log message to handle fallback configs gracefully

## Testing
- ✅ Existing test `POST /v1/chat/completions should accept image URL for Claude/Bedrock models (Issue #5862)` passes
- ✅ Local manual test with Claude + image URL succeeded:
  ```bash
  curl -s "http://localhost:16385/v1/chat/completions" \
    -H "x-enter-token: $TOKEN" \
    -d '{"model": "claude", "messages": [{"role": "user", "content": [{"type": "text", "text": "Describe this image"}, {"type": "image_url", "image_url": {"url": "https://picsum.photos/id/237/200/300"}}]}]}'
  ```
  Response: `{"model": "claude-sonnet-4-5-20250929", "provider": "vertex-ai", "choices": [{"message": {"content": "Adorable"}}]}`

### DIFF
--- a/text.pollinations.ai/transforms/imageUrlToBase64Transform.js
+++ b/text.pollinations.ai/transforms/imageUrlToBase64Transform.js
@@ -187,9 +187,10 @@ export function createImageUrlToBase64Transform() {
             return { messages, options };
         }
 
-        log(
-            `Processing messages for ${provider || "fallback"} image URL conversion`,
-        );
+        const providerInfo = provider
+            ? provider
+            : `fallback[${targets.map((t) => t.provider).join(", ")}]`;
+        log(`Processing messages for ${providerInfo} image URL conversion`);
 
         // Process all messages in parallel
         const processedMessages = await Promise.all(


voodoohop:
poly can you review

voodoohop:
polly can you review

PR #7637, opened 2026-01-24
fix: preserve Gemini thought fields and signature for tool calls

FabioArieiraBaia:
### Description
This PR fixes a critical issue where Gemini/Vertex AI tool calls would fail with a 400 error (Missing Thought Signature).

### Problem
When Gemini returns a response that includes reasoning (thinking) and a tool call, it provides a 	hought_signature. Vertex AI requires that this exact signature be returned in the messages history when the user provides the tool result. 
Pollinations was discarding this field during message normalization and object cleaning, causing the provider to reject the follow-up request.

### Solution
1. **New Transform**: Added passthroughToolSchemas.js to explicitly allow 	hought_ prefixed fields in the pipeline.
2. **Message Normalization**: Updated 	extGenerationUtils.js to preserve 	hought and 	hought_signature fields.
3. **Object Cleaning**: Updated objectCleaners.js to prevent the recursive cleaner from removing 	hought_ fields.
4. **Request Extraction**: Updated 
equestUtils.js to recognize 	hought_signature from incoming API calls.
5. **Model Integration**: Applied the passthrough transform to all Gemini models in vailableModels.ts.

This ensures compatibility with Gemini's cryptographic validation of reasoning-to-tool-result chains.


voodoohop:
## Analysis: thought_signature is already preserved

I've analyzed the code and **the current implementation already preserves `thought_signature` correctly**. Here's the evidence:

### Testing shows it works:

```javascript
// Test with actual Gemini message format
const geminiResponse = {
    role: "assistant",
    content: "",
    tool_calls: [{
        id: "call_abc123",
        type: "function",
        function: {
            name: "search_web",
            arguments: '{"query": "weather"}',
            thought_signature: "gemini_thought_sig_xyz789"
        }
    }]
};

// After validateAndNormalizeMessages() - PRESERVED ✓
// After cleanNullAndUndefined() - PRESERVED ✓
```

### Why it works:

1. `validateAndNormalizeMessages()` copies entire `tool_calls` array: `if (msg.tool_calls) normalizedMsg.tool_calls = msg.tool_calls;`
2. `cleanNullAndUndefined()` doesn't recursively clean inside arrays, only objects
3. Test snapshots confirm this - `text-3f46c644ae328a98395429f874891599.json` shows `thought_signature` preserved in requests

### The only gap:

Message-level `thought`/`thought_signature` fields would be stripped, but **Gemini doesn't use these** - it puts `thought_signature` inside `tool_calls[].function.thought_signature` per [Portkey gateway code](https://github.com/Portkey-AI/gateway/blob/main/src/providers/google-vertex-ai/chatComplete.ts#L273-L275).

### Suggested simplification:

This PR adds 200+ lines across 5 files, but the issue is already solved. If we want extra safety, we could add just this 3-line change to `objectCleaners.js:48`:

```javascript
// Protect thought_signature from cleaning (just in case)
if (key === "modalities" || key === "audio" || key === "thought_signature") {
    return;
}
```

But even this is likely unnecessary since arrays aren't recursively cleaned.

### Evidence from production:
- Multiple test snapshots show `thought_signature` preserved in tool calls
- The field location matches Google's docs: `tool_calls[].function.thought_signature`
- No cleaning happens inside arrays, so nested signatures are safe

Would you like me to submit a minimal PR showing this works, or shall we close this as already functioning?

FabioArieiraBaia:
While some implementations might nest the signature inside tool_calls, these newer reasoning models often deliver thought and thought_signature (or reasoning_content) as top-level fields within the assistant message.

The current validateAndNormalizeMessages implementation reconstructs messages from scratch using a strict whitelist, which silently strips these top-level fields. When the application sends back the tool result in the second request, Google/Vertex AI rejects it with a 400 Bad Request because the preceding assistant message in the history no longer contains the original 'proof of reasoning' (the thought and its signature).

The Object.keys(msg).forEach approach we implemented is necessary to ensure these critical reasoning fields are preserved during normalization, maintaining the integrity of the conversation history as required by Vertex AI for sequential tool execution.

voodoohop:
Superseded. The Vertex/Gemini-3 tool-call failure this targets was fixed by #8909 (sets `strict-openai-compliance: false`). On current main the signature lives nested in `tool_calls[].function.thought_signature`, which gen preserves (`validateAndNormalizeMessages` copies `tool_calls` wholesale; `cleanNullAndUndefined` never recurses into arrays). Also targets the deleted `text.pollinations.ai/` tree. Any residual top-level-field coverage would be a ~2-line change in gen, not this 200-line PR.

PR #7650, opened 2026-01-25
fix: add deferUpdates and backgroundTasks for D1 contention fix

voodoohop:
## Summary
Fixes 401 "Invalid API key" errors for high-volume users (Roblox games with 1.5k+ concurrent players) caused by D1 database contention.

## Problem
When many concurrent requests use the same API key, better-auth tries to update `lastRequest`, `requestCount`, and `remaining` fields synchronously. This causes D1 write conflicts under high load, resulting in:
```
ERROR [Better Auth]: Failed to validate API key:
```

### 1. Enable `deferUpdates: true` (auth.ts)
Moves non-critical DB updates to background tasks instead of blocking the request.

### 2. Add `backgroundTasks.handler` with `ctx.waitUntil()` (auth.ts)
Ensures deferred updates complete properly in Cloudflare Workers environment.

### 3. Pass `ExecutionContext` to `createAuth()` 
Updated all call sites:
- `src/index.ts`
- `src/middleware/auth.ts`
- `src/routes/stripe.ts`

### 4. Increase edge rate limiter (wrangler.toml)
- Production: 100/10s → 10000/10s (1000 req/s per IP)
- Staging: 100/10s → 10000/10s
- Test: 100/10s → 10000/10s

### 5. Disable better-auth rate limiting
Rate limiting is handled by the pollen system instead.

## Why This is Safe
- Deferred fields (`lastRequest`, `requestCount`) are non-critical analytics
- Rate limiting is disabled anyway (handled by pollen)
- Auth validation still happens synchronously
- `waitUntil()` guarantees background tasks complete

## Testing
Already deployed to production and verified working - no more 401 errors in logs.

### DIFF
--- a/enter.pollinations.ai/src/auth.ts
+++ b/enter.pollinations.ai/src/auth.ts
@@ -84,13 +84,17 @@ export function createAuth(env: Cloudflare.Env, ctx?: ExecutionContext) {
         advanced: {
             // Configure background tasks for Cloudflare Workers
             // Required for deferUpdates to work properly
-            backgroundTasks: ctx ? {
-                handler: (promise: Promise<any>) => {
-                    ctx.waitUntil(promise.catch((e) => {
-                        console.error("Background task failed:", e);
-                    }));
-                }
-            } : undefined,
+            backgroundTasks: ctx
+                ? {
+                      handler: (promise: Promise<any>) => {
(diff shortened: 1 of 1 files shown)


voodoohop:
polly can you review

PR #7676, opened 2026-01-26
fix: optimize spore→seed upgrade to avoid GitHub API rate limits

voodoohop:
## Problem
Script hitting GitHub API rate limits (1,000 points/hour) when validating 16k+ spore users.

## Solution (stateless - no checkpoint files or extra DB columns)

1. **Day-based slicing with LIMIT/OFFSET**
   - Split older users into 7 equal slices by position (oldest to newest)
   - Monday = slice 1/7, Tuesday = slice 2/7, etc.
   - ~2,300 users/day instead of 16k all at once

2. **New user priority**
   - Users created in last 24h always checked first

3. **Reduced API cost**
   - Changed `repositories(first: 100)` → `first: 5` (ordered by stars DESC)
   - ~20x reduction in GraphQL points

## Result
~150 API points/day, full coverage in 7 days, repeats weekly.


voodoohop:
polly can you review?

PR #7705, opened 2026-01-27
fix: use git fetch + reset instead of pull to handle divergent branches

voodoohop:
## Problem

The deploy-enter-services workflow failed because the server had divergent branches:

```
err: hint: You have divergent branches and need to specify how to reconcile them.
```

This happened because someone had checked out a different branch on the server (`fix/register-endpoint-auth-bypass`) and the local `production` branch was 20 commits behind the remote.

## Solution

Replace `git pull origin <branch>` with:
```bash
git fetch origin <branch>
git checkout <branch> 2>/dev/null || git checkout -b <branch> origin/<branch>
git reset --hard origin/<branch>
```

This ensures:
1. The server always matches the remote branch exactly
2. Works regardless of what branch was previously checked out
3. Handles divergent branches by force-resetting to remote state

## Changes

- Updated both staging and production deployment scripts in `deploy-enter-services.yml`

## Testing

Manually fixed the server by running the same commands via SSH - deployment now works correctly.


voodoohop:
polly can you review

PR #7711, opened 2026-01-27
fix(enter): reject API keys with no user to prevent infinite pollen bug

### LINKED ISSUE #7709: Bug: Polar sync overwrites local pollen balance causing "Infinite Pollen" reset
User reports an "Infinite Pollen Glitch" where their balance decreases after usage but instantly resets to the full Tier limit (10.0 for Flower).

**Analysis:**
The issue appears to be in `src/middleware/polar.ts`.
The sync logic fetches `polarTier` (from Polar API) and unconditionally overwrites the local `tierBalance` in D1:
```typescript
if (polarTier + polarPack > 0) {
    tierBalance = polarTier; // <--- Overwrites local usage!
    await db.update(userTable).set({ tierBalance, packBalance })
```
If Polar is not updated with real-time usage (which it likely isn't for performance), it returns the full grant amount. This overwrites any local decrements since the last sync.

**Impact:**
- Users effectively have infinite daily pollen (up to the sync interval).
- Usage tracking is accurate in logs but not enforced in balance.

**Suggested Fix:**
- Only sync from Polar if the Polar balance *changed* (e.g. new grant ID or increased amount).
- Or, take `min(localBalance, polarBalance)` (though this risks losing grants).
- Better: Track "synced_grant_id" and only apply new grants, treating local balance as truth for decrement.

**Reported by:** `monkedevhacc` (Discord)

---
**Author:** `monkedevhacc` (UID: `1144599607389728908`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1465758281887645808/1465763123238666333)

ElliotEtag:
## Summary
Fixes #7709

## Problem
Users reported an "Infinite Pollen Glitch" where balance decreases after usage but instantly resets to full tier limit.

**Root cause:** When an API key has a broken/missing user association in D1, the balance deduction in `track-helpers.ts` was silently skipped because `userId` was undefined.

## Solution
1. **Reject requests** with HTTP 402 when API key exists but has no associated user
2. **Add backup logging** in track-helpers.ts if billing is somehow still skipped

## Changes
- `enter.pollinations.ai/src/middleware/auth.ts`: Add check for missing user, throw 402
- `enter.pollinations.ai/src/utils/track-helpers.ts`: Add error logging as safety net

## Testing
- API keys with valid user: ✅ Works normally
- API keys with missing user: ✅ Returns 402 with helpful message

### DIFF
--- a/enter.pollinations.ai/src/middleware/auth.ts
+++ b/enter.pollinations.ai/src/middleware/auth.ts
@@ -129,8 +129,22 @@ export const auth = (options: AuthOptions) =>
                 userId: fullApiKey?.user?.id,
             });
 
+            // CRITICAL: Reject if API key exists but has no associated user
+            // This would cause billing to silently skip balance deduction (infinite pollen bug)
+            // Use 402 because this is a billing issue - key is valid but we can't bill
+            if (!fullApiKey?.user) {
+                log.error(
+                    "AUTH_NO_USER: API key {keyId} has no associated user - rejecting request",
+                    { keyId: keyResult.key.id },
+                );
+                throw new HTTPException(402, {
+                    message:
+                        "API key is not associated with a valid user. Please regenerate your API key.",
+                });
+            }
+
             return {
-                user: fullApiKey?.user as User,
+                user: fullApiKey.user as User,
                 apiKey: {
                     id: keyResult.key.id,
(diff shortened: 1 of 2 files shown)


ElliotEtag:
polly please review with care

PR #7719, opened 2026-01-28
fix(enter): pass ExecutionContext to better-auth hooks for Tinybird logging

ElliotEtag:
The ExecutionContext was not being passed through the better-auth plugin chain,
causing waitUntil() calls to silently skip when ctx.context.executionCtx was undefined.

This fix threads ExecutionContext from createAuth -> tierPlugin -> onAfterUserCreate
so tier events are properly sent to Tinybird on user registration.

This fix threads ExecutionContext from createAuth -> tierPlugin -> onAfterUserCreate so tier events are properly sent to Tinybird on user registration.

Also removes redundant timestamp field (added internally by sendTierEventToTinybird).


ElliotEtag:
Hey polly, I'm really not sure this is addressing the bug. The bug is that we cannot see anything entering the Tinybird tier datasource table when we are in production, staging works (note that in Tinybird, staging and production share the same table; it's only that we have a column for the environment.) 

Can you just review this and give me your opinion on that?

PR #7767, opened 2026-01-29
fix(mcp): invalidate model cache when API key is set or cleared

Tatine13:
### Description
   This PR fixes a bug in the MCP server where the list of available models was not updated after setting or clearing an API key.

   ### Problem
   The `modelCache.js` uses a global in-memory cache with a 5-minute TTL. If an MCP client (observed using `gemini-cli` via `npx`) calls `listTextModels` before setting
   an API key, it caches the anonymous model list. Subsequent calls to the `setApiKey` tool update the credentials but do not clear this cache, leaving the user stuck
   with a limited model list until the TTL expires or the server restarts.

   ### Fix
   - Import and call `clearModelCache()` inside `setApiKey` and `clearApiKey` handlers in `authService.js`.
   - This ensures that the next call to `listTextModels` fetches a fresh list corresponding to the new authentication tier (Seed/Flower/Spore).

   ### Testing
   Verified in a local environment. Before this fix, `listTextModels` required a server restart to show authenticated models after a key was set. With this fix, the mod
   list refreshes immediately after the tool call.

### DIFF
--- a/packages/mcp/src/services/authService.js
+++ b/packages/mcp/src/services/authService.js
@@ -1,4 +1,4 @@
-import { createMCPResponse, createTextContent } from "../utils/coreUtils.js";
+import { createMCPResponse, createTextContent } from \"../utils/coreUtils.js\";
 import {
     setApiKey as storeApiKey,
     getApiKey,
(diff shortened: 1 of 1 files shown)


voodoohop:
Thanks @Tatine13! Simplified the fix by removing caching entirely:

- Removed cache logic from `modelCache.js` → renamed to `models.js`
- Removed `clearModelCache` function and calls
- Removed `refresh` parameter from list tools
- Fixed escaped quotes issue

~150ms API call is fine for occasional model list requests. No more staleness bugs.

PR #7875, opened 2026-02-03
fix(enter): fix tier refill idempotency and add microbe tier

ElliotEtag:
## Problem

The daily tier refill was **skipping most users**. Only ~607 out of ~23,000 users were getting refilled.

**Root cause:** The idempotency check used `MAX(last_tier_grant)` across all users to determine if the bulk refill had already run today. However, individual users also get their `lastTierGrant` updated on login (in `auth.ts`). This caused the bulk refill to skip whenever any single user had logged in that day.

**Timeline of the bug:**
1. Feb 1st 23:45 UTC - Last successful bulk refill (22,998 users)
2. Feb 2nd ~00:00 UTC - Some user logged in, got individual refill → MAX(last_tier_grant) became "today"
3. Feb 2nd 00:00 UTC - Cron ran → **skipped** because it thought refill already happened
4. Same pattern repeated on Feb 3rd

## Solution

1. **Use KV store for bulk refill tracking** - Store bulk refill timestamp in KV (`tier:bulk_refill:last_timestamp`) separately from individual user data
2. **Add microbe tier to bulk refill SQL** - The new microbe tier (0.1 pollen/day) was missing from the CASE statement

## Changes

- `enter.pollinations.ai/src/routes/admin.ts`:
  - New KV-based idempotency check (`getLastBulkRefillTime` / `setLastBulkRefillTime`)
  - Added `microbe` tier to bulk refill SQL CASE statement
  - Removed unused `D1Database` type import

## Testing

- Existing tests pass
- After merge and deploy, manually trigger the refill workflow to refill all users

### DIFF
--- a/enter.pollinations.ai/src/routes/admin.ts
+++ b/enter.pollinations.ai/src/routes/admin.ts
@@ -1,7 +1,6 @@
 import type { Logger } from "@logtape/logtape";
 import { getLogger } from "@logtape/logtape";
 import { eq, sql } from "drizzle-orm";
-import type { D1Database } from "drizzle-orm/d1";
 import { drizzle } from "drizzle-orm/d1";
 import { Hono } from "hono";
 import { HTTPException } from "hono/http-exception";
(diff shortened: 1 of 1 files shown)


ElliotEtag:
polly please review this, it's very important

PR #7915, opened 2026-02-04
feat(enter): settle negative pack balance from tier refill

ElliotEtag:
## Summary

When users overspend their pack balance (going negative), the debt is now automatically settled from their next daily tier refill.

### 1. Debt Settlement Logic (`admin.ts`)
- Modified `/trigger-refill` endpoint to deduct negative pack balance from tier refill
- If pack_balance is -$2, and tier refill is $10, user gets $8 tier balance and pack resets to $0

### 2. Balance Gauge (`pollen-balance.tsx`)
- Always shows 💎 segment (even when 0 or negative)
- Red color for negative balance
- Tooltip explains debt will be settled from next refill

### 3. Explanation Text (`Pricing.tsx`)
- Updated 'How Pollen Works' section to explain:
  - Spending order (daily → purchased)
  - Paid Only models behavior
  - Negative balance auto-settlement

### 4. Test Added
- `should settle negative pack balance from tier refill` test verifies the logic

## How It Works

| Day | Pack Before | Tier Refill | Pack After | Tier After |
|-----|-------------|-------------|------------|------------|
| 1   | -$0.20     | $10.00     | $0.00     | $9.80     |
| 2   | $0.00      | $10.00     | $0.00     | $10.00    |

The debt is paid from the tier refill, not forgiven for free.

### DIFF
--- a/enter.pollinations.ai/src/routes/admin.ts
+++ b/enter.pollinations.ai/src/routes/admin.ts
@@ -254,8 +254,11 @@ export const adminRoutes = new Hono<Env>()
             .where(sql`tier IS NOT NULL`);
 
         // Bulk update all tier balances
-        // If pack_balance is negative, use part of the tier refill to top it up to 0
-        // This settles any "debt" from overspending the previous day
+        // Debt settlement logic:
+        // 1. If pack_balance is negative, that debt is moved to tier_balance (pack reset to 0)
+        // 2. If tier_balance is already negative (from previous debt), carry it forward
+        // 3. Daily tier refill chips away at the total debt until fully recovered
+        // Example: microbe user with -$0.3 debt takes 3 days to recover (0.1/day)
         const result = await db.run(sql`
             UPDATE user
             SET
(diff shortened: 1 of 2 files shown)


ElliotEtag:
polly please review

fisventurous:
+1 to polly's edge case concern, also wondering about the abuse scenario where a paid user intentionally overspends into deep debt then cancels subscription before it can be recouped. might be worth considering:
hard cap on how negative pack balance can go, maybe -$20? or blocking "paid only" models when balance is negative.

ElliotEtag:
@fisventurous pack balance cannot go very deep fortunatelly

ElliotEtag:
image

ElliotEtag:
polly, what about tnow?

voodoohop:
Closing in favor of #8056 which takes a simpler approach: cap `pack_balance` at 0 instead of tracking and settling debt.

Thanks for the thorough implementation @ElliotEtag - we decided the simpler fix (preventing negative balances entirely) is preferable to debt tracking.

PR #8040, opened 2026-02-07
fix(ui): add cursor-pointer to select all buttons

fisventurous:
adds the `cursor-pointer` class to the 'Select all' and `Deselect all` buttons and to the text, image, and video category toggle buttons in model permissions


ElliotEtag:
polly, seems very simple but please double check

PR #8090, opened 2026-02-08
fix: add audio models to API key permissions UI

ElliotEtag:
## Summary
- Add `AUDIO_SERVICES` import to model permissions selector and account permissions input
- Audio models (elevenlabs, elevenmusic, whisper) now appear as an "Audio" category alongside Text, Image, and Video
- Update `getModelDisplayName` to resolve audio model display names
- Fix `totalModels` count to include audio models

**Root cause:** The model permissions UI only imported `IMAGE_SERVICES` and `TEXT_SERVICES` from the shared registry — `AUDIO_SERVICES` was never included, so audio models could not be selected when restricting API keys to specific models.

## Test plan
- [ ] Open Enter dashboard, create a new API key
- [ ] Toggle "Limited to selected models" — verify Audio category appears with elevenlabs, elevenmusic, whisper
- [ ] Select audio models and create key — verify they persist in key permissions
- [ ] Verify "Select all" / "Deselect all" works for Audio category


ElliotEtag:
polly review please

PR #8104, opened 2026-02-08
fix: preserve aspect ratio in klein-9b dimension clamping

voodoohop:
Fixes klein-9b only generating 1:1 images.

**Problem:** The dimension clamping was limiting each dimension individually to 1024, forcing non-square images to become square.

**Fix:** Now only scales based on total pixel count (1 megapixel max), preserving aspect ratio.

Example: 1920x1080 → 1360x768 (maintains 16:9 ratio)

---
*Description condensed by Claude for readability*


voodoohop:
polly can you review critically

voodoohop:
Polly can you review critically

PR #8111, opened 2026-02-09
[FEAT] : Polly now uses openai embeddings

Circuit-Overtime:
## Changes Made:- 

- Added `text-embedding-small` embedding model for polly 
- Initialized to 1536 dim 
- Polly saves the indexing amount in % for quick reference 
- uses hash functions to create a comparison diff to reduce compute load

## Check List:- 
- [x] New embedding model infographs 
- [x] Token count and dimension analysis


Itachi-1824:

PR #8117, opened 2026-02-09
fix(enter): single-bucket balance deduction — no splitting, allow negative

voodoohop:
## Summary

- Replace cascading split deduction (tier→crypto→pack with MAX/MIN) with single-bucket model
- Pick the first bucket with balance > 0, deduct full amount from it — can go negative
- Tier negatives auto-reset on daily refill, pack/crypto negatives absorb on next purchase
- Balance checks use any-bucket-positive (`||`) instead of summing
- Clamp negative values in API response and frontend gauge display

## Why this approach?

Simpler than #8056 (cap at zero) — no silent cost absorption, perfect accounting, and tier-only users (majority) never see negative pack balances. One request = one bucket = one line in the logs.

## Changes

- `balance-deduction.ts` — CASE-based single-bucket SQL for both `atomicDeductUserBalance` and `atomicDeductPaidBalance`
- `balance.ts` — `requirePositiveBalance`/`requirePaidBalance` check `tier > 0 || crypto > 0 || pack > 0` instead of `tier + crypto + pack > 0`
- `account.ts` — clamp each bucket at 0 before summing in `/balance` API
- `pollen-balance.tsx` — clamp display values so gauge never shows negative
- `tier-balance.test.ts` — updated all deduction tests + 6 new negative balance tests

## Supersedes

#8056

## Test plan

- [x] 26 tests pass (3 pre-existing failures unrelated to balance logic)
- [ ] Verify frontend gauge renders correctly with negative tier balance
- [ ] Verify `/balance` API returns correct clamped total

### DIFF
--- a/enter.pollinations.ai/test/fixtures.ts
+++ b/enter.pollinations.ai/test/fixtures.ts
@@ -1,9 +1,11 @@
-import { SELF } from "cloudflare:test";
+import { env, SELF } from "cloudflare:test";
 import type { Logger } from "@logtape/logtape";
 import { getLogger } from "@logtape/logtape";
 import { createAuthClient } from "better-auth/client";
 import { adminClient, apiKeyClient } from "better-auth/client/plugins";
+import { drizzle } from "drizzle-orm/d1";
 import { test as base, expect } from "vitest";
+import { user as userTable } from "@/db/schema/better-auth.ts";
 import { ensureConfigured } from "@/logger.ts";
 import { createFetchMock, teardownFetchMock } from "./mocks/fetch.ts";
 import { createMockGithub } from "./mocks/github.ts";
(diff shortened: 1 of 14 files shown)


voodoohop:
polly can you review critically

voodoohop:
## Addressed review feedback (295eef2c)

- **NULL safety fix:** COALESCE now wraps subtraction operands too, not just CASE conditions — prevents silent deduction loss when balance columns are NULL
- **Removed unused `packBalance` param** from `calculateDeductionSplit` and updated caller in `track-helpers.ts`
- **Concurrent test:** added comment explaining why only the total sum is asserted (D1 write order is non-deterministic)
- **Check/deduction race:** added comment documenting the pre-existing meter source mismatch (cosmetic, deduction is always correct)
- **Gauge 50/50 nit:** skipped — minor edge case when all balances are zero/negative

voodoohop:
polly can you review again critically taking into account the changes and the new comment

voodoohop:
polly can you re-review critically?

PR #8430, opened 2026-02-20
fix(docs): improve URL contrast in Scalar API reference

### LINKED ISSUE #7001: fix: URL highlight fix on the docs
Bug
I am Kushal and I was exploring around the page, so as to explore pollinations and then I found out that in the docs page, there is a box in which url is displayed which can be copied. The issue is the text in URL is highlighted in gray, which makes it difficult to be read.

Look at the screenshots below, for the same.

Fix
Change it to some other color (like darker gray or something else) so that it is properly visible to the user.

I have good knowledge around development and I also have knowledge on open source contributions, so would like to work on this issue.

Image

Image

mohityadav8:
Fixes #7001

Improves visibility of URL/code text in the API documentation by adding custom CSS to override default styling in Scalar.


ElliotEtag:
polly check this please, just wondering if that doc is autogenerated or if it make sense to edit it

mohityadav8:
The OpenAPI schema itself is auto-generated, but the Scalar UI configuration (including customCss) is defined manually in createDocsRoutes. This change only affects the rendering layer, not the generated schema.

Circuit-Overtime:
polly can you check the implementation and let me know? 
@mohityadav8 I am outside I will checkout to this branch soon and take a look too

voodoohop:
Thanks for this. The `// @ts-expect-error` removal is already done on main, so you'll need to rebase to clear that conflict.

The CSS fix itself is reasonable — `saturn` theme is dark so `#ffffff` works. One suggestion: scope it tighter to avoid affecting code blocks where Scalar already handles syntax highlighting:

```css
.references-classic code:not(pre code) {
    color: #ffffff;
}
```

This targets inline `code` in the docs without overriding code inside `<pre>` blocks. Also avoids `!important`.

Please rebase on main and update the CSS, then this is good to merge.

voodoohop:
Closing as obsolete. `enter…/docs.ts` no longer renders Scalar (it 301-redirects to gen/docs), and the Scalar dependency was removed entirely by the docs-generator rewrite #11302 / #11454. Link contrast is already handled in `gen…/routes/docs-styles.ts`.

voodoohop:
Closing as superseded. The docs surface moved from `enter` to `gen` and now uses newer Scalar CSS. If inline-code contrast is still an issue, it should be fixed narrowly in the gen docs styles.

PR #8446, opened 2026-02-20
fix: simplify polly router config + debug output

Itachi-1824:
## Summary
- Both polly workflows broken since ~Feb 19 — router crashes on new default model name (`claude-sonnet-4-6` replaced `claude-sonnet-4-5-20250929`)
- Simplified router config: only `default` → `claude-large` and `webSearch` → `perplexity-reasoning`, removed all unused routes (background, think, longContext)
- Enabled `show_full_output` temporarily so we can see actual error details if it still fails
- Aligned Polly persona system prompt across both workflows

## Test plan
- [ ] Trigger auto-polly by adding `polly` label to an issue
- [ ] Trigger main polly by commenting "polly" on an issue/PR
- [ ] Check Actions logs for successful run or detailed error output
- [ ] Remove `show_full_output: true` once confirmed working


Itachi-1824:
polly review please

PR #8453, opened 2026-02-20
fix: strip thinking params for gemini-2.5-flash-lite

voodoohop:
## Summary
- `gemini-2.5-flash-lite` doesn't support thinking config
- When users pass `thinking_budget: 0`, Portkey translates it to `thinking_config.include_thoughts` which Vertex AI rejects with 400: "include_thoughts is only enabled when thinking is enabled"
- Adds `"none"` model type to `createGeminiThinkingTransform` that strips thinking params without forwarding anything to the API
- Updates `gemini-fast` and `gemini-search` (both flash-lite) to use `"none"`

## Production evidence
```
vertex-ai error: Unable to submit request because Thinking_config.include_thoughts 
is only enabled when thinking is enabled.
```

## Test plan
- [ ] Deploy to text service
- [ ] Send request with `thinking_budget: 0` to `gemini-fast` — should succeed instead of 400

### DIFF
--- a/text.pollinations.ai/availableModels.ts
+++ b/text.pollinations.ai/availableModels.ts
@@ -104,7 +104,7 @@ const models: ModelDefinition[] = [
         transform: pipe(
             createSystemPromptTransform(BASE_PROMPTS.conversational),
             sanitizeToolSchemas(),
-            createGeminiThinkingTransform("v2.5"),
+            createGeminiThinkingTransform("none"),
         ),
     },
     {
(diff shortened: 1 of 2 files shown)


voodoohop:
Closing — targets the now-dead `text.pollinations.ai/` directory (removed by #10555); the `"none"` thinking type was never ported to gen's `createGeminiThinkingTransform`. gen's transform is opt-in (only emits `thinking.budget_tokens` when the caller sets it), so it doesn't inject unsupported params by default, and flash-lite routing has since moved to Vertex (#10118). If the Vertex 400 still reproduces on gemini-fast/gemini-search, please file a fresh targeted config fix against gen rather than reviving this stale diff.

PR #8579, opened 2026-02-25
fix: make polly OpenAI-compatible with full param passthrough

Itachi-1824:
- Polly's `/v1/chat/completions` now returns proper OpenAI format (`id`, `object`, `choices`, `usage`) — fixes the Zod validation crash when calling polly through gen.pollinations.ai
- All OpenAI generation params (temperature, max_tokens, top_p, frequency_penalty, reasoning_effort, thinking, logit_bias, etc.) are accepted and passed through to the underlying model
- Real token usage is accumulated across tool loop iterations and returned in the response
- System prompt: clarified tiers are sequential (can't skip), added polly/nomnom model identity so Polly doesn't confuse itself with the API model
- Tested live on production — `gen.pollinations.ai` with `model: "polly"` now works


Itachi-1824:
polly review please concisely!

PR #8603, opened 2026-02-25
Add tierbalance and paidbalance fields to /account/balance endpoint (only 10 lines added!)

XHyperDEVX:
Adds `tierbalance` and `paidbalance` fields to the `/account/balance` endpoint response while maintaining backward compatibility with the existing balance field.

### Example:
`{
  "balance": 8,
  "tierbalance": 3,
  "paidbalance": 5
}`
instead of just
`{
  "balance": 8
}`


voodoohop:
Thanks for the PR! Before we merge this — what's your use case for the balance breakdown?

We're considering exposing tier/pack splits but want to make sure the API shape matches real needs. Are you building something specific that needs this?

XHyperDEVX:
@voodoohop 
My final goal is to check whether the user of my app currently has paid tokens. If not, the option in my app that hides premium models is automatically activated.
I would also like to display a bar in the app showing how many tokens (free and paid) are still available. (Similar to the Pollinations Dashboard).

However, I generally welcome disclosure. An API that provides a lot of data is (in my opinion) a good API. Whether, who, and how it is then used is another matter, but it always offers more possibilities.

voodoohop:
@XHyperDEVX the reason we're not exposing those is that users may not want to expose their total balances to other apps. its a security/privacy consideration for the future.

https://github.com/pollinations/pollinations/pull/8684 check this PR as I think it solves the same problem.

XHyperDEVX:
@voodoohop
Am I understanding this correctly? Because you can set whether or not the balance can be accessed for each API key, right? If a user doesn't want that, they simply disable the permission for it and the app won't know about it... Or am I misunderstanding the problem?

#8684 looks good. It solves the problem at the API level, but without the freedom to decide for yourself. However, I also really like the “Unauthenticated requests still see all models” approach so as not to interfere with model list apps.

XHyperDEVX:
obsolete by #8684

PR #8626, opened 2026-02-26
feat: add xAI official API models (image + video)

ElliotEtag:
Adds three new models using the official xAI API (`api.x.ai`), replacing the unreliable airforce proxy for xAI generation.

**New models:**
- `grok-imagine-pro` — image via `POST /v1/images/generations` (paid, $0.02/image, 720p max)
- `grok-video-pro` — video via `POST /v1/videos/generations` with async polling (paid, $0.07/sec, 720p)
- `grok-video-pro-hd` — same but 1080p (paid, $0.14/sec)

**Implementation:**
- `xaiModel.ts` — maps width/height to closest xAI aspect ratio string
- `xaiVideoModel.ts` — submit → poll → download pattern with exponential backoff (3min timeout)
- `XAI_API_KEY` added to `secrets/env.json` (SOPS-encrypted, never plaintext)
- Registry entries with `paidOnly: true` and correct cost definitions
- 18 vitest tests covering endpoints, model IDs, aspect ratios, resolution mapping, return types

The existing `grok-video` (airforce) and `grok-imagine` (airforce) are kept untouched.

### DIFF
--- a/image.pollinations.ai/test/xaiModel.test.ts
+++ b/image.pollinations.ai/test/xaiModel.test.ts
@@ -7,7 +7,7 @@ let lastFetchBody: Record<string, unknown> = {};
 process.env.XAI_API_KEY = "test-xai-key";
 
 vi.stubGlobal("fetch", async (url: string, init?: RequestInit) => {
-    if (typeof url === "string" && url.includes("example.com")) {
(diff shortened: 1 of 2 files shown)


ElliotEtag:
Superseded by #9448 — updated pricing, fixed model IDs, added basic image tier, removed HD video split

Issue #61, opened 2021-08-19
Clean up ugly way we are constructing the form schema


voodoohop:
- [x] It was using the names of the variables to select what kind of form widget is shown. Should use the type, which we have access to already.
- [x] fix boolean checkbox not displaying default value

PR #8649, opened 2026-02-27
fix: tier_event pollen_amount Int32 → Float64, add tinybird logging

ElliotEtag:
- tier_event Tinybird datasource had `pollen_amount` as `Nullable(Int32)` but code sends floats (1.5 for spore, 0.1 for old microbe) — rows were silently quarantined
- Changed schema to `Nullable(Float64)`, added CAST in FORWARD_QUERY for backward compat
- Tinybird schema already deployed (deployment #61), verified float ingestion works
- Added logging to `sendBulkTierRefillEvents` so `waitUntil` errors are visible — logs skip reasons, Tinybird response (successful/quarantined rows), and fetch errors
- No data in tier_event since Feb 20; these changes + logging will help identify remaining root cause on next cron run


voodoohop:
polly please review

Issue #8406, opened 2026-02-19
Tier or Paid Banlance


Dtamtech:
Please specify on the https://gen.pollinations.ai/models AI models that need Paid or Tier pollen!

YoannDev90:
Done

PR #8684, opened 2026-02-28
feat: hide paid-only models from /models when user has no paid balance

voodoohop:
## Summary

- Authenticated users without paid balance (pack/crypto) no longer see `paid_only` models in `/v1/models`, `/text/models`, `/image/models`, `/audio/models`
- Unauthenticated requests still see all models (no change)
- Lightweight DB check — only queries `packBalance` + `cryptoBalance`, fails open on error

## How it works

- `checkPaidBalance(c)` returns `undefined` (no user), `true`, or `false`
- `filterModelsByPermissions` now accepts optional `hasPaidBalance` param
- Only filters when `hasPaidBalance === false` — unauthenticated and errors show everything

## Test plan

- [x] User with no paid balance → `claude-large` hidden from `/text/models`
- [x] User with paid balance → `claude-large` visible
- [x] Unauthenticated → all models visible
- [x] Existing `paid_only` field test still passes

Addresses #8406, #8010


voodoohop:
polly can you review

PR #8754, opened 2026-03-02
[PATCH]: Added fallback and heuristic to the PR-UPDATES realtime.

Circuit-Overtime:
## Changes Made:- 

* Added `MODEL_FALLBACK` and `IMAGE_MODEL_FALLBACK` constants in `common.py`, and updated all relevant functions to accept an optional model parameter and use these fallbacks if the primary model fails. 
* In `generate_realtime.py`, both PR analysis (`analyze_pr`) and image generation (`generate_gist_image`) now attempt to use the fallback model if the initial attempt fails, with clear logging for each step.
* In `publish_realtime.py`, snippet generation now tries the fallback model if the primary fails, and falls back to a plain format if both fail, with user-friendly logging.
* Updated imports in `generate_realtime.py` and `publish_realtime.py` to include the new fallback model constants.

### DIFF
--- a/social/scripts/common.py
+++ b/social/scripts/common.py
@@ -20,12 +20,12 @@ GITHUB_API_BASE = "https://api.github.com"
 POLLINATIONS_API_BASE = "https://gen.pollinations.ai/v1/chat/completions"
 POLLINATIONS_IMAGE_BASE = "https://gen.pollinations.ai/image"
 
-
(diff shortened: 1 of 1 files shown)


ElliotEtag:
polly can you review please?

Circuit-Overtime:
polly looks good to merge?

Issue #62, opened 2021-08-20
It seems many runs stop at image 20 for other people. 


voodoohop:
Havent seen this behavior myself yet though. A friend told me and I see really low iteration pollinations being posted too

<img width="1081" alt="Screenshot 2021-08-19 at 19 13 30" src="https://user-images.githubusercontent.com/5099901/130187917-a3c72fd2-37bb-4836-8668-238a7db81664.png">

The strange thing is these images are being posted to social media which means the notebook finishes its run.

An example can be seen here: https://pollinations.ai/p/QmQHT5LHmevpitV41s8E7oVVupQHd7j1nLSUyYRPZK7HJK

voodoohop:
Another example: https://pollinations.ai/p/QmPYaapMc9M1XdyrTjuJRCSxb8AXj1Hkp6ywmvw3AWzmjM

voodoohop:
It is due to the GPUs running out of memory since Google reduced the GPUs in their free tier. We need to reduce the usage of the notebooks.

voodoohop:
Ok. it was not completely clear but I definitely seeing a connection to the recent release of Colab+. I remember seeing a few people complaining about Colab Pro being downgraded to only one session at certain times. 

It seems we do have 16gb of ram but we needed to reduce the `cuts` parameter so the model uses a little less GPU

voodoohop:
Some instances were getting 8gb of rams so i switched the clip+vqgan to generating 256x256px images and then doing upscaling

PR #8756, opened 2026-03-02
add daily D1 to Tinybird sync for auth tables

ElliotEtag:
- Syncs user, apikey, session, account tables from D1 to Tinybird daily at 03:00 UTC
- Sensitive columns excluded at SQL level (tokens, keys, passwords, IPs never leave D1)
- GitHub Actions cron triggers `POST /api/admin/trigger-d1-sync` on the Enter worker
- Auth via `TINYBIRD_SYNC_TOKEN` (already set as CF secret + GH secret)
- Tinybird datasources already deployed (deployment #62)
- Per-table status reporting in the workflow output


ElliotEtag:
polly please review

ElliotEtag:
can you check with the last changes? polly

also are we sure that chuncking was necessary?

ElliotEtag:
**Note from token cleanup PR #8772:**

`TINYBIRD_SYNC_TOKEN` currently points to `d1_sync_token` which has `ADMIN` scope. This works but is overpowered — the sync only needs APPEND + DELETE on 4 datasources.

The delete rows API (`/v0/datasources/{name}/delete`) requires `DATASOURCES:CREATE` scope (Tinybird bundles row deletion under that scope). So the minimum viable token would be `DATASOURCES:CREATE` scoped to `d1_user`, `d1_apikey`, `d1_session`, `d1_account` — which also covers APPEND.

Consider creating a scoped token and replacing the ADMIN one when this PR is finalized.

PR #8816, opened 2026-03-04
fix: detect empty upstream responses and add fetch timeout

### LINKED ISSUE #8786: Mistral model too slow and frequently returns empty responses
**Bug Description**
The Mistral model is experiencing significant performance issues:
1. Response times are very slow
2. Frequently returns empty responses

**Expected Behavior**
Model should respond promptly with complete, non-empty outputs.

**Actual Behavior**
- Slow response times
- Empty responses returned "many times" / frequently

**Model**
- `mistral`

**Additional Context**
Reported by: `onetap05432` (Discord)

---
*Submitted via Pollinations Discord bot*

---
**Author:** `onetap05432` (UID: `1011999723328131152`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1478261175224500234/1478261491621826611)

samrahimi:
## Summary
- Detect empty/invalid responses from upstream providers (empty `choices` array, missing `message`) and throw 502 instead of silently returning 200 OK
- Add 30s fetch timeout via `AbortSignal.timeout()` to prevent requests from hanging indefinitely on slow upstream providers
- Replace silent empty-string fallback in `sendContentResponse` with 502 error

## Root Cause
When upstream providers (e.g., OVHcloud for Mistral) return empty `choices: []` or `choices: [{}]`, the code at `genericOpenAIClient.ts:161` silently converted this to `choices: [{}]` via `data.choices?.[0] ?? {}`. This propagated through the response pipeline as a valid 200 OK with no actual content. Additionally, no fetch timeout was configured, so slow upstream responses would hang indefinitely.

## Changes
- `genericOpenAIClient.ts`: Validate upstream response has non-empty choices with a message object; throw 502 on empty responses. Add 30s `AbortSignal.timeout()` to fetch, with proper 504 error on timeout.
- `server.ts`: Replace `return c.text("")` fallback with 502 error throw for empty message content.

## Test plan
- [ ] Verify Mistral model returns non-empty responses reliably
- [ ] Verify other models (openai, deepseek, gemini) still work correctly
- [ ] Verify timeout triggers 504 on slow responses instead of hanging
- [ ] Verify empty upstream responses return 502 error instead of empty 200

Fixes #8786

### DIFF
--- a/text.pollinations.ai/genericOpenAIClient.ts
+++ b/text.pollinations.ai/genericOpenAIClient.ts
@@ -18,6 +18,8 @@ import { cleanNullAndUndefined } from "./utils/objectCleaners.js";
 const log = debug("pollinations:genericopenai");
 const errorLog = debug("pollinations:error");
 
+const FETCH_TIMEOUT_MS = 30_000;
+
 function createApiError(
     response: { status: number; statusText: string },
     details: unknown,
@@ -103,6 +105,7 @@ export async function genericOpenAIClient(
             method: "POST",
             headers,
             body: JSON.stringify(requestBody),
+            signal: AbortSignal.timeout(FETCH_TIMEOUT_MS),
         });
 
         if (!response.ok) {
@@ -158,7 +161,24 @@ export async function genericOpenAIClient(
             `[${requestId}] Completed in ${Date.now() - startTime}ms, model: ${data.model || modelName}`,
         );
 
-        const originalChoice = data.choices?.[0] ?? {};
+        if (
+            !data.choices ||
+            data.choices.length === 0 ||
+            !data.choices[0].message
+        ) {
+            errorLog(
+                `[${requestId}] Empty or invalid response from upstream: choices=%j`,
+                data.choices,
+            );
+            const error = new Error(
+                `Empty response from upstream model ${modelName}`,
+            ) as ServiceError;
+            error.status = 502;
+            error.model = modelName;
+            throw error;
+        }
+
+        const originalChoice = data.choices[0];
         const formattedChoice = (
             formatResponse
                 ? formatResponse(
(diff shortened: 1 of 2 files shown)


samrahimi:
## Verification evidence

### Red-green cycle (mock upstream, `genericOpenAIClient` directly)

**Pre-fix (main @ 7a8a36b43):**
```
✗ choices:[] throws 502        → did not throw; got choices=[{}]
✗ choices:[{}] throws 502      → did not throw; got choices=[{}]
✗ missing choices throws 502   → did not throw; got choices=[{}]
✓ normal response passes
✓ tool_calls (content:null) passes
✓ 429 surfaces as 429
```

**Post-fix (this branch):**
```
✓ choices:[] throws 502        (status=502, "Empty response from upstream model...")
✓ choices:[{}] throws 502      (status=502)
✓ missing choices throws 502   (status=502)
✓ normal response passes       ({"role":"assistant","content":"hello"})
✓ tool_calls (content:null) passes  (tool_calls preserved — no false positive)
✓ 429 surfaces as 429          (not misclassified as 502/504)
✓ AbortSignal.timeout → error.name="AbortError" → caught at line 201 → 504
```
7/7 pass, 0 regressions.

### Live integration (real OVH Mistral, anonymous tier)
```
✓ [1/5] 789ms → "One."
✓ [2/5] 556ms → "two"
✓ [3/5] 561ms → "three"
~ [4/5] 429 rate limited (anon tier — correctly surfaced)
✓ [5/5] 740ms → "five"
```
0 silent-empty responses. Latency 0.56–0.79s.

### Lint/type
- `biome check`: clean
- `tsc --noEmit`: no new errors (1 pre-existing unrelated zod import in `shared/registry/model-info.ts`, present on main too)

### Note
30s timeout applies to all models including streams. `AbortSignal.timeout` counts from signal creation, so a legitimate stream exceeding 30s total would abort mid-stream. Not a concern for mistral; if o1/o3 non-streaming needs longer, follow-up config.

voodoohop:
@samrahimi in general good. id remove the timeout from this pr as some models take very long to respond

voodoohop:
Closing — targets the now-dead `text.pollinations.ai/` directory (removed by #10555). The underlying concerns were both deliberately decided in the *opposite* direction since: #11113 removed artificial fetch timeouts (the 90s cap was generating 1180+ synthetic 504s/day, now covered by a regression test), and #10968 removed the empty-completion 502 guard on purpose (a 200 OK from upstream should pass through — thin-client behavior). Porting either half would re-introduce reverted behavior, so no fresh issue is needed.

PR #8857, opened 2026-03-05
[PATCH]: Unified workflow for both "Latest News" and "Recent Apps" README sections

Circuit-Overtime:
## Changes:- 

- Added a single daily workflow (07:00 UTC) that updates both "Latest News" and "Recent Apps" README sections, then creates a PR.
- Added update_readme_local() function that reads highlights.md and README.md
- Renamed generate_highlights_and_readme() → generate_highlights(), returns only highlights content. 
- Removed create_readme_pr() function
- Updated to call generate_highlights() 
- app-review-submission.yml already doesn't update README directly
- app-update-readme.js is used as-is by the new workflow


ElliotEtag:
polly please review please, also i think the update readme py script should live in the git folder to follow the logics?

Circuit-Overtime:
The fixes that I have addressed: -

 - Dead code removed from update_readme.py and removed sys, base64, datetime, and common imports that were only used by dead code.
 - update_highlights.py kept as-is since its main() and create_highlights_pr() form a usable standalone script (can be run manually). Not called by any workflow but not strictly dead @ElliotEtag thoughts on this please! 
 - Inline Python replaced with CLI entrypoint — update_readme.py now has `if __name__ == "__main__"` with argparse that defaults --repo-root to the correct relative path. The workflow now just calls python social/scripts/update_readme.py
 - pip install step removed since update_readme.py no longer imports common.py or requests, it uses only uses internal libs nothing to install.
 - and haha nice catch on the step numbering i forgot but i have fixed it now

polly check now

Circuit-Overtime:
> Alright Elliot, suggested deletion:- 

## Changes Made:- 

- Moved 5 functions (load_gists_as_changelog, create_highlights_prompt, parse_highlights_response, merge_highlights, generate_highlights) + HIGHLIGHTS_PATH constant into generate_daily.py
- Added missing imports to generate_daily.py: os, get_repo_root, read_news_text_file, OWNER, REPO
- Renamed parse_response → parse_highlights_response to avoid potential name conflicts
- Deleted update_highlights.py
- Updated comment in buffer-schedule.yml that referenced the deleted file

Polly give me a final review

ElliotEtag:
polly please make a final review

PR #8946, opened 2026-03-08
Update App.jsx in virtual makeup app 

Sugamdeol:
This PR migrates the image upload functionality in the `virtual-makeup` app from Cloudinary to the new Pollinations media service (`media.pollinations.ai`).

Key changes:
- Removed all Cloudinary-related configuration and code.
- Integrated the Pollinations media upload API using a publishable key.
- Improved error handling by adding a user-visible error message and a shake animation for better feedback.
- Verified that the app still builds and loads correctly through a local preview and automated screenshot.

### DIFF
--- a/image.pollinations.ai/src/createAndReturnImages.ts
+++ b/image.pollinations.ai/src/createAndReturnImages.ts
@@ -1139,6 +1139,7 @@ const generateImage = async (
         case "flux-2-dev":
         case "imagen-4":
         case "grok-imagine":
+        case "dirtberry":
             return await callAirforceImageAPI(
                 prompt,
                 safeParams,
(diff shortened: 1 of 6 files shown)


Circuit-Overtime:
hii @Sugamdeol rather than a publishable api, we can also use the BYOP feature of pollinations if you wanna update

Sugamdeol:
It uses nano banana which is paid so maybe all users are not able to use it through BYOP, I will change the model to gptimage and add BYOP @Circuit-Overtime .

Circuit-Overtime:
Yes gptimage or nanobana itself works!

voodoohop:
polly can you review?

PR #8975, opened 2026-03-09
Add simple prompt example for image generation

developerasaad:
Adds a small prompt example showing how to generate an image with Pollinations.


Circuit-Overtime:
Hii @developerasaad we already have this feature in our [API DOCS](https://gen.pollinations.ai/api/docs)

PR #8985, opened 2026-03-09
perf: optimize seed upgrade workflow (25h → seconds)

Itachi-1824:
The D1 upgrade phase was spawning a Node subprocess per user (~5s each). With 18k approved users, that's ~25 hours. The GitHub API validation was also unnecessarily slow with fixed 2s sleeps.

- Replaces per-user `npx tsx` subprocess with batch SQL UPDATE (500 users per query)
- Adds concurrent GitHub API validation (3 parallel workers via ThreadPoolExecutor)
- Implements adaptive rate limiting based on X-RateLimit-Remaining header (no sleep when quota > 100, 1s when > 50, 2s otherwise)
- Removes unused Polar delay (tier-update-user.ts doesn't call Polar)
- Shows API quota remaining in progress bar

Expected improvement: ~25h upgrade phase → seconds, ~30min validation → ~10min. Same GraphQL call count, just less idle time.


Itachi-1824:
polly review thoroughly yet concisely!

Itachi-1824:
polly now?

PR #9071, opened 2026-03-13
fix: tier upgrade defers balance refill to daily cron

ElliotEtag:
Tier upgrades were immediately setting tier_balance to the new tier's pollen amount. This bypassed the daily cron refill schedule.

- Removes `tier_balance` assignment from `tier-update-user.ts` (used by app-submission workflow)
- Removes `tier_balance = 3.0` from batch SQL in `user_upgrade_spore_to_seed.py`
- Both paths now only update the `tier` column — balance is set by the midnight UTC cron refill


ElliotEtag:
polly review please

PR #9092, opened 2026-03-14
Add Playground

GamerC0der:
- The playground is now available via a new tab in the Pollinations Enter site.
- You can create text, image, audio, and video in one place!
- Advanced settings like seed are available!
- Uses your pollen balance!
- Enables users to test out the Pollinations API without leaving their dashboard.


voodoohop:
Thanks for the effort\! We're keeping the dashboard minimal for now, so we won't be adding a playground. Closing this one out.

PR #9121, opened 2026-03-14
fix: construct Azure GPT Image endpoints from base URL

voodoohop:
## Summary
- Construct endpoint URLs from base deployment URL + path + api-version instead of regex-replacing parts of the full URL
- Handles both legacy full-URL env vars and base-only URLs
- Cleaner follow-up to #9065

## Test plan
- [ ] gptimage text2img still works (generations endpoint)
- [ ] gptimage img2img/edit still works (edits endpoint with correct api-version)
- [ ] gptimage-large works in both modes


voodoohop:
polly please review

PR #9128, opened 2026-03-14
fix: pass image URLs directly to Pruna p-image-edit

voodoohop:
## Summary
- Pruna API rejects base64 data URIs with misleading "property input is required" error
- Pass image URLs directly instead of downloading and re-encoding to base64
- Confirmed: URLs work, data URIs don't

## Test plan
- [x] Confirmed via direct API call: URLs accepted, data URIs rejected
- [ ] p-image-edit integration test should pass after deploy


voodoohop:
polly please review

voodoohop:
polly please review the latest changes — now also fixes p-video and cleans up dead code

PR #9140, opened 2026-03-15
fix: skip redundant KV writes for API key verify tracking

### LINKED ISSUE #9125: fix: request deduplication skips billing for deduplicated requests
- Dedup middleware runs before track middleware in the stack
- When request B hits the dedup cache, it returns cached response directly and skips `next()`
- Track middleware's `waitUntil` deduction only runs for the original request, not deduplicated ones
- **Deduplicated requests are never billed**

Additionally:
- TOCTOU race between balance check (pre-request) and deduction (`waitUntil` post-response) allows concurrent requests to pass the check before any deduction runs
- The deduction SQL allows negative balances by design (no `WHERE balance >= amount` guard)
- `waitUntil` deduction failures are caught and swallowed — if D1 errors, balance never decrements

**Key files:**
- `enter.pollinations.ai/src/middleware/track.ts` (line 145) — deduction in `waitUntil`
- `enter.pollinations.ai/src/utils/balance-deduction.ts` (lines 16-46) — SQL allows going negative
- Request deduplication middleware — returns cached response, skips billing entirely

voodoohop:
## Summary

- Better Auth writes 3 KV PUTs per `verifyApiKey` call (tracking `lastRequest`, `requestCount`, `updatedAt`) — purely for dashboard display
- At 3-5M req/day this exceeds KV write limits, flooding logs with `KV PUT failed: 429 Too Many Requests`
- Fix: only allow the first KV write per API key per isolate (cache-warm on miss), skip subsequent verify tracking writes
- Scoped to `api-key:` prefix — session writes are unaffected
- Dashboard reads `lastRequest` from D1 (already updated via `deferUpdates`)

## What changed

Single file: `enter.pollinations.ai/src/auth.ts`
- Module-level `Set` tracks which API keys have been written to KV
- `set()`: skip if key starts with `api-key:` and already written in this isolate
- `delete()`: clears the Set entry so next access re-warms the cache

## Test plan

- [ ] Deploy and monitor KV write volume in Cloudflare dashboard
- [ ] Verify API key auth works for both `sk_` and `pk_` keys
- [ ] Check `wrangler tail` for absence of KV 429 errors
- [ ] Verify key creation/deletion still works (admin routes)

Fixes #9125

### DIFF
--- a/enter.pollinations.ai/src/auth.ts
+++ b/enter.pollinations.ai/src/auth.ts
@@ -15,6 +15,14 @@ import { user as userTable } from "./db/schema/better-auth.ts";
 import { sendTierEventToTinybird } from "./events.ts";
 import { DEFAULT_TIER, getTierPollen } from "./tier-config.ts";
 
+// Track which API keys have been written to KV in this isolate.
+// Better Auth writes tracking data (lastRequest, requestCount, updatedAt) to KV
+// on every verify call. At ~3-5M req/day this exceeds KV's write limits (429s).
+// First write per key (cache-warm on KV miss) goes through; subsequent writes
+// (verify tracking updates) are skipped. Session writes are unaffected.
+// Dashboard reads lastRequest from D1 (updated via deferUpdates).
+const kvWrittenKeys = new Set<string>();
+
 function addKeyPrefix(key: string) {
     return `auth:${key}`;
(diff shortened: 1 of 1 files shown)


voodoohop:
polly please review

PR #9144, opened 2026-03-15
fix: check response content-type before SSE parsing for billing

voodoohop:
## Summary

- **Throw 502** when `stream: true` requested but upstream returns non-SSE content-type (e.g. `application/json`)
- **Prevents caching broken responses** — 502 naturally skips cache (previously 290-byte JSON blobs were cached as "streaming" responses)
- **Add model name** to all warn/error log properties in trackResponse
- **Safety net** in trackResponse: content-type check returns `isBilledUsage: false` if the 502 somehow doesn't fire
- **Add text service mock** to test fixtures with `forceNonStreaming` flag
- **Regression tests**: 502 for non-SSE upstream, correct billing for real SSE

### Production evidence

Caught live: upstream intermittently returns 200 JSON (2 chunks, 290 bytes) for `stream: true` requests. ~3 occurrences per minute in bursts. Multiple IPs/users affected.

\`\`\`
ERROR "No usage object found in event stream"  (hono:track:stream)
ERROR "Failed to extract model usage"          (hono:track:response)
\`\`\`

## Test plan
- [x] 147/147 text integration tests pass (1 pre-existing failure on main unrelated)
- [x] New test: 502 when stream requested but upstream returns JSON
- [x] New test: correct billing when stream requested and upstream returns SSE
- [x] Verified production occurrence via wrangler tail

### DIFF
--- a/enter.pollinations.ai/src/routes/images.ts
+++ b/enter.pollinations.ai/src/routes/images.ts
@@ -6,6 +6,9 @@
 import type { Context } from "hono";
 import type { ContentfulStatusCode } from "hono/utils/http-status";
 import { getDefaultErrorMessage, UpstreamError } from "@/error.ts";
+import type { AuthVariables } from "@/middleware/auth.ts";
+import type { BalanceVariables } from "@/middleware/balance.ts";
+import type { ModelVariables } from "@/middleware/model.ts";
 import {
     type CreateImageEditRequest,
     CreateImageEditRequestSchema,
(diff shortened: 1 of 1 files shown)


voodoohop:
polly please review

voodoohop:
Updated — now throws instead of silently falling back:
- `Stream requested but upstream returned content-type: application/json`
- Removed the header-extraction fallback — if stream was requested but upstream returned JSON, that's a bug we want to surface
- Added regression test for SSE streaming billing (happy path)
- Removed the non-SSE fallback test (can't assert on waitUntil throws in CF test framework)

polly please review

voodoohop:
**Production evidence** — caught live during log tailing:

```
2026-03-15T11:15:42 ERROR "No usage object found in event stream"
  logger: hono:track:stream
  requestId: 3956ba0f-506f-44db-b66a-cd59fa99e7e5
  IP: 2a06:98c0:3600::103

2026-03-15T11:15:42 ERROR "Failed to extract model usage"
  logger: hono:track:response
  (same requestId)
```

- Same IP from previous investigation — sends `stream: true` but upstream sometimes returns non-SSE JSON
- Intermittent: appeared within 20s of first tail, then nothing for 90s+ on subsequent tails
- With the throw, this becomes an explicit error instead of a silent billing leak

voodoohop:
polly please review

PR #9074, opened 2026-03-13
feat: hourly pollen refills for spore/seed

voodoohop:
## Summary
- Spore: 1.5p/week → 0.01p/hour (~72% cost savings)
- Seed: 3p/day → 0.15p/hour (~40% cost savings)
- Flower/Nectar/Router: unchanged (daily)
- Cron runs hourly, separate idempotency for hourly vs daily tiers
- UI updated to show per-hour amounts

## Test plan
- [ ] Deploy to staging, trigger refill, verify spore/seed get hourly amounts
- [ ] Verify flower/nectar still get daily amounts
- [ ] Confirm idempotency: second call within same hour skips hourly tiers

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9145, opened 2026-03-15
feat: hourly pollen refills for spore/seed via Cloudflare Cron

voodoohop:
## Summary

- Hourly pollen refills for spore/seed via Cloudflare Cron (`0 * * * *`)
- Daily refills for flower/nectar/router (unchanged)
- Additive refill: `MIN(balance + amount, cap)` — negative balances recover gradually instead of instant reset
- Pre-request balance check: rejects requests when `availableBalance < estimatedCost` (merged from #9152)
- Prevent negative balance spillover: when paid balance (crypto+pack) ≤ 0, deductions always go to tier
- Microbe: cadence "none" (no refill)
- UI text, FAQ, pricing page updated for hourly/daily model
- Refill logic extracted into `runTierRefill()` callable from both HTTP and scheduled handler

## Balance protection stack

1. **Pre-request cost check** — blocks if balance < estimated model cost (7-day avg from Tinybird)
2. **Additive refill** — going negative is a natural penalty that recovers over time
3. **Spillover guard** — tier negatives never leak into paid (crypto/pack) balances

## Test plan

- [x] 36 integration tests passing (tier-balance + balance-check)
- [x] Staging deploy verified: balance deductions, spillover protection, 402 rejections
- [x] Hourly cron fires on staging (`0 * * * *`)

Closes #9043
Supersedes #9074
Includes #9152

### DIFF
--- a/enter.pollinations.ai/test/integration/tier-balance.test.ts
+++ b/enter.pollinations.ai/test/integration/tier-balance.test.ts
@@ -165,7 +165,7 @@ describe("Tier Balance Management", () => {
             // Setup: users with negative balances
             const testUsers = [
                 { id: "neg-spore", tier: "spore", tierBalance: -0.005 },
-                { id: "neg-seed", tier: "seed", tierBalance: -0.50 },
+                { id: "neg-seed", tier: "seed", tierBalance: -0.5 },
                 { id: "neg-flower", tier: "flower", tierBalance: -5 },
                 { id: "neg-nectar", tier: "nectar", tierBalance: -15 },
             ];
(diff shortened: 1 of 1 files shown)


voodoohop:
polly please review - merged pre-request balance check from #9152, added additive refill (negative balances recover gradually), and spillover fix

ElliotEtag:
### Code review

Found 4 issues:

1. **Daily tier refill changed from hard-reset to additive without test coverage for negative balances.** The SQL uses `MIN(COALESCE(tier_balance, 0) + X, X)` instead of a direct reset. A flower user at `-5` pollen gets `MIN(-5+10, 10) = 5` instead of the full `10`. The policy flip-flopped across commits in this branch (hard reset -> additive), but tests only cover users starting at positive balances, so this behavior change is untested. Comment on line 220 still says "resets pollen" but the code is additive.

https://github.com/pollinations/pollinations/blob/fba06ec2c9f7d20acc15816bf1598d80a791fa6b/enter.pollinations.ai/src/routes/admin.ts#L220-L232

2. **Pre-request balance check is more restrictive than the deduction logic.** `getAvailableBalance()` sums only positive buckets via `Math.max(0, ...)`, returning `0` when all balances are negative. But `atomicDeductUserBalance` allows deducting from tier even when negative (the "no paid balance -> always tier" path). A free-tier user with `tierBalance: -0.01` gets a 402 from the pre-check, even though the deduction would have allowed the request. This blocks spore/seed users between hourly refills more aggressively than before.

https://github.com/pollinations/pollinations/blob/fba06ec2c9f7d20acc15816bf1598d80a791fa6b/enter.pollinations.ai/src/routes/proxy.ts#L975-L990

3. **`getNextRefillAt` falls back to `"daily"` cadence for null/undefined tier instead of `"none"`.** When `tier` is falsy, the code defaults to `"daily"` and returns a midnight UTC timestamp -- implying a daily refill is coming. For users with no tier (or microbe, which has `cadence: "none"`), the correct fallback is `"none"` (returning `null`). This sends misleading data on the `/account/profile` endpoint.

https://github.com/pollinations/pollinations/blob/fba06ec2c9f7d20acc15816bf1598d80a791fa6b/enter.pollinations.ai/src/routes/account.ts#L15-L18

4. **`TIER_POLLEN` values now have mixed time dimensions.** Spore (0.01) and seed (0.15) are per-hour amounts, while flower (10), nectar (20), and router (500) are per-day amounts. The same `TIER_POLLEN.X` constant is used as both the refill increment and the cap in `MIN(balance + X, X)`. Any code that reads `TIER_POLLEN` values assuming a uniform cadence will produce wrong results. The cap for hourly tiers is effectively the per-hour amount (e.g., spore can never exceed 0.01 pollen), which may be more restrictive than intended.

https://github.com/pollinations/pollinations/blob/fba06ec2c9f7d20acc15816bf1598d80a791fa6b/enter.pollinations.ai/src/tier-config.ts#L1-L19

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #9147, opened 2026-03-15
fix(enter): model list cut off in edit key dialog

fisventurous:
fixes the model list being clipped in the edit key dialog when there are many models. 

the inner model panel (`account-permissions-input.tsx`) used `overflow-hidden` for its expand/collapse animation, which prevented the list from scrolling when it grew past the modal boundary. this occurred during #9044

\- add `overflow-y: auto` and `max-height: 50vh` to the model list container so it scrolls independently
\- uses `scrollbar-width: thin` to keep the scrollbar subtle


voodoohop:
@ElliotEtag could you review this one?

PR #9219, opened 2026-03-17
fix: invalidate KV cache when API key permissions change

ElliotEtag:
When only `allowedModels` was updated on an API key (no name/budget/expiry changes), the KV cache wasn't invalidated. `verifyApiKey` would serve stale data without the new permissions, making model restrictions silently ineffective.

- Moves KV cache invalidation to trigger on permission changes, not just D1 field updates
- Fixes failing `text.test.ts` "Restricted API key should deny access" test
- Also fixes `api-keys.test.ts` "should enforce model restrictions" test

All 158 text integration tests pass after this fix.


ElliotEtag:
polly review please

voodoohop:
polly please review the latest simplification commit

PR #9220, opened 2026-03-17
fix(enter): prevent wrapping in dev points tooltip

fisventurous:
this fixes:
image

\- increase tooltip width to `w-80`
\- add `whitespace-nowrap` to table cells to prevent ugly layout


fisventurous:
@ElliotEtag 😬

Issue #9255, opened 2026-03-18
pollinations account issues


ABunchOfRibbons:
hello! so basically, I've submitted an app yesterday, but it didn't take since apparently _**I haven't registered into pollinations with my github account.**_ but no, I do. 

I think there may be some issues since I've recently changed my github accounts username, and everytime I try to re-login with **THIS** account, the user still says my old username in pollinations.

yes, I logged out and logged in again. yes, I also logged out of github and logged in again. no, it didn't fix it, and now it seems like even though I've already logged into pollinations with this github account, it isn't.

[link](https://github.com/pollinations/pollinations/issues/9237) to the app I tried to submit. Check the comments. Someone said that I haven't registered this github account to pollinations.

PR #9267, opened 2026-03-18
fix(dashboard): read githubUsername from D1 profile endpoint

fisventurous:
fixes stale github username showing on dashboard after relog

I believe the session object is KV-cached at login time, so `githubUsername` could be stale even after the background sync updated D1. the dashboard now fetches it directly from `/api/account/profile` (D1) in parallel with the other loader calls, with a fallback to the session value if the fetch fails.

\- add profile fetch to `Promise.all` in the route loader
\- derive `githubUsername` from D1 result, falling back to session value
\- pass `githubUsername` from loader data instead of `user.githubUsername`

// #9255


ElliotEtag:
polly review please

ElliotEtag:
The Problem (well-identified)
The bug is real and well-understood. Here's the chain:

User logs in → Better Auth creates a session, caches user data (including githubUsername) in KV
onAfterSessionCreate fires via waitUntil() — it's non-blocking, meaning it runs after the session is already cached
That background task syncs the username from GitHub API → writes updated value to D1
But the KV-cached session still holds the old username
Dashboard reads context.user.githubUsername which comes from the session (KV) → shows stale value
This is a classic race condition between a background sync and a cached read path.

The Solution (mostly sound, with caveats)
Adding a /api/account/profile fetch to the loader's Promise.all and preferring the D1-sourced githubUsername over the session value is architecturally correct. D1 is the source of truth, and the profile endpoint reads directly from D1.

The fallback chain (profileResult?.githubUsername ?? context.user?.githubUsername ?? "") is sensible — if the profile fetch fails, you still show something.

Concerns
1. Timing — the race isn't fully eliminated

The onAfterSessionCreate sync runs via waitUntil() — it's fire-and-forget. On the very first page load after login, the D1 write from the sync might not have completed yet. The profile fetch could still return the old username. This is a narrow window but worth acknowledging. The PR improves the situation significantly (subsequent page loads will be correct), but doesn't guarantee correctness on the initial post-login render.

2. Extra network roundtrip on every page load

The loader now makes 4 parallel fetches instead of 3. Since /api/account/profile hits D1 directly (no cache), this adds latency and D1 read load on every dashboard visit. For something that only matters when a user renames their GitHub account (extremely rare), this is a cost paid on every single page load by every user.

Alternative worth considering: Invalidate the KV session cache after the username sync writes to D1. That way context.user would naturally have the fresh value on next load, and you wouldn't need the extra fetch. Or use a shorter TTL on the session KV entry.

3. Raw fetch instead of the typed apiClient

The existing loader uses apiClient.tiers.view.$get(), apiClient["api-keys"].$get(), etc. — typed Hono RPC clients. The PR adds a raw fetch("/api/account/profile", { credentials: "include" }) which:

Bypasses type safety (response is untyped)
Is inconsistent with the existing pattern
Requires manual JSON parsing and null-checking
If apiClient.account.profile.$get() (or similar) is available on the Hono client, that should be used instead.

4. Error handling — silent degradation

The fallback to context.user?.githubUsername when the fetch fails is good. But depending on how it's implemented, a network error or 500 from the profile endpoint could throw and break the entire loader (which would show an error page instead of the dashboard). The .then() / .catch() pattern needs to be airtight — the way the other fetches use .then((r) => (r.ok ? r.json() : null)) is the right defensive pattern. The PR should follow the same pattern.

5. Scope creep — only githubUsername?

If the session can be stale for githubUsername, it can be stale for image (avatar URL), tier, name, etc. The PR only patches one field. This isn't necessarily wrong (fix the reported bug), but it raises the question: is fetching the whole profile and then cherry-picking one field the right granularity? If you're fetching the profile anyway, should the component use profileResult.image too?

Verdict
The diagnosis is correct. The fix works. But it's treating the symptom, not the root cause.

The root cause is that onAfterSessionCreate writes to D1 but doesn't invalidate the KV session cache. A cleaner fix would be to update the KV-cached session after the D1 write completes inside onAfterSessionCreate, so that context.user is always fresh. That would be zero extra API calls, zero extra latency, and would fix staleness for all fields, not just githubUsername.

If the KV session invalidation approach is too complex (Better Auth may not expose a clean API for it), then the PR's approach is a reasonable pragmatic fix — but the implementation should use the typed client and follow the existing .then(r => r.ok ? r.json() : null) pattern for consistency and safety.

@fisventurous 
thoughts?

fisventurous:
@ElliotEtag makes sense. the issue seems to be that the dashboard is reading `context.user` from the better-auth session, while the github username sync happens later in `onAfterSessionCreate` through `waitUntil()`

your KV invalidation idea is probably the cleaner long-term fix, but I still lean toward keeping this PR as the smaller fix. mostly because the auth caching is sitting inside better-auth, and from this hook we only have `session.userId`, so updating D1 is straightforward but refreshing the cached session cleanly looks like a wider auth change rather than a small dashboard change

for the reported bug, fetching the profile from D1 in the dashboard loader still feels reasonable to me. it is narrower and it directly uses the source of truth without touching the auth layer more deeply.

I should probably change the implementation details in this PR by using the typed `apiClient` instead of raw `fetch`, and keep the same defensive non-throwing pattern as the other loader calls.

your note about the first post-login load is clear and fair too. this improves the behaviour, but it does not fully guarantee that the very first render after login will already have the updated username if the background sync has not finished yet.

so overall, I see your suggestion as definitely the better root-cause direction, but probably better as a follow-up(?) than something I would fold into this PR. this PR only changes the dashboard read path, while KV invalidation changes auth/session behaviour more broadly. if we do that separately, we can then decide whether this extra profile fetch is still needed

what do you think?

ElliotEtag:
Yes! I think that scope split makes sense @fisventurous 

You know the auth/cache side better than I do, so if KV invalidation is a broader Better Auth change, I’m fine treating that as follow-up work instead of folding it into this PR.

I’d still tighten the current implementation here:

- switch the profile request to the typed apiClient so it matches the rest of the loader
- keep the same non-throwing `r.ok ? r.json() : null` pattern so `/account/profile `failures don’t take down the dashboard
- maybe call out in the PR that the first render right after login can still race the background sync


With those adjustments, this feels like a reasonable pragmatic fix.

fisventurous:
updated as requested. swapped to the typed `apiClient` for the profile request, so it matches the rest of the loader now. 
and genuinely, no, that comparison isn't fair to you, please.😅 I'm just an observer with some scattered knowledge.

quick note: fetching in the loader means there's a small chance the very first render after login races the background sync. subsequent loads should pull the fresh username fine though. 👀

ElliotEtag:
- Extra D1 read on every page load — The /api/account/profile endpoint hits D1 directly (no cache). This is a new read on every dashboard visit to fix a scenario that happens almost never (user renames GitHub account). The cost is low (D1 reads are fast and cheap), but it's worth noting.

- Only fixes githubUsername — The session cache can also be stale for image (avatar URL). Since the profile endpoint already returns image, you could also prefer profileResult?.image over context.user?.image for free. Not a blocker, but a missed opportunity.


what do you think of those?

fisventurous:
re point 1) no extra D1 read added. the previous iteration manually fetched `/api/account/profile` in the loader, but that's been removed. loader now calls `tiers.view`, `api-keys`, and `customer.balance` as before. user object comes from `context.user` (better-auth session), so no new D1 hit per page load.

re point 2) with no profile fetch in the loader, there's no `profileResult` to pull a fresh image from. adding one back would contradict point 1. avatar falls back to `context.user?.image`, which updates on next login, same as username sync. this is a session-level concern, not something to fix per-request.

the username staleness issue that motivated this PR was fixed server-side. this PR just migrates the client loader from raw fetch to the typed `apiClient`.

PR #9321, opened 2026-03-19
Feat/update reimagine app

CloudCompile:
updating the reimagine app (pr 1/3, updating some of the other apps in the /apps folder too)

issue https://github.com/pollinations/pollinations/issues/8750


voodoohop:
polly please review

voodoohop:
## Code Review

### Summary
Adds Pollinations auth (BYOP flow) to the Reimagine app, removes ImgBB dependency, and cleans up React patterns (refs for stale closure fixes, memoization improvements). Also includes two unrelated tsconfig changes.

### Bugs / Blockers

- **`ProfileScreen.tsx` references undefined variables `token`, `login`, `logout`, `toggleTheme`, `isDark`** — The diff removes `isDark` and `toggleTheme` destructuring from `useTheme()` and never imports `useAuth`, but the template still uses `token`, `login`, `logout`, `toggleTheme`, and `isDark`. This will crash at runtime. Needs:
  ```tsx
  const { theme, isDark, toggleTheme } = useTheme();
  const { token, login, logout } = useAuth();
  ```
  and the `useAuth` import.

- **`ImageUploadService.ts` — broken upload URL** — `https://api.image.com/1/upload` is not a real service. The ImgBB references were renamed to generic `image` placeholders but the actual replacement upload endpoint was never wired in. The `imageResponse` interface still matches the ImgBB response shape, so this will fail against any other API. If the intent is to remove ImgBB entirely, this service needs a real replacement (or should be removed and image uploads handled differently).

- **`image_API_KEY` config removed but still referenced** — `IMGBB_API_KEY` was removed from `app.config.js` extras, but the new `image_API_KEY` was never added. `Constants.expoConfig?.extra?.image_API_KEY` will always be `undefined`, falling back to the placeholder `'xxxxxxxxxxxxxxxxxxxxxxx'`.

### Issues

- **Hardcoded publishable key in `AuthService.ts`** (`pk_IKINQsX7rwBZkifu`) — Should come from app config / env variable rather than being hardcoded in source. Makes rotation or per-environment config impossible.

- **Wildcard dependency versions** — `expo-auth-session`, `expo-secure-store`, `expo-web-browser` are pinned to `"*"`. Other expo deps use exact ranges (e.g., `~16.1.4`). Use compatible Expo SDK 52 versions to avoid breaking installs.

- **`uploadImage` vs `uploadToimage` naming inconsistency** — `ImagePreparationService` and `TransformationDetailScreen` call `ImageUploadService.uploadImage()` but the method is actually named `uploadToimage()`. This will be a runtime error.

### Suggestions

- **Unrelated changes should be in separate PRs**: `media.pollinations.ai/tsconfig.json` and `social/reddit/tsconfig.json` changes (removing `@cloudflare/workers-types`, removing `"node"` types, adding `noImplicitAny: false`) are unrelated to the reimagine app update and could break those services. The reddit tsconfig removing `"node"` types will cause TS errors for any Node.js API usage.

- **`selectionVersion` change from number to string** in `HomeScreen.tsx` — verify that all consumers handle the new string type (was previously a number incremented via `useState`).

- **Auth error handling** — `login()` in `AuthContext` silently swallows failures (sets `isLoading: false` with no error state). Consider exposing an error state or showing an alert on auth failure.

- **`WebBrowser.maybeCompleteAuthSession()` at module scope** — This is fine for Expo, but add a comment noting it must run before any component mounts for the redirect to complete.

### What looks good
- Clean auth context/provider pattern
- Ref-based stale closure fix in `HomeScreen` for `fetchTrendingImages`
- Removing unused `isDark` prop from `TransformationItem`
- Adding missing newlines at EOF

---
*Review by Claude (triggered manually)*

CloudCompile:
image
@voodoohop does this matter?

voodoohop:
@CloudCompile thats a bug in our workflows that dont work 100% when you are working in a remote fork. need to fix it

voodoohop:
@CloudCompile did you check the review?

CloudCompile:
oh sorry @voodoohop didn't see this I had to go to my tennis tournament, will get to work in about half an hour

CloudCompile:
also I think I did do that just didn't push the changes from my macbook

CloudCompile:
addressed the problems @voodoohop

PR #9329, opened 2026-03-20
Improve error handling 

hugow0528:
Bug Fixes
image.pollinations.ai/src/util.ts — Added getErrorMessage(error: unknown): string utility that safely extracts a string message from any thrown value (Error objects, plain strings, or other primitives).

image.pollinations.ai/src/createAndReturnImages.ts — Fixed 22 instances of unsafe error.message access in catch blocks. In TypeScript/JavaScript, caught values are typed as unknown, meaning accessing .message directly can crash if a non-Error value is thrown (e.g. a plain string, number, or object). All instances now use the new getErrorMessage() helper.

text.pollinations.ai/server.ts — Added a type guard before Buffer.from(audio.data as string, "base64"). The code was casting audio.data to string without validation. If the API returns audio.data as a non-string (e.g., an object or number), Buffer.from would either crash or produce garbage. Now it throws a clear error if the value isn't a string.

text.pollinations.ai/transforms/imageUrlToBase64Transform.ts — Added nullish coalescing (?? "Unknown error") on two error.message accesses. The error is typed with an optional .message property, so interpolating it directly could produce "undefined" in log messages.

Code Optimization
text.pollinations.ai/auth/googleCloudAuth.ts — Added a tokenRefreshScheduled boolean flag to guard against duplicate setInterval registrations. While Node.js module caching normally prevents this, the flag makes the intent explicit and prevents accidental double-registration if the initialization code is ever refactored.

### DIFF
--- a/enter.pollinations.ai/src/routes/proxy.ts
+++ b/enter.pollinations.ai/src/routes/proxy.ts
@@ -505,7 +505,7 @@ export const proxyRoutes = new Hono<Env>()
         validator(
             "param",
             z.object({
-                prompt: z.string().min(1).meta({
+                prompt: z.string().min(1).max(100_000).meta({
                     description: "Text prompt for generation",
                     example: "Write a haiku about coding",
                 }),
@@ -597,7 +597,7 @@ export const proxyRoutes = new Hono<Env>()
         validator(
             "param",
             z.object({
-                prompt: z.string().min(1).meta({
+                prompt: z.string().min(1).max(10_000).meta({
                     description: "Text description of the image to generate",
                     example: "a beautiful sunset over mountains",
                 }),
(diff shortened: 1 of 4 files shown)


fisventurous:
thanks for the effort on the error handling patches @hugow0528 ! it's great to see more robustness in the API.

however, we need to split this up a bit to keep things manageable. mixing a codebase-wide refactor with 750 lines of a new app is a lot for one PR.

regarding the Pollination Hub: it's already looking good in `APPS.md` (lines 272 and 742), and usually community apps stay in their own repos while the `apps/` directory here is kept for internal tools and core collaborations. it's better for everyone if it stays in your repo so you have full control over it.

also, since `image.pollinations.ai` is a legacy service (returns 410 gone), maybe focus the error handling improvements strictly on the active `enter` and `text` modules instead?

if you can remove the `apps/pollination-hub` folder and just resubmit the necessary fixes for the current modules in a minimal PR, that would be much easier for us to review and get merged. cheers.

hugow0528:
@fisventurous 
thank you for your suggestion,i submit a new Pull Requests now : https://github.com/pollinations/pollinations/pull/9394
please check

PR #9389, opened 2026-03-21
fix(image): pass seed to vertex ai models

fisventurous:
fixes an uncommon issue where during image edits with nano banana variants kept returning the exact same image even with 'varied' seeds.

changes
\- pass the `seed` parameter down to vertex AI API requests for `nanobanana` and `nanobanana-2` models (added a future-proof seed passing for `nanobanana-pro` IN CASE they enable it, which is unlikely now, but still this does no harm).

since the seed was never added, vertex AI _sometimes_ received identical payloads on every regen and either hit an internal deduplication cache or anchored too hard to the base image. although optional, passing the seed explicitly ensures and forces the API to calculate a new distinct edit on each retry. varied noises!


ElliotEtag:
Suggestions
In vertexAIImageGenerator.ts, consider using the same conditional spread pattern used in the client: ...(processedParams.seed !== undefined && { seed: processedParams.seed }) instead of always passing seed (which may be undefined with an unsafe cast).
The as number cast masks a potential undefined value — prefer letting TypeScript enforce type safety here.

fisventurous:
yes, good one. updatedto use the conditional spread pattern for the `seed` parameter as suggested.

PR #9394, opened 2026-03-21
Enhance error handling

hugow0528:
Bug Fixes
image.pollinations.ai/src/util.ts — Added getErrorMessage(error: unknown): string utility that safely extracts a string message from any thrown value (Error objects, plain strings, or other primitives).

image.pollinations.ai/src/createAndReturnImages.ts — Fixed 22 instances of unsafe error.message access in catch blocks. In TypeScript/JavaScript, caught values are typed as unknown, meaning accessing .message directly can crash if a non-Error value is thrown (e.g. a plain string, number, or object). All instances now use the new getErrorMessage() helper.

text.pollinations.ai/server.ts — Added a type guard before Buffer.from(audio.data as string, "base64"). The code was casting audio.data to string without validation. If the API returns audio.data as a non-string (e.g., an object or number), Buffer.from would either crash or produce garbage. Now it throws a clear error if the value isn't a string.

text.pollinations.ai/transforms/imageUrlToBase64Transform.ts — Added nullish coalescing (?? "Unknown error") on two error.message accesses. The error is typed with an optional .message property, so interpolating it directly could produce "undefined" in log messages.

Code Optimization
text.pollinations.ai/auth/googleCloudAuth.ts — Added a tokenRefreshScheduled boolean flag to guard against duplicate setInterval registrations. While Node.js module caching normally prevents this, the flag makes the intent explicit and prevents accidental double-registration if the initialization code is ever refactored.


fisventurous:
good work stripping out the problematic parts from the previous PR, this is already an improvement, and the error handling refactor is a good improvement as well. catching `unknown` and extracting messages safely is the right call in TypeScript.

a few things though:

1. your PR covers three separate concerns. schema input limits, error handling, and an auth guard, and since they're independent of each other, once again splitting them up would make each a lot easier to review and get merged.

2. on the schema limits specifically, some of these could quietly break existing users. capping messages at 500 risks rejecting valid long multi turn conversations, and the 100k content string limits feel erratic without first checking what the upstream providers actually accept. anything touching the API contract needs a bit of discussion before it lands.

3. also looks like this is branching off your main with 12 commits, which can drag in unrelated changes. cleaner to cut a dedicated branch off `pollinations:main` for each PR.

suggestions:
- `getErrorMessage` + the catch block fixes are clean and self-contained, would be great as its own PR
- schema limits deserve their own PR with per-field justification (upstream limits, any observed abuse, etc.)
- the `tokenRefreshScheduled` guard is a one-liner and can be its own small PR

closing this for now, but please do resubmit as a fresh PR with just the error handling refactor on a fresh feature branch based off pollinations:main, that one's the strongest piece and should be an easy merge. the rest can follow separately once there's more discussion around the specific values.

hugow0528:
@fisventurous Hello I fixed base on your suggestion 
Please check https://github.com/pollinations/pollinations/pull/9414

PR #9438, opened 2026-03-22
feat(docs): add REST-based docs search API for AI assistants

fisventurous:
simple REST docs-search API so assistants can query the docs quickly without reading everything

changes
\- adds `GET /api/docs/search` returning metadata
\- adds `POST /api/docs/search` for keyword-based doc search
\- updates `README.md` and `enter.pollinations.ai/README.md` to reflect the REST approach
\- aligns the fallback host in the search endpoint to `enter.pollinations.ai`

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,13 @@ AI assistants can:
 
 For more advanced usage, check out our [API documentation](APIDOCS.md).
 
+### Documentation Search API (for AI Assistants)
+
+We provide a lightweight docs-search API that AI assistants and tools can use to look up endpoints, models, and examples on the fly.
+
+**API info:** `GET https://enter.pollinations.ai/api/docs/search`  
+**Search:** `POST https://enter.pollinations.ai/api/docs/search` with `{"query": "..."}`
+
 ## 🔐 Authentication
 
 Get your API key at [enter.pollinations.ai](https://enter.pollinations.ai)
--- a/enter.pollinations.ai/README.md
+++ b/enter.pollinations.ai/README.md
@@ -5,6 +5,7 @@ Provides authentication, payment infrastructure and authorizing API gateway that
 
 - **[API Authentication Guide](./API_AUTHENTICATION.md)** - How to use API keys for server-to-server requests
 - **[API Documentation](https://gen.pollinations.ai/docs)** - Interactive API reference
+- **[Docs Search API](https://enter.pollinations.ai/api/docs/search)** - Search API for AI assistants to query the documentation
 - **[FAQ](./POLLEN_FAQ.md)** - Frequently asked questions about Pollen and pricing 
 
 # Local development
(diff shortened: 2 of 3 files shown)


ElliotEtag:
@fisventurous

1. MCP compatibility
- `/api/docs/mcp` is described as an MCP endpoint for clients like Claude Desktop and Cursor.
- The implementation is a custom `GET /mcp` manifest plus `POST /mcp/search`, not an MCP server speaking JSON-RPC (`initialize`, `tools/list`, `tools/call`).
- As written, standard MCP clients will not be able to connect.
- Smallest fix: rename/reframe this as a docs-search API in the route/docs, rather than MCP.

2. Search route validation
- `POST /api/docs/mcp/search` does not validate the request body.
- Malformed JSON or a non-string `query` can currently turn into a `500` instead of a `400`.
- Smallest fix: return `400` when the body is invalid or `query` is missing/non-string.

Circuit-Overtime:
@ElliotEtag I have implemented the remaining fixes we are good to merge once you verify the changes

ElliotEtag:
quick framing question before this lands — the title/body call this an MCP endpoint, but what's shipped is a REST docs-search API (`GET/POST /api/docs/search` returning MCP-tool-result-shaped JSON). worth deciding which direction we want before iterating further:

**option A — ship as docs search, drop the MCP framing**
useful endpoint as-is. what's missing:
- rename PR + update body to describe the REST contract (no MCP claims)
- README/docs framing consistent with "docs search for assistants/tools" rather than "MCP server"
- fix the host inconsistency in the no-match response (`gen.pollinations.ai/api/docs/llm.txt` vs the README using `enter.pollinations.ai/api/docs/search`)

**option B — keep the MCP goal, finish the implementation**
what's missing to actually be MCP-compatible:
- a real manifest at a stable URL (the body promises `/api/docs/mcp` with `server` / `capabilities.tools` / etc.)
- JSON-RPC envelope (`initialize`, `tools/list`, `tools/call`) — current `POST /api/docs/search` isn't speaking the protocol
- a tool definition with `name` + `inputSchema` so MCP clients can discover and call it
- a transport story (streamable HTTP is probably easiest for a hosted endpoint) — or register the tool inside the existing `packages/mcp/` server instead of building a parallel surface

both are fine outcomes — just want to make sure expectations match the shipped behaviour. which direction were you aiming for?

fisventurous:
@ElliotEtag I agree the REST approach is the cleanest option here and avoids unnecessary protocol overhead on the gateway. I've renamed the PR and updated the description to reflect the `GET/POST /api/docs/search` REST contract, removed the MCP claims, and fixed the fallback host inconsistency.

voodoohop:
Closing as stale against the current tree — the docs-search feature isn't superseded, but its target was gutted by the docs consolidation.

- This branch adds a `POST /api/docs/search` endpoint to `enter.pollinations.ai/src/routes/docs.ts`.
- #11336 (merged 2026-05-27) consolidated developer docs into gen and shrank enter's `docs.ts` to ~29 lines — enter now only emits its own schema, so the search endpoint no longer has a home there.
- No docs-search endpoint exists on main, so the feature is still genuinely missing. If it's still wanted, re-implement it in `gen.pollinations.ai` (where the docs now live) rather than rebasing this branch onto the stripped enter route.

PR #9485, opened 2026-03-23
fix: make cost estimation work as soon as data comes in

voodoohop:
## Summary
- Include today's data in model stats pipe (was excluding current day entirely)
- Add `HAVING priced_success_count >= 3` so estimates kick in after 3 datapoints
- Add `priced_success_count` column to TypeScript type


voodoohop:
polly please review

PR #9492, opened 2026-03-23
fix: spread spore-to-seed upgrade across 42 slots

voodoohop:
## Summary
- Spreads daily spore-to-seed upgrade into 6 runs/day (every 4h)
- 42 time slots instead of 7, ~750 users/run instead of ~4,500
- Eliminates GitHub API rate limiting (HTTP 403) that was hitting every run
- 8h overlapping new-user window for resilience if a run is skipped
- Tested locally: 16 batches in ~30s, quota usage negligible

## Test plan
- [x] Local dry-run: 763 users processed in ~30s, 1 approved
- [ ] Trigger workflow_dispatch dry-run on GitHub Actions
- [ ] Monitor first few automated runs for rate limit behavior


voodoohop:
polly please review

PR #9497, opened 2026-03-23
fix: default audio to true for video generation

voodoohop:
## Summary
- Changes video `audio` param default from `false` to `true`
- Video models that support audio (wan, wan-fast) will now generate audio by default


voodoohop:
polly please review

PR #9531, opened 2026-03-24
add API key rotation to developer dashboard

ElliotEtag:
- Adds `POST /api/api-keys/:id/rotate` endpoint that creates a new key with identical settings (permissions, budget, expiry, metadata), preserves the original `createdAt`, then deletes the old key
- Adds rotate button (↻) in the key list row alongside edit/delete
- Opens a confirmation modal → spinner → new key display with copy-and-close (same design pattern as key creation)
- Create-first-then-delete order avoids any window where the user has zero working keys
- If old key deletion fails, user still receives the new key with a warning to manually clean up
- KV cache invalidated for both old and new keys after rotation
- Conditional messaging: secret keys warn "you won't see it again", publishable keys don't
- Clipboard failure keeps dialog open instead of silently closing

### DIFF
--- a/enter.pollinations.ai/src/client/components/api-keys/api-key-list.tsx
+++ b/enter.pollinations.ai/src/client/components/api-keys/api-key-list.tsx
@@ -113,13 +113,13 @@ export const ApiKeyList: FC<ApiKeyManagerProps> = ({
                                             <div className="flex gap-1 shrink-0 ml-2 items-center">
                                                 <button
                                                     type="button"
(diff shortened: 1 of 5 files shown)


ElliotEtag:
polly review carefully please

ElliotEtag:
polly please review

Addressed items 2, 3, 5 from your review. Notes on the rest:

- **#1 (createdAt sort order)**: Intentional — preserving original createdAt is correct because the key was rotated, not newly created. Showing "created just now" would be misleading. Consistent position in the list is less confusing than the key jumping to the top after rotation.
- **#4 (rate limiting on rotate)**: Pre-existing concern across all mutation endpoints (create, delete, update). Not worth adding in isolation for rotate — should be addressed holistically if at all.
- **#6, #7**: Already confirmed fine in the review.

Issue #9629, opened 2026-03-26
Conflicts during the application review process


YoannDev90:
When I review apps, some automatic merges don't work because of conflicts that arise when two apps are being added at the same time.
I suggest that the workflow include a check to see if the same workflow is already running, so that it waits for the other one to finish.

YoannDev90:
eg https://github.com/pollinations/pollinations/pull/9628

PR #9630, opened 2026-03-26
Refactor concurrency group name and enable automatic merge conflict resolution

YoannDev90:
This pull request updates the `app-review-submission.yml` GitHub Actions workflow to improve concurrency handling and add automated merge conflict resolution. The most important changes are:

Workflow concurrency improvements:
* Changed the `concurrency.group` value to a static string (`app-review-submission`) instead of dynamically including issue and run identifiers, simplifying how concurrent workflow runs are grouped.

Automated merge conflict resolution:
* Added a new workflow step to detect and attempt to automatically resolve merge conflicts by merging with `origin/main` using the `theirs` strategy, committing, and pushing the result if conflicts are found.


YoannDev90:
It should solve #9629

Circuit-Overtime:
polly check if it's good give me a LGTM!

Circuit-Overtime:
polly review this and give me a LGTM

Circuit-Overtime:
@YoannDev90 I have checked the diff today by checking out the branch -- here's some of the issues I have spotted 

- if: always() — This runs even if prior steps failed (e.g., PR creation failed). It should be conditional on the PR step succeeding.

- Double merge: It runs git merge origin/main first, detects conflicts, then runs git merge --strategy-option theirs origin/main again. The second git merge won't work because you're already in a conflicted merge state. It should use git checkout --theirs . or start with the strategy option directly.

- || true everywhere: Every command is swallowed with || true, so failures are silently ignored. If the push fails, you'd never know.

- git add .  Could stage unintended files. Should be git add apps/APPS.md specifically. (still ok)

- No branch context This step doesn't check out the PR branch first. It runs after gh pr merge --auto, so the working directory state is unclear.

- --strategy-option theirs: Accepting all of main's changes defeats the purpose. The PR's own change (the new app row) would be discarded in favor of main's version, losing the submission.

This can make the flow incorrect and dangerous

YoannDev90:
The conflict-resolution step now runs only on success, checks out the PR branch, merges `origin/main` once, resolves `APPS.md` conflicts with `--ours`, stages only that file, commits, pushes, and no longer hides errors with `|| true.`

Circuit-Overtime:
@YoannDev90 I added a || true flag and this is the current situation 

| Area                          | Status     | Notes                                                                 |
|-------------------------------|------------|-----------------------------------------------------------------------|
| app-review job  | Unchanged  | No risk                                                               |
| Global concurrency group      | Safe       | Serializes submissions and prevents race conditions on APPS.md          |
| if: success()                | Correct    | Only runs after PR is successfully created                            |
| Branch checkout              | Correct    | Explicitly checks out the PR branch                                   |
| `git merge --ours`           | Correct    | Keeps the PR's app row, accepts main's other changes                  |
| git add apps/APPS.md         | Safe       | Scoped to only the relevant file                                      |
| No-conflict else push        | Harmless   | Creates an extra merge commit but no data risk                        |
| Git identity                | Inherited  | Bot config set in earlier step carries over                           |

Circuit-Overtime:
Global concurrency group, serializes app submissions to prevent APPS.md race conditions
Conflict resolution step,  with the || true fix applied, correctly merges main into PR branch and resolves conflicts by keeping the app addition (--ours)

PR #9635, opened 2026-03-27
fix: enforce audio route access parity

ElliotEtag:
- Apply the same audio access checks as other generation routes: auth, model permissions, API key budget, and estimated-cost balance gating.
- Reuse a shared generation-access helper for the standalone audio routes and the GET audio proxy route.
- Add regressions for restricted keys, exhausted key budgets, and positive-but-insufficient balance on expensive audio requests.
- Update audio happy-path tests to use a funded key where the current model estimates require it.

polly please review

### DIFF
--- a/enter.pollinations.ai/src/utils/generation-access.ts
+++ b/enter.pollinations.ai/src/utils/generation-access.ts
@@ -1,4 +1,4 @@
-import { getServiceDefinition } from "@shared/registry/registry.ts";
+import { getModelDefinition } from "@shared/registry/registry.ts";
 import { HTTPException } from "hono/http-exception";
 import type { AuthVariables } from "@/middleware/auth.ts";
 import {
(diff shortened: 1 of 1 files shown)


ElliotEtag:
polly review please

voodoohop (approved):
LGTM — fixes a real security gap where audio routes were missing `requireModelAccess()` and `requireKeyBudget()` checks that image/text routes enforce.

**What's good:**
- `requireGenerationAccess()` bundles auth + model access + key budget + balance — matches the pattern in `images.ts:requireAuthAndBalance` (lines 51-54)
- `checkBalance` properly moved to shared `generation-access.ts` instead of being inline in proxy.ts
- Dead code removal (`parseMp4Duration`) is correct — function was only referenced in a comment, never called
- Tests cover model restrictions, budget exhaustion, and insufficient balance
- Tests correctly switch to `paidApiKey` where balance checks now apply

**Minor note for follow-up:**
- `images.ts` has its own `requireAuthAndBalance` doing the same thing — could consolidate with `requireGenerationAccess` later

Resolved merge conflict (ACE-Step `generateAceStepMusic` import from main + `generation-access.ts` import from this branch).

PR #9636, opened 2026-03-27
fix: default audio transcriptions to whisper

ElliotEtag:
- Let resolveModel accept a route-specific default model.
- Set /v1/audio/transcriptions to default to whisper-large-v3, matching the docs and handler behavior.
- Add a regression that omits model and verifies the request still tracks as whisper.
- Verification: npx vitest run test/integration/audio.test.ts --testNamePattern='POST /v1/audio/transcriptions returns text'

polly please review


ElliotEtag:
polly review please

voodoohop (approved):
LGTM — clean, minimal fix.

- Transcription defaulting to `elevenlabs` (TTS model) was clearly wrong; `whisper-large-v3` is the right default
- `ResolveModelOptions` is backward-compatible — existing callers unaffected
- Test correctly validates the default-when-omitted behavior
- Operator precedence on `options?.defaultModel || (...)` is correct

Minor nit: `"whisper-large-v3"` is hardcoded rather than using a registry constant, but it's single-use and the alias resolver handles it fine. Not a blocker.

polly please review

PR #9699, opened 2026-03-28
fix: BYOP attribution - redirect_url matching and device-flow

voodoohop:
## Summary
- **Fix broken redirect_url matching**: `app-lookup` Strategy 2 (matching `redirect_url` against registered `appUrl` values) was failing in production with D1 error "LIKE pattern too complex". Replaced SQL `LIKE` with JS-side prefix matching on ~350 candidate rows.
- **Fix device-flow attribution**: Child keys created via device-flow never got `createdForApp`/`createdForUserId` metadata, even when `app_key` was provided. Removed the `!isDeviceMode` gate so both flows attribute correctly.

## Impact
- 89% of BYOP traffic (41K+ requests/month) was unattributed due to the broken Strategy 2
- All device-flow keys (153 requests from 9 keys) had no app attribution
- With this fix + pk_ keys with correct `appUrl`, new BYOP keys will automatically get attribution

## Test plan
- [ ] Verify `GET /api/app-lookup?redirect_url=https://openclaw.pollinations.ai/` returns attribution
- [ ] Verify device-flow authorization with `app_key` embeds `createdForApp` in child key metadata
- [ ] Verify existing Strategy 1 (`app_key=pk_xxx`) still works

### DIFF
--- a/apps/sirius-cybernetics-elevator-challenge/src/game/logic.ts
+++ b/apps/sirius-cybernetics-elevator-challenge/src/game/logic.ts
@@ -166,14 +166,16 @@ export const useGuideMessages = (
         }
     }, [lastMessage, addMessage, gameState.marvinJoined]);
 
-    // floor changed
+    // floor changed — intentionally only depend on currentFloor to avoid
+    // firing on every gameState change (which caused repeated "Now arriving" spam)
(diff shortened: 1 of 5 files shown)


voodoohop:
polly please review

PR #9752, opened 2026-03-30
fix: wrap pollen pack buttons on mobile

ElliotEtag:
- Fix mobile overflow in the pollen purchase grid
- Use 1 column on very narrow screens, 2 columns on typical mobile widths, and 3 columns when there is room
- Keep the existing compact 3-up layout on larger screens
- Verified with \Checked 1 file in 28ms. No fixes applied. and \


ElliotEtag:
polly please review

Issue #9307, opened 2026-03-19
[QUEST]: Automate "Trending App of the Week" social media posts


voodoohop:
## Goal

Automate a weekly "Trending App" social media post that highlights the community app with the most usage that week — with an auto-generated screenshot.

## Context

- We already have a social media pipeline (`social/`) that posts per-PR, daily, and weekly updates
- We have usage data in Tinybird that can identify top apps
- Community apps are tracked in `apps/APPS.md`

## Tasks

1. **Query Tinybird for top app by usage this week**
   - Identify which app had the most API calls / generations in the past 7 days
   - Pull app metadata (name, URL, description, author) from `APPS.md`

2. **Auto-generate a screenshot/preview image**
   - Use Playwright or similar to capture a screenshot of the app's URL
   - Optionally overlay branding / "Trending App of the Week" banner
   - Fall back to generating an image via Pollinations if the app has no web URL

3. **Post to social channels**
   - Integrate with the existing `social/` pipeline (Buffer for Twitter/Instagram, Discord webhook)
   - Post format: app name, screenshot, brief description, link, usage stats highlight
   - Schedule: weekly (e.g., Wednesday or Friday for mid-week engagement)

4. **Add to existing workflows**
   - New GitHub Action workflow or extend `NEWS_publish.yml`
   - Store post history to avoid featuring the same app repeatedly

## Acceptance Criteria

- [ ] Script queries Tinybird and identifies the top app of the week
- [ ] Screenshot/image is auto-generated
- [ ] Post is staged to Buffer / posted to Discord
- [ ] Runs on a weekly cron schedule
- [ ] Doesn't repeat the same app within 4 weeks

ChunkyPanda29:
## 🚀 Request to claim Quest #9307: Trending App of the Week

I'd like to take on this quest! Here's my implementation plan after reviewing the existing social pipeline and brand guidelines:

### ✅ What I'll Build

**1 weekly post for top app of the week as default, with 2 additional modes.**

2nd mode- Post to highlight top 3 apps of the week in a pollinations themed podium style image.

3rd mode- Posting both on 1 day a week as below for example:

| Time | Post | Format | Characters |
|------|------|--------|------------|
| **Sunday 10:00 UTC** | App of the Week (Morning) | Single winner spotlight | **App screenshot if web-based** OR **Polly presenting golden trophy + app image** |
[...]

PR #9781, opened 2026-03-31
Feat/app of the week v1

ChunkyPanda29:
Implementation of v1 scope for PR #9307


ElliotEtag:
hey @ChunkyPanda29 
Could you clean up the PR? 
Looks like we have apps/sirius-cybernetics-elevatorm changes in here.

I'll be reviewing next

ChunkyPanda29:
@ElliotEtag cleaned up, should be good to review 😊

ElliotEtag:
hey @ChunkyPanda29 !

Made an update and tested locally. app of the week only work when 'app key' is set. Which means that currently only BYOP app will work, but we'll do a code update so the App Key would allow any app usage (SK or PK) to be tracked, not only BYOP usage.

If not doing that app of the week would have become 'Dev of the week'

ChunkyPanda29:
Thanks @ElliotEtag, do you need me to do any updates or changes or are you completing this for the PK and SK usage/tracking?

voodoohop:
- Closing as stale: App of the Week depended on the old BYOP-only attribution path; current app-key, activity, and quest tracking replaced those assumptions.

PR #9816, opened 2026-04-01
[PATCH]: Fix the BYOP to localstorage in the sirius app

Circuit-Overtime:
## Changes Made:- 

- Update the code to use localstorage to store the authenticated BYOP key
- the BYOP key lives in the localstorage and serves the use of all the API related works for pollinations 

> Initial work by @ChunkyPanda29

### DIFF
--- a/apps/operation/economics/provisioning/dashboards/tier-simulation.json
+++ b/apps/operation/economics/provisioning/dashboards/tier-simulation.json
@@ -15,7 +15,7 @@
       }
     ]
   },
-  "description": "Simulate tier cost changes — adjust daily allowances per tier, and optionally simulate tier reassignment from scoring_snapshot with per-tier downgrade locks.",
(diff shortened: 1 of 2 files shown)


ElliotEtag:
Pushed two commits to fix a redirect loop bug in the original implementation:

- **Problem**: The `useEffect(() => { if (!apiKey) login(); }, [apiKey, login])` in `index.tsx` fires before the fragment extraction effect in `useBYOP()` updates state — so when a user returns from auth with `#api_key=xxx`, it redirects them back to login before the key is read, causing an infinite loop.
- **Fix**: Moved the auto-redirect into `useBYOP()`'s mount effect, right after the fragment check. Since both happen in the same synchronous block, there's no race condition.
- Removed `login` from hook's public API (now internal-only)
- Removed the separate `useEffect` + `login` import from `index.tsx`
- Kept `if (!apiKey) return null` guard to prevent rendering during redirect

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #10054, opened 2026-04-08
[Feature] NanoBanana 2: Text/font rendering broken on image endpoint — needs Chat Completions API with modalities support


wonghk-lab:
### Summary

NanoBanana 2 (Gemini 3.1 Flash Image) accessed via `image.pollinations.ai` cannot render text/fonts correctly in generated images. The same model accessed via OpenRouter's Chat Completions API (`modalities: ["image", "text"]`) renders text perfectly. This is because `image.pollinations.ai` uses a simplified image-only URL endpoint that bypasses Gemini's multimodal pipeline, which is where text rendering capability lives.

### Steps to Reproduce

1. Generate an image with text via Pollinations:
```bash
curl 'https://image.pollinations.ai/prompt/A%20poster%20with%20the%20title%20Come%20Play%20Ball%20in%20bold%20blue%20font?model=nanobanana-2&width=1024&height=1024&key=YOUR_KEY'
```

2. Generate the same image via OpenRouter (Gemini Chat API):
```bash
curl https://openrouter.ai/api/v1/chat/completions \
  -H "Authorization: Bearer $KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "google/gemini-3.1-flash-image-preview",
    "modalities": ["image", "text"],
    "messages": [{"role": "user", "content": "Generate a poster with the title Come Play Ball in bold blue font"}]
  }'
```

### Expected vs Actual

| | OpenRouter (Chat API) | Pollinations (Image URL) |
|---|---|---|
| English text | ✅ Crisp, readable | ❌ Garbled/missing |
| Chinese text | ✅ Correct rendering | ❌ Completely broken |
| Multi-line text | ✅ Proper layout | ❌ Overlapping/unreadable |

### Root Cause Analysis

NanoBanana 2's text rendering works through Gemini's **multimodal Chat Completions pipeline** — the model reasons about text placement, font selection, and layout as part of its response generation. The `image.pollinations.ai` URL endpoint strips this away by treating it as a simple text-to-image call, similar to how Flux or SDXL works. But NanoBanana 2 is fundamentally different — it's a **multimodal model that outputs images as part of a chat response**, not a standalone image generator.

OpenRouter solves this by routing NanoBanana 2 requests through `/v1/chat/completions` with `modalities: ["image", "text"]`, which preserves the full multimodal pipeline.

### Proposed Solution

[...]

PR #10196, opened 2026-04-13
feat(image): add reasoning flag for gemini image models

fisventurous:
re: #10054

the original issue (#10054) suggested moving nanobanana to the chat completions API to unlock multimodal reasoning for better text rendering, which would mean a full architectural overhaul of how we handle gemini image generation. instead of that, this adds a `reasoning` boolean query parameter to the existing image API that taps into gemini's native thinking mode directly through the vertex AI request body, same pipeline without significant structural changes

changes
\- add `reasoning` to `ImageParamsSchema` in `params.ts`, defaults to false
\- forward the flag through `vertexAIImageGenerator.ts` into the vertex request
\- in `vertexAIClient.ts`, conditionally inject `thinkingConfig` into `generation_config` when `reasoning=true`
\- gemini 3.x models get `thinkingLevel: "HIGH"`, gemini 2.5 gets `thinkingBudget: 1024`
\- non-gemini models (flux, zimage, gptimage, etc.) silently ignore the flag since they never hit the vertex code path
\- document the param in both the LLM docs (`docs.ts`) and the openapi spec
\- billing stays accurate because vertex already includes thinking tokens in `usageMetadata.candidatesTokenCount`

we could test this one out.


Circuit-Overtime:
@fisventurous we have an `enhance` parameter, can we use a enable the reasoning of COT for the models like gptimage, wan, qwen since they are really good at producing text on images using a proxy call to the text based models via pollinations.

PR #10200, opened 2026-04-13
validate model family and track gateway-rejected requests

### LINKED ISSUE #1: A text input length limit and warning on the frontend.
Users can enter a long text right now, yet these are rejected on later steps. A warning can appear when there are more than 77(?) characters entered.

ElliotEtag:
## What
Wrong-family model requests (e.g. \`gptimage\` on \`/v1/chat/completions\`, \`openai-audio\` on \`/audio/*\`) now 400 at the gateway AND land in Tinybird. Previously they either reached downstream unfiltered or were silently dropped before tracking.

## Why
Two goals, both satisfied by this PR:
1. **Don't reach downstream** — invalid (model, route-family) combos get 400'd at the gateway
2. **Don't hide malformed requests** — rejected requests still land in Tinybird so we can see which integrations are broken

Previous state created "Unregistered model" noise on the monitor. Naively adding a throw in \`resolveModel\` would have fixed #1 but broken #2 (pre-track throws skip the tracking middleware).

### Gateway (\`enter.pollinations.ai\`)
- **\`middleware/model.ts\`** — \`resolveModel\` no longer throws on family mismatch or unknown alias. It sets \`c.var.model = { requested, resolved: null, error }\` and continues.
- **\`middleware/model.ts\`** — new \`enforceModel\` middleware throws \`HTTPException(400)\` when \`resolved\` is null. Runs after \`track\`.
- **\`middleware/track.ts\`** — \`await next()\` wrapped in \`try/finally\` so \`waitUntil(sendToTinybird)\` fires even when downstream middleware throws. \`trackRequest\` handles null \`resolved\` (skips price/provider lookup).
- **\`routes/proxy.ts\`, \`routes/audio.ts\`** — middleware order is now \`validator → resolveModel → track → enforceModel → handler\` at all 8 generation endpoints.
- Type ripples (\`auth.ts\`, \`generation-access.ts\`, \`track-helpers.ts\`, \`db/schema/event.ts\`) for \`ModelName | null\`.

### Model Monitor (\`apps/model-monitor\`)
- Bundled registry is now the single source of truth; drops \`/image/models\` / \`/text/models\` / \`/audio/models\` fetches.
- Stats keyed by \`(event_type, model)\` so same-name-different-family historical rows don't collide.
- Hidden registry entries kept for matching but filtered from render unless they have traffic.
- Drops obsolete \`anomaly\` / \`catalog-unavailable\` / \`registry-only\` badges (gateway fix eliminates the anomaly).
- Gateway health panel relabeled "Unresolved Model Requests"; shows one row per endpoint family using the aggregated \`total_4xx\` the pipe actually emits.

## End-to-end flow
\`POST /v1/chat/completions\` with \`{model: "gptimage"}\`:
1. \`resolveModel\` — gptimage not in \`TEXT_SERVICES\` → \`resolved: null\`
2. \`track\` — captures \`modelRequested: "gptimage"\`, runs next() in try/finally
3. \`enforceModel\` — throws 400 "Model \"gptimage\" is not valid for this endpoint."
4. Handler — never runs
5. \`track\`'s finally — sends Tinybird row with \`response_status: 400\`, \`resolved_model_requested: null\` (ingested as \`'undefined'\` via datasource DEFAULT)
6. Monitor — row shows under "Unresolved Model Requests · generate.text"

## Validation
- \`biome check --write\` on all modified files
- Targeted tests pass: \`test/integration/text.test.ts\` (reject image models) and \`test/integration/audio.test.ts\` (reject text-chat audio on /audio and /v1/audio/transcriptions). Each asserts both the 400 response and the matching Tinybird row.
- Production build of model-monitor succeeds.

## Out of scope / follow-up
- Tinybird pipe extension for granular per-status columns (\`errors_400\`, \`errors_401\`…). The frontend already has fallback logic, so this can ship later.
- Auth/billing rejections (401/402/403) still stay on per-model rows (they happen after model resolution). The renamed panel reflects this — it's specifically for unresolved-model rejections, not all pre-2xx failures.


ElliotEtag:
polly please review

PR #10239, opened 2026-04-15
refactor: simplify daily usage pipe, dashboard key filter, and tests

voodoohop:
## Pipe

- Adds new `user_usage_daily_filtered.pipe` with optional `api_key_name` filter param
- **Deletes** `user_usage_daily.pipe` and `user_usage_daily_by_api_key.pipe` — both unused, no remaining source references
- Deploy only the new pipe; Tinybird will drop the old ones

## Server (`src/routes/account.ts`)

- `/account/usage/daily` accepts optional `api_key_name` query param, forwards it to the new pipe
- Removed `granularity` param and all `KeyedDailyUsageRecord` / `normalizeKeyedDailyUsage` logic
- `DailyUsageRecord` no longer carries an `api_key_names` array
- KV cache key now `usage:daily:{user}:{days}:{apiKeyName ?? "all"}`
- CSV export on `/account/usage/daily` preserved

## Dashboard (`src/client/...`)

- Key filter changed from multi-select to single-select (`selectedKey: string | null`)
- Dropdown options sourced from non-expired D1 keys via the existing route loader — no longer derived from Tinybird response, so keys show up even before they have any usage
- Removed the Daily Summary CSV download button, its dropdown state, `useRef`, `useEffect`, and click-outside handler — one plain "Download CSV" button remains for Detailed Usage
- `useUsageData` hook refetches when selected key changes

## SDK (`packages/sdk`)

- `DailyUsageOptions` adds `api_key_name?: string`
- `accountUsageDaily()` forwards it as a query param

## Tests

- Tinybird mock no longer reimplements pipe SQL in JS (`aggregateDailyUsage`, `isBilledUsage`, `compareEventsNewestFirst`, etc. deleted — 243 → 89 lines)
- Mock now returns canned `dailyResponse` / `usageResponse` rows and records each pipe call; tests assert both the worker's response AND the query string forwarded to Tinybird
- `account-usage.test.ts` rewritten: no 88-line event factory, no 50k-row loop, 197 → 94 lines
- Two focused tests remain: (1) daily pipe receives `api_key_name` when filter is active, (2) CSV handler passes `limit` through and renders rows with correct `Content-Disposition`

## Deploy

- Deploy the new pipe first: `tb --cloud deploy --check --wait` then without `--check`
- Deploy worker next

## Test plan

- [x] `account-usage.test.ts` passes (2/2)
- [x] `events.test.ts` passes (6/6)
- [x] `npx tsc --noEmit` clean
- [x] `npx biome check --write` clean
- [ ] Manual: dashboard key filter dropdown populates from D1 and filters chart via server roundtrip
- [ ] Manual: Detailed Usage CSV download still works


voodoohop:
polly please review

PR #10241, opened 2026-04-15
fix: normalize multi-part content for GLM/Fireworks providers

Circuit-Overtime:
## Summary
- Adds `normalizeContent` transform to text.pollinations.ai that flattens multi-part content arrays to plain strings for providers that don't support them (Fireworks, DashScope)
- GLM (via Fireworks) rejects Claude Code's `cache_control` fields and array content with 400 errors
- Keeps `glm` as the default model in polly workflows (cheaper than claude-large)

## What was happening
Claude Code sends messages like:
```json
[{"type": "text", "text": "Hello", "cache_control": {"type": "ephemeral"}}]
```
Fireworks API expects plain strings — returns `Input should be a valid string` + `Extra inputs are not permitted: cache_control`.

## Fix
New transform (`normalizeContent`) runs in the text service pipeline after `sanitizeMessages`. Detects Fireworks/DashScope via `custom-host` in model config and flattens content arrays to concatenated strings.

## Test plan
- [ ] Deploy text.pollinations.ai with the new transform
- [ ] Trigger polly on a PR and verify GLM handles the request without 400 errors


Circuit-Overtime:
@voodoohop If claude-large billing isn't an issue this PR will solve the issue of polly failing with the GLM model

PR #10252, opened 2026-04-15
refactor(image): replace boolean reasoning with fast/balanced/pro modes

fisventurous:
replaces the boolean `reasoning` toggle with a mode-based system (`fast`, `balanced`, `pro`) that maps correctly to each gemini model's supported thinkingConfig

the old `reasoning=true` was sending `thinkingBudget: 1024` to nanobanana (2.5), which actually *capped* its default 8k budget. it also sent `thinkingLevel` to gemini-3-pro-image-preview which only supports `HIGH`, causing 500s for any other value. this fixes both issues

changes
\- `params.ts`: reasoning now accepts `fast`, `balanced`, `pro` (or boolean for backward compat -- `true` -> `pro`, `false` -> `balanced`)
\- `vertexAIClient.ts`: new `getThinkingConfig` helper routes the mode to model-specific configs:
  - **3.1 flash**: `MINIMAL` or `HIGH` (thinkingLevel)
  - **3 pro**: always `HIGH` (only supported value), `fast` gracefully omits config
  - **2.5 flash**: `1024` or `24576` token budget (thinkingBudget)
\- `vertexAIImageGenerator.ts`: removed stale `as boolean` type cast

using modes instead of raw values hides the messy split between 2.5 and 3.x parameters (budget integers vs level strings), keeps the public API stable **even if Google changes CoT later**, and turns what used to be a placebo parameter into real cost vs quality control.

supersedes #10196


ElliotEtag:
I tested the PR live against Vertex.

Works:
- `nanobanana-2` / `gemini-3.1-flash-image-preview`
  - `fast -> MINIMAL` works
  - `balanced -> no config` works
  - `pro -> HIGH` works

Does not work:
- `nanobanana` / `gemini-2.5-flash-image`
  - rejects explicit `thinkingConfig` with 400
- `nanobanana-pro` / `gemini-3-pro-image-preview`
  - works with no config
  - rejects explicit `thinkingLevel: HIGH` with 400

Before merge:
- Only send `thinkingConfig` for `nanobanana-2`.
- For `nanobanana` and `nanobanana-pro`, accept the param but use provider defaults.
- If `pro` increases Google-side thinking tokens, map `thoughtsTokenCount` to existing `completionReasoningTokens` and price it. The schema/data source already supports this; no new column needed.

PR #10274, opened 2026-04-16
feat(polli): add --play flag to gen audio

voodoohop:
## Summary

- Adds `--play` flag on `polli gen audio`: plays the generated audio after saving
- Cross-platform: `afplay` on macOS, `ffplay`/`mpv`/`mpg123` on Linux (whichever is installed), PowerShell `System.Windows.Media.MediaPlayer` on Windows
- Zero new dependencies — shells out to platform players already present
- Adds Examples block to `polli gen audio --help`

## Why hand-rolled instead of an npm package

Researched `play-sound`, `sound-play`, `audic`, `node-speaker`, `symphonia`.

- `sound-play` and `play-sound` both default to PowerShell's `System.Media.SoundPlayer` on Windows — **wav-only**, silently fails on mp3 (which is what `polli gen audio` produces).
- Native-binding options (`node-speaker`, `symphonia`) require native compilation, breaks `npm install` on clean systems.
- The hand-rolled version is ~40 LOC, no deps, and actually handles mp3 on all three OSes.

## Test plan

- [x] macOS: `polli gen audio "hello" --play` plays audibly
- [ ] Linux: verify with `ffplay`, `mpv`, `mpg123` each installed individually
- [ ] Windows: verify mp3 playback (the reason we didn't use `play-sound`)
- [ ] Fallback: graceful warn when no player found

### DIFF
--- a/.claude/skills/polli-video/SKILL.md
+++ b/.claude/skills/polli-video/SKILL.md
@@ -6,7 +6,7 @@ allowed-tools: Bash(vhs *), Bash(ffmpeg *), Bash(ffprobe *), Bash(afplay *), Bas
 
 # polli-video — record polli CLI demos
 
-Hybrid pipeline: VHS records silent terminal video; polli generates audio (speech + music); ffmpeg merges. VHS cannot capture system audio — always overlay in post.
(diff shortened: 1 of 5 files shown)


voodoohop:
Superseded by scoped follow-up PR — the branch grew beyond just `--play`, so renaming into a new PR.

PR #10314, opened 2026-04-18
fix(audio): add seed param to enable deterministic caching

### LINKED ISSUE #10275: bug: audio generation not caching (elevenmusic charged repeatedly for identical prompts)
## Context

Recording a polli CLI demo. Ran `polli gen audio --model elevenmusic --instrumental` with the **identical prompt + duration** multiple times iterating on a VHS tape. Expected cache hit → ~0 pollen; got full charge each time (~0.70 pollen/request). Drained a 50-pollen top-up in ~an hour of demo iteration.

Similar concern for `polli gen audio` TTS calls with identical text (e.g. regenerating `narr-1-theme.mp3` with string "let's write the launch post for polli" across attempts).

## Repro
```bash
polli gen audio --model elevenmusic --instrumental --duration 60 \
  --output a.mp3 "lofi instrumental hip hop, warm analog, jazzy bass"
# ~25s API latency, 0.70 pollen charged

# Exact same prompt:
polli gen audio --model elevenmusic --instrumental --duration 60 \
  --output b.mp3 "lofi instrumental hip hop, warm analog, jazzy bass"
# Should hit cache (~0s, ~0 pollen). Actual: ~25s + 0.70 pollen again.
```

## Expected
Deterministic audio prompts (same model + text + duration + flags) should be cache-hit on the server, returning bytes instantly at ~0 cost.

## Impact
- Iterating on demos / TTS scripts burns pollen fast
- Discourages the "rerun freely to preview" workflow we want for the CLI

## Notes
Likely server-side (enter.pollinations.ai gateway or upstream elevenlabs-music). Worth checking if cache key includes something non-deterministic (request id, timestamp) or is just missing entirely for audio.

Ref: surfaced during polli CLI launch demo work, PR #10274.

fisventurous:
Adds optional `seed` parameter to audio generation endpoints for
best-effort determinism with ElevenLabs models (music_v1, eleven_v3),
which both accept a `seed` integer (0–4294967295) in the request body.

Without a seed, behavior is unchanged.

**GET /audio/{text}** — already cached by full URL via `audioCache`
middleware. Adding `seed` to the URL partitions cache entries, so
callers can request distinct deterministic variants instead of being
stuck with whatever the first uncached call produced. `seed=-1` still
bypasses the cache.

**POST /v1/audio/speech** — not cached today. Adds `seed` to the
request body so OpenAI-SDK callers can get reproducible output from
ElevenLabs.

Changes:
- `seed` field on `CreateSpeechRequestSchema` (POST body)
- `seed` query param on GET `/audio/:text`
- threaded through `generateSpeech()` and `generateMusic()` into the
  ElevenLabs request body
- `--seed <n>` flag on `polli gen audio`
- documented in llm.txt

Follow-up worth considering: caching POST `/v1/audio/speech` by
hashing the request body. Out of scope here.

> closes #10275


Circuit-Overtime:
@ElliotEtag Elliot please take a look at this, it's ready from my end and it allows us to cache the GET responses with even fine grained control over the seed value for the audio generation. Do let me know your thoughts

ElliotEtag (commented):
Don't merge yet — two issues:

1. **`seed=-1` escape hatch is broken.** Middleware skips cache on `-1`, but the handler still forwards `-1` to ElevenLabs, which only accepts `0..4294967295` → 400 upstream. Normalize `-1 → undefined` before calling `generateSpeech`.
2. **GET `seed` validation is too loose.** `proxy.ts:733` does `parseInt` with no bounds/NaN check, unlike the POST schema in `audio.ts:70`. Match the POST: `z.coerce.number().int().min(0).max(4294967295).optional()`.

Also: `zod` descriptions say "same cached result" but docs correctly say "best-effort" — align them. And the polli CLI flag reads "Random seed for deterministic output" — drop "Random".

Circuit-Overtime:
@ElliotEtag Thank you Elliot, I have addressed your review and also resolved the merge conflict, please do let me know further.

PR #10358, opened 2026-04-20
[codex] restack embeddings modality

ElliotEtag:
## What

Restacks embeddings onto the current `gen.pollinations.ai` layout.

## Changes

- Adds `POST /v1/embeddings` and `GET /embeddings/models` in `gen.pollinations.ai`
- Moves embeddings handling out of text-specific code into `src/embeddings`
- Moves shared Google auth helper from `src/text/auth` to `src/utils`
- Adds embeddings registry, usage headers, events, pricing, and API-key UI support
- Adds focused gen worker tests for embedding routing, billing, model listing, and media validation

## Validation

- `gen.pollinations.ai`: `npx vitest run test/embeddings.test.ts`
- `gen.pollinations.ai`: `npx vitest run test/text/utils/modelResolver.test.ts test/index.test.ts`
- `enter.pollinations.ai`: `npm run decrypt-vars`
- `enter.pollinations.ai`: `npx vitest run test/pricing-data.test.ts`

Note: the enter pricing test passed, then Vitest printed an esbuild shutdown deadlock after success while exiting with status `0`.


ElliotEtag:
- Finished the remaining embeddings hardening on this branch.
- Updated `enter.pollinations.ai/test/mocks/vcr.ts` to handle safe env host parsing and replay against canonical/current/historical service hosts.
- Updated `text.pollinations.ai/generateEmbeddings.ts` so malformed non-base64 `data:` payloads return `400` instead of leaking `500`.
- Extended the existing `enter` text mock with `/embeddings` and added 2 security cases in `enter.pollinations.ai/test/integration/embeddings.test.ts`: private/internal media URL rejection and oversized `data:` URL rejection.

- Passed: `./node_modules/.bin/vitest run test/integration/embeddings.test.ts --reporter=verbose`
- Result: `1 file passed`, `8 tests passed`.
- Passing cases: success + billing tracking, custom dimensions, batch input, wrong-modality model rejection, private media URL rejection, oversized data URL rejection, unauthenticated rejection, embeddings model list.

- Cache/VCR smoke: targeted `text-cache` run still shows broader legacy snapshot drift outside embeddings.
- Passed there: `identical POST requests produce cache hit`.
- Failed there: `different POST bodies produce cache miss` timed out after falling into record mode; `cache hit works without authentication` also fell through to missing legacy snapshots and started recording.
- Root cause looks broader than embeddings: older `openai-fast` cassettes still miss because legacy snapshot drift is not only host-based; some hashes also diverge from prior provider/modelId state.

- Cleaned up transient snapshot artifacts after the smoke runs; no stray snapshots left in the worktree.

PR #10360, opened 2026-04-20
refactor: merge account:balance scope into account:usage

voodoohop:
- Collapse `account:balance` + `account:usage` into a single `account:usage` scope
- `/account/balance` now gated by `account:usage` (key's own budget still readable with no scope)
- Consent + key-edit UI show one "Usage" row instead of two
- Silently drops legacy `"balance"` scope on new keys — no migration

Why: own-key balance and usage are both free, so the scope only gated account-wide aggregates. One scope is enough. Pre-launch — no external integrators affected.


voodoohop:
polly please review

PR #10363, opened 2026-04-20
feat(enter): 1 pollen grant on first BYOP key mint

voodoohop:
- Grants **1 pollen** to `tier_balance` when a spore user mints their first BYOP key via `/authorize` (redirect flow or device flow)
- Idempotency: new `user.byop_granted_at` column, atomic single-SQL guard (`byop_granted_at IS NULL`) — race-safe on concurrent mints
- Eligibility: `tier = 'spore'` AND key's `metadata.createdVia ∈ {redirect-auth, device-flow}` AND `metadata.clientId` present
- Refill SQL patched to preserve above-cap balances (otherwise hourly cron would strip the grant back to 0.01)
- Constant in `tier-config.ts` — set `BYOP_FIRST_KEY_GRANT_POLLEN = 0` to disable, no redeploy path beyond code
- Tests: updated refill assertions to reflect preserved above-cap balances; 41 tier tests pass

### DIFF
--- a/enter.pollinations.ai/drizzle/0018_add_byop_granted_at.sql
+++ b/enter.pollinations.ai/drizzle/0018_add_byop_granted_at.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `user` ADD `byop_granted_at` integer;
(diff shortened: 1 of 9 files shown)


voodoohop:
- Closing as rejected: current policy explicitly gives no starting tier balance (#12267), and legacy user tiers were removed (#11928).

PR #10364, opened 2026-04-20
feat(enter): 1 pollen welcome grant on every new signup

voodoohop:
- Grants **1 pollen** to `tier_balance` for every new user in `auth.ts:onAfterUserCreate`
- No BYOP gate, no eligibility marker — one-shot by definition (hook runs once per user)
- Refill SQL patched to preserve above-cap balances (otherwise the hourly cron would strip it)
- Constant `SIGNUP_GRANT_POLLEN` in `tier-config.ts` — set to 0 to disable
- Tests updated to assert above-cap preservation; 41 tier tests pass

**Parallel proposal to #10363** (BYOP-scoped variant). Pick one, close the other.

### DIFF
--- a/enter.pollinations.ai/src/auth.ts
+++ b/enter.pollinations.ai/src/auth.ts
@@ -16,7 +16,11 @@ import {
     user as userTable,
 } from "./db/schema/better-auth.ts";
 import { sendTierEventToTinybird } from "./events.ts";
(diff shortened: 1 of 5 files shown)


voodoohop:
- Closing as rejected: current policy explicitly gives no starting tier balance (#12267), and legacy user tiers were removed (#11928).

PR #10372, opened 2026-04-20
fix(mcp): stop orphan processes from busy-looping on parent exit (Windows)

Project12x:
## Summary

Two interacting bugs cause `@pollinations_ai/mcp` processes to persist after their MCP client exits on Windows, with each orphan consuming ~100% of a CPU core.

## Root cause

**1. Unconditional `startMcpServer()` at module load.** `src/index.js` ended with a bare `startMcpServer()` call (not gated by an entry-point check). Any `import` of the module — including MCP client retries that re-import — starts an additional server instance in the same process tree.

**2. No shutdown path on Windows.** The server installed `SIGINT`/`SIGTERM` handlers, but Windows does not deliver `SIGTERM` when the parent MCP client exits. Without a handler for stdin EOF, orphaned processes had no way to learn their client was gone and entered a busy-wait consuming a full CPU core each.

A third, smaller issue: `uncaughtException` was logged but not exited on, leaving the event loop running in an unknown state.

## Fix

- **Gate the auto-start** behind a proper entry-point check using `pathToFileURL(process.argv[1]).href` (handles Windows `file:///C:/...` triple-slash correctly, unlike manual string construction).
- **Add stdin EOF handlers** (`end` and `close`) that call `process.exit(0)` — the reliable shutdown signal on Windows when the MCP client disconnects.
- **Register shutdown handlers before `server.connect()`** so an early client disconnect during startup is still observed.
- **Exit on `uncaughtException`** — by definition the process is in an unknown state.
- **Keep `unhandledRejection` log-only** — a rejected tool promise shouldn't take the whole server down.

No external API or behavior change — only the startup and shutdown paths.

## Observed impact (user report)

A single workstation accumulated 11+ orphaned Pollinations MCP processes over ~2.5 days of normal Claude Code usage, collectively consuming significant continuous CPU. Other MCP packages on the same machine (Playwright, chrome-devtools, context7, etc.) held at 0% idle — confirming the issue is specific to this package's shutdown path, not a platform-wide MCP behavior.

## Test plan

Verified on Windows, Node 24:
- [x] `node src/index.js` — entry-point guard matches, single server starts
- [x] `node pollinations-mcp.js` — guard misses (correct); bin wrapper calls `startMcpServer` explicitly, no double-start
- [x] Start server via stdio, close the client — process exits (no orphan, no CPU burn)
- [x] Trigger an uncaught exception — process exits instead of hanging


Circuit-Overtime:
@Project12x I have edited the PR description a bit to make it clear to us, the claim vs the solution fits LGTM!

PR #10392, opened 2026-04-21
feat(social/instagram): trend-driven meme pipeline with slide-role schema

Itachi-1824:
Every Instagram post was the same pixel art bee — same style, same format, ~40 views and 2 likes per post. The format was never going to work on Instagram regardless of how good the content was.

## What changed

**New module: `social/scripts/generate_instagram_meme.py`**

Replaces the static platform-post generator for Instagram with a two-step pipeline:

1. `gemini-search` fetches a structured trend brief — what's actually going viral on Instagram this week, broken into six sections (viral formats, hot topics, winning visual styles, what AI brands are doing, engagement mechanics, what to avoid). This is the most important step; the quality of everything downstream depends on it.

2. A creative director prompt (`gemini-large`) receives the trend brief as its system context and the day's PR updates as the user message. It outputs a carousel plan with per-slide roles and specificity anchors.

**Slide role schema.** Each slide declares its role — `meme_hook`, `info_payload`, `meme_bridge`, `cta`, or `standalone`. Memes earn the swipe; info slides deliver the actual update as bold typography rather than a corporate bullet list. No fixed alternation; the creative director decides the mix per post based on what fits the day's content. Multi-slide posts without at least one payload slide are rejected — we never post memes without the update.

**Specificity mandate.** Every slide must declare its `specificity_anchors` — the specific trend/format/moment from the brief it's exploiting, and the specific PR detail it's anchored on. Vague prompts that could apply to any brand on any day get rejected. The point is that "AI slop" is a specificity problem, not a repetition problem, so the defence is concrete anchoring, not forced variance.

**Rendering.** Images render via `gpt-image-2` with strong text-in-image handling. The creative director decides all visual choices (palette, font personality, mood) — nothing is hardcoded in the prompt. The carousel is told to share a visual thread across slides so it reads as one post.

**`common.py`** — `generate_image()` gains a `skip_style_suffix` flag that drops the pixel art style suffix, bee character injection, and `characters-ref.jpg` reference image. All three were hardwired into every image generation call before; Instagram needs none of them.

**`generate_daily.py` / `generate_weekly.py`** — both swap in the new meme pipeline for Instagram. The old 3-image carousel cap is gone; the creative director decides the right count (1–10) based on the format.

## Why these specific decisions

- `gemini-search` for trend fetching, not `gemini-large` with search grounding — keeping research and creative direction as separate calls gives cleaner, more focused output from each
- Fail loud throughout (`sys.exit(1)`) rather than silent fallback — a post with no trend intelligence, no PR content, or no payload slide is worse than no post
- Hashtags go only in the `hashtags` array, not embedded in the caption — `build_instagram_post_text` concatenates both, so putting them in the caption too would post them twice
- `pr_summary` is now truncated (800 chars) instead of excluded when long — weekly digests are bigger and shouldn't lose their detail
- Brand copy must reflect that Pollinations now requires an API key with a generous free tier, not the old "no signup, no API key" framing


voodoohop:
- Closing as stale: this April trend-prompt experiment predates the current social pipeline, consolidated brand assets, updated models, and weekly Instagram schedule.

PR #10400, opened 2026-04-22
feat(enter): structured 5xx error signal in logs + Tinybird

voodoohop:
- keeps Workers Logs as the primary 5xx signal and removes noisy production request/response chatter
- emits one structured server_error envelope with fingerprint, route, model, and upstream context
- adds a 3-hour Tinybird error_event datasource and recent_server_errors endpoint for agent queries

PR #10418, opened 2026-04-23
fix: backport production-only fixes to main

voodoohop:
## Summary

Three fixes that were applied directly to `production` but never made it back to `main`, discovered during a branch diff audit.

- **`enter.pollinations.ai/src/error.ts`**: Check `UpstreamError` before `HTTPException` in the error handler — bundling/prototype chain issues can break `instanceof` checks, so the name-based fallback ensures upstream errors are handled correctly rather than falling through
- **`image.pollinations.ai/src/index.ts`**: Use `socket.remoteAddress` (direct connecting IP) instead of `getIp()` (forwarded headers) for security/IP logging
- **`enter.pollinations.ai/test/deduplication.test.ts`**: Fix import order + update error assertion from `"nonexistent-model-12345"` to `"Invalid option"` to match current error message format

## Test plan

- [ ] `cd enter.pollinations.ai && npm run test` passes
- [ ] No regression in error handling for upstream errors

### DIFF
--- a/enter.pollinations.ai/src/error.ts
+++ b/enter.pollinations.ai/src/error.ts
@@ -93,6 +93,16 @@ export const handleError: ErrorHandler<Env> = async (err, c) => {
     // Set the error on the context for it to be tracked
     c.set("error", err);
 
+    // Check for UpstreamError first (more specific than HTTPException)
+    // Use name check as fallback for bundling/prototype chain issues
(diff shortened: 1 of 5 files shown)


voodoohop:
Superseded by #10400, which ships a more thorough UpstreamError handler (with structured server_error emission, upstream context, and Tinybird error_event datasource) directly to main. Production has now been merged with main (2a172bdbf), so the fix is live without this backport.

PR #10419, opened 2026-04-23
fix(security): prevent credential/internal details leaking in error responses

voodoohop:
## Summary

Three fixes preventing upstream provider error context (including Google/Vertex AI diagnostic data) from reaching API clients.

- **`enter/error.ts`**: Remove `cause` field from serialized HTTP error responses — cause is logged server-side but must not be forwarded to clients (could contain internal error objects from model registry or upstream providers)
- **`text/server.ts`**: Only include upstream `details` in 4xx responses, never 5xx — Vertex AI 5xx errors can contain project IDs, service account emails, or partial JWT claims from failed token exchanges
- **`text/genericOpenAIClient.ts`**: Redact `Authorization` header before debug-logging outgoing requests to avoid access tokens appearing in log output

## Test plan

- [ ] Trigger a 5xx from a text model — confirm `details` is absent from response body
- [ ] Trigger a 4xx (bad model name) — confirm `details` still present for client errors
- [ ] Confirm `cause` field no longer appears in any error response
- [ ] Check debug logs — Authorization should show `[REDACTED]`

### DIFF
--- a/enter.pollinations.ai/src/error.ts
+++ b/enter.pollinations.ai/src/error.ts
@@ -148,7 +148,6 @@ function createErrorResponse(
             code: getErrorCode(status),
             timestamp,
             ...(details && { details }),
-            ...(!!error.cause && { cause: error.cause }),
         },
         status,
     };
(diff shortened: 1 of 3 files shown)


voodoohop:
Consolidated into #10418

PR #10436, opened 2026-04-24
draft: compare production drift against main

ElliotEtag:
Review-only comparison PR.

- Shows the current drift from `production` into `main`.
- Do not merge unless reviewers explicitly decide production-only changes should enter `main`.
- Follow-up cleanup should make `production` match `main` instead, preserving `main` as the source of truth.


voodoohop:
Closing — drift is resolved. main was merged into production at 2a172bdbf, and the production-only image IP-logging change (getIp → socket.remoteAddress) was reverted on production since all traffic is proxied through the enter Worker, making socket.remoteAddress a Cloudflare Worker egress IP rather than the end-user IP.

PR #10445, opened 2026-04-24
feat(image): add seed support for Nova Reel

fisventurous:
## Summary

Adds `seed` support to Nova Reel video generation and fills documentation gaps so `seed` is discoverable for all supporting video models.

## Changes Made

- **[image.pollinations.ai/src/models/novaReelModel.ts](image.pollinations.ai/src/models/novaReelModel.ts)** — pass `safeParams.seed` into `videoGenerationConfig` for both single-shot (`TEXT_VIDEO`) and multi-shot (`MULTI_SHOT_AUTOMATED`) task types. Nova Reel's Bedrock async invoke API accepts a numeric `seed` field on `videoGenerationConfig`; wiring it through gives users reproducible generations consistent with other video models.
- **[enter.pollinations.ai/src/routes/docs.ts](enter.pollinations.ai/src/routes/docs.ts)** — two doc fixes surfaced during review:
  - `GET /image/{prompt}` seed param description now lists `nova-reel` alongside flux, zimage, seedream, klein, seedance.
  - "🎬 Video Generation" section's parameter bullet now includes `seed` (noted as supported by veo, seedance, nova-reel)  previously only `duration`, `aspectRatio`, `audio`, `image` were listed.
  - seed=-1 till works (preprocess converts to random in [0, `MAX_RANDOM_SEED`) before validation).
seed=9999999999 or negative values other than -1,  fall back to default 42 (same pattern as other invalid inputs).
Valid [0, 2147483647] 32 bit range  pass through.

Docs are the LLM-facing `/docs` output and feed `APIDOCS.md` via generation, so both surfaces update automatically.

## Test Plan

- [ ] `curl "https://gen.pollinations.ai/video/sunset?model=nova-reel & seed=42"` twice, confirm identical output
- [ ] Vary `seed` and confirm output changes
- [ ] Multi-shot path: `duration=12` with fixed `seed`, verify reproducibility
- [ ] Hit `/docs` and confirm `nova-reel` appears in the seed description and `seed` appears in the Video parameters bullet


Circuit-Overtime:
@ElliotEtag This is an implementation adding the seed param to the AWS NOVA REEL model docs - https://docs.aws.amazon.com/nova/latest/userguide/video-gen-access.html 

> Additionally an edge case is handled here which was persistent in the past. The problem was zod's z.int() has no upper bound, so a user supplying seed=9999999999 would get 400 error

PR #10452, opened 2026-04-24
patch: code snippet  hardening from repo code check (shared, text, gen)

Circuit-Overtime:
Low-priority findings from a code-scan sweep, scoped to non-SDK services. Each claim was verified against the current source before fixing. See PR #10451 for the SDK portion.

## Changes Made

- **fix(shared): harden request token/referrer extraction** — [shared/extractFromRequest.js](shared/extractFromRequest.js)
  - Wrap `new URL()` in try/catch inside `extractReferrer` and `extractFromQuery` so a malformed request URL cannot throw out of the extractor and crash every caller that touches it.
  - In `extractFromHeaders`, re-check truthiness after the `Bearer ` prefix strip so an `Authorization: Bearer ` header with no actual token is not returned as a positive internal match.
- **fix(text): filter empty/non-string citations before rendering** — [text.pollinations.ai/server.ts](text.pollinations.ai/server.ts) — sparse upstream citation arrays and non-string entries were rendered as `"[N] undefined"`. Filter to string-typed non-empty values and skip the Sources block entirely if nothing remains.
- **fix(text): cap JSON.parse input in error paths** — [text.pollinations.ai/server.ts](text.pollinations.ai/server.ts), [text.pollinations.ai/genericOpenAIClient.ts](text.pollinations.ai/genericOpenAIClient.ts) — upstream error bodies were fed into `JSON.parse()` unbounded. Cap at 64KB in both `parseErrorDetails` and `parseJsonSafe` (truncate-to-string in the former, `null` in the latter).
- **fix(gen): normalize trailing slash in router path matching** — [gen.pollinations.ai/src/index.ts](gen.pollinations.ai/src/index.ts) — exact-match routes (`/docs`, `/models`) and prefix checks (`/account`, `/api/`) all missed when the request had a trailing slash, falling through to the catch-all `/api/generate/*` rewrite and 404'ing. Strip a single trailing slash (except for root) before matching.

Thanks to @CloudCompile  for reporting 

## Test Plan

- [x] Fire `extractReferrer`/`extractToken` with malformed URLs (`"http://["`) — returns `null` instead of throwing
- [x] `Authorization: Bearer ` (trailing space, empty token) still yields `null` from `extractToken`
- [x] Response with citations `["http://a", null, "http://b", ""]` renders only `[1] http://a` and `[2] http://b`
- [ ] Very large upstream error body (>64KB) does not OOM and is truncated/skipped
- [ ] `GET /docs/` → 301 to `/api/docs` (same as `/docs`)
- [ ] `GET /models/` → same rewrite as `/models`
- [ ] `GET /` still redirects to docs (trailing-slash guard is `> 1`)


voodoohop:
Closing — grab-bag hardening that doesn't survive review against current main.

- **gen index trailing-slash + robots.txt**: already implemented (`gen.pollinations.ai/src/index.ts:60-76,83-85` via the Hono rewrite); PR hunk is malformed and targets a structure that no longer exists.
- **`parseErrorDetails` cap**: no equivalent on main (gen uses `serializeDetails(...)`) — moot.
- **`text.pollinations.ai/server.ts` + `genericOpenAIClient.ts` hunks**: target the deleted pre-migration dir.
- **`shared/extractFromRequest.js` URL/Bearer guards**: apply cleanly but are largely speculative (thin-client/YAGNI — let upstream/platform reject).

Two bits do address real (non-speculative) upstream-data bugs — citations rendering `[N] undefined` (`gen/text/handler.ts:173-179`) and uncapped `parseJsonSafe` (`gen/text/genericOpenAIClient.ts:89-95`) — but they're low-impact and better done as a small focused change than carried in a stale grab-bag PR. Closing; can redo those two cleanly if they ever bite.

PR #10526, opened 2026-04-28
BYOP consent UI: warning chrome for unrecognized apps

### LINKED ISSUE #10478: BYOP consent UI: clearer treatment for unrecognized apps
## Context

After [#10447](https://github.com/pollinations/pollinations/pull/10447), the `/authorize` consent screen no longer infers app identity from the redirect URL — without a valid `client_id`, attribution returns `{ found: false }` and the page falls back to `attribution?.appName || redirectHostname`.

That fallback is correct but quiet. A user landing on \`/authorize?redirect_url=http://localhost:3000/cb\` sees \"localhost:3000\" in neutral chrome, indistinguishable from a registered app.

## What to build

When attribution is `{ found: false }` on the consent screen:

- **Visual:** warning chrome (amber/red border + icon), not the same neutral treatment as a recognized app.
- **Copy:** \"This app is not registered with Pollinations. We can't verify who's asking for your key. Only continue if you trust the site at \`<hostname>\`.\"
- **No fake identity:** no `@username`, no name field — just hostname.
- **Asymmetry:** registered apps get name + GitHub avatar + \"by @user\"; unregistered gets none of that plus the warning.
- **Localhost hint:** when `isLoopbackUrl(redirect_url)` (already exported from `enter.pollinations.ai/src/routes/url-utils.ts`), add a dev-mode note: \"This appears to be a local development server.\"

## Files

- `enter.pollinations.ai/src/client/routes/authorize.tsx` — consent screen
- `enter.pollinations.ai/src/routes/url-utils.ts` — `isLoopbackUrl` helper

## Why now

- The structural defect is fixed in #10447, but localhost (and any unregistered app) now silently shows a hostname. Most users won't notice the new state means \"unverified.\"
- Pairs with [DEV-FEATURE] per-key redirect allowlist (separate issue) — together they make registered/unregistered a clear binary on the consent screen.

## Out of scope

- Any backend changes (the `{ found: false }` contract is already in place)
- Per-key redirect allowlist (tracked separately)

Circuit-Overtime:
## Changes Made

- **`app-attribution.tsx`** — split into two visual treatments:
  - **Recognized** (`attribution.found === true`): existing neutral chrome — app name from `client_id` lookup, "by @githubUsername" credit, hostname footer.
  - **Unrecognized** (no `client_id`, lookup miss, or rejected redirect from PR #10524's allowlist): red alert chrome with `role="alert"`, warning icon, hostname-only display, and explicit copy: "Unverified app — \<hostname\> is not registered with Pollinations. We can't verify who's asking for your key. Only continue if you trust this site."
  - **Loopback hint** (only on the unrecognized branch): when `isLoopbackUrl(redirect_url)` is true, append "This appears to be a local development server." Helps users distinguish "local dev URL I expected" from "random unregistered app."
- **`authorize.tsx`** — forwards `redirect_url` to `AppAttribution` at both call sites (signed-out card + signed-in consent dialog) so the loopback hint can fire.

## Tasks

- [x] Add visual asymmetry between recognized and unrecognized apps so they aren't interchangeable
- [x] Drop fake identity for unrecognized apps (no `@username`, no name field — hostname only)
- [x] Add explicit "we can't verify" copy referencing the hostname
- [x] Surface dev-server hint via `isLoopbackUrl` when the redirect URL is loopback
- [x] Wire `redirect_url` through `authorize.tsx` to both call sites
- [x] Use `role="alert"` for accessibility on the warning treatment

## Notes

- No backend changes — relies on the `{ found: false }` contract already in place from #10447.
- Pairs with #10524 (per-key redirect allowlist). With both shipped, an attacker presenting a stolen `client_id` with a redirect URI not on the allowlist will hit the unrecognized branch instead of getting branded chrome.
- Reuses `isLoopbackUrl` from `src/routes/url-utils.ts` (web-standard URL/regex only — safe to import into the client bundle).

> Fixes #10478


voodoohop:
- Closing as superseded by the current app-key/OAuth authorization flow and registered redirect-URI validation.

PR #10548, opened 2026-04-29
fix: allow unlimited BYOP budget

### LINKED ISSUE #10534: BYOP authorization: budget=5 hardcoded default, UI blocks "Unlimited", string values fail JSON validation
## Bug Description
The BYOP (Bring Your Own Pollen) authorization flow has three related issues with the `budget` parameter:

### 1. Default budget is 5 (not unlimited)
When `budget` is omitted from the authorization URL, the generated API key has a hardcoded limit of **5 pollen** instead of being unlimited.

**Expected:** Omitting `budget` = no spending cap (up to user's total balance)
**Actual:** Omitting `budget` = 5 pollen limit

### 2. UI shows "Unlimited" but disables save button
In the authorization UI:
- The budget field shows "Unlimited" as placeholder text
- **BUT** leaving it empty **disables the "Save" button**
- Users cannot actually select "Unlimited" through the UI

### 3. String values fail JSON validation
Attempting to bypass via URL parameter:
- `budget=unlimited` → **JSON validation error** (backend expects number, not string)

## Reproduction Steps
1. Generate BYOP URL without budget: `https://enter.pollinations.ai/authorize?redirect_url=...&models=grok`
2. Observe: UI shows "Unlimited" placeholder, but Save button is disabled
3. OR force URL with `budget=unlimited` → JSON validation error on save
4. Final key has `budget=5` regardless of user's actual balance

## Current Workaround
Use an arbitrarily large number: `budget=999999999`

## Suggested Fix
- Backend: Treat `budget` absence as `null` (unlimited) instead of defaulting to `5`
- UI: Allow empty budget field to submit with `null`/`undefined` value
- OR: Add explicit `unlimited=true` parameter support

---
*Reported via Discord discussion*

---
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1496596607809093684/1498797878053703905)

voodoohop:
- Allows the BYOP consent screen to submit an empty budget as unlimited (`pollenBudget: null`).
- Keeps omitted URL `budget` at the existing default cap of 5.
- Documents budget behavior in BYOP docs and SDK comments.
- Adds focused coverage for explicit unlimited budget.

Tested:
- `npx biome check --write BRING_YOUR_OWN_POLLEN.md packages/sdk/src/types.ts packages/sdk/src/client.ts enter.pollinations.ai/src/client/routes/authorize.tsx shared/auth/authorize-config.ts enter.pollinations.ai/test/authorize-config.test.ts`
- `npm run decrypt-vars && npx vitest run test/authorize-config.test.ts`

Fixes #10534


voodoohop:
polly please review

### Verdict: ship — with one scope clarification

CI green. Diff is small, focused, biome-clean. The change unblocks the consent screen's "clear field for unlimited" path, which was the blocker most users hit.

### What this PR fixes (vs. issue #10534)

- ✅ Issue #2 (UI blocks "Unlimited"): `hasBudget` removed from `canAuthorize` → users can now clear the field and submit `pollenBudget: null`.
- ⚠️ Issue #1 (omitted `budget` = unlimited): **not addressed**. Omitted/empty/non-numeric `budget` still defaults to 5; users must manually clear the consent input. The PR documents this ("users can clear the budget field on the consent screen") but the issue's expected behavior was "omitting `budget` = no spending cap." Worth confirming with the reporter that the manual-clear UX is acceptable, or follow up to flip the URL default.
- ⚠️ Issue #3 (`budget=unlimited` JSON validation): not addressed. Still parsed as `null` by `parseNumber`, which on the URL path makes it equivalent to omission → defaults to 5.

### Code notes

- `shared/auth/authorize-config.ts:64` — `?? → === undefined ?` is a no-op for the only prod caller (authorize.tsx strips `null` before calling via `if (budget !== null) result.budget = budget` at authorize.tsx:109). The change makes the function correctly distinguish `null` (unlimited) vs `undefined` (default) for any future caller, and the new test pins that contract. Fine.
- The `expiry` field still uses `??` — consistent in spirit (no "unlimited expiry" concept), so leaving it asymmetric is correct.
- `client.ts:1690` doc tweak is good but could mention that `budget: null` would be the explicit "unlimited" signal — currently the SDK type is `budget?: number`, so callers can't actually request unlimited via the SDK without a string trick. Minor.

### Suggested follow-up (separate PR)

If the intent is to fully close #10534, two small changes:
1. Treat URL omission as unlimited: in `authorize.tsx:108-109`, pass `budget: search.budget === undefined ? undefined : (parsed ?? null)` (or similar) so explicit `budget=` / `budget=unlimited` flows to `null`.
2. Widen SDK `AuthorizeOptions.budget` to `number | null`.

Otherwise, this is a clean partial fix and safe to merge.

Issue #10479, opened 2026-04-26
BYOP per-key redirect allowlist (RFC 8252 port-agnostic loopback)


voodoohop:
## Context

[#10447](https://github.com/pollinations/pollinations/pull/10447) removed URL-based identity inference: app identity for `/authorize` now comes from `client_id` only. `appUrl` on a publishable key is currently opaque metadata — it's stored but not validated against the incoming `redirect_uri`.

That matches the conservative half of how Google/GitHub/Auth0 do it (no URL-based identity inference) but skips the other half (per-client redirect allowlist). Without an allowlist, a stolen `client_id` could be used with any redirect URL.

## What to build

Replace the singular `appUrl` on `pk_` key metadata with a `redirectUris: string[]` allowlist. On `/authorize`:

1. Look up the key by `client_id`.
2. Validate `redirect_uri` against `key.metadata.redirectUris`.
3. Reject mismatches with 400 \"redirect_uri not in allowlist for this client_id.\"

**RFC 8252 §7.3 port-agnostic loopback matching:** if a `redirectUris` entry has a loopback hostname (`127.0.0.1`, `::1`, `localhost`), accept any port on the incoming `redirect_uri`. Per RFC 8252 §8.3, also prefer IP literals over `localhost`.

This is the standard OAuth pattern (GitHub implements exactly this, Google for Desktop client types).

## Schema migration

- `apikey.metadata.appUrl: string` → `apikey.metadata.redirectUris: string[]`
- Migration: existing `appUrl` values populate `redirectUris[0]` (idempotent)
- Backwards compatibility for one release: read both, write only `redirectUris`

## Dashboard

- Replace single \"App URL\" field with a list editor (add/remove rows) on the publishable key edit panel.
- Show inline help: \"Loopback URLs match any port (RFC 8252).\"

## Files

- `enter.pollinations.ai/src/routes/app-lookup.ts` — add allowlist check after Strategy 1 succeeds
- `enter.pollinations.ai/src/routes/api-key-creation.ts` — accept `redirectUris[]` in metadata
- `enter.pollinations.ai/src/routes/api-keys.ts` — `UpdateMetadataSchema` accepts `redirectUris`
- `enter.pollinations.ai/src/routes/url-utils.ts` — add `redirectUriMatchesAllowlist(uri, allowlist)` with port-agnostic loopback handling
- `enter.pollinations.ai/src/client/routes/authorize.tsx` — surface a clear error when redirect rejected
[...]

PR #10549, opened 2026-04-29
security: restrict better-auth trustedOrigins to Pollinations domains

Circuit-Overtime:
## Changes Made

- Replaced the `trustedOrigins: ["*"]` wildcard in better-auth's config with an explicit allowlist of Pollinations-owned origins: `pollinations.ai`, `*.pollinations.ai` (covers `enter`, `staging.enter`, `dev.enter`, `gen`, `openclaw`, and any future subdomain via better-auth's glob support), plus `http://localhost:3000` and `http://127.0.0.1:3000` for local dev.
- Added an inline comment explaining what the field gates and why this is restricted, including a pointer to the per-key redirect-URI allowlist (#10479) that handles the BYOP `/authorize` flow separately so reviewers don't conflate the two boundaries.

## Why

`trustedOrigins` is the field better-auth uses to gate two security checks on its own `/api/auth/*` endpoints:

1. **CSRF Origin/Referer header validation**: rejects state-changing requests where the browser-attached `Origin`/`Referer` doesn't match the allowlist. With `"*"` every cross-site origin passed the check, meaning any malicious site could trigger authenticated actions (sign-out, account changes, anything mounted under `/api/auth`) against a logged-in user's session.
2. **`callbackURL` parameter allowlisting**: validates the post-login redirect target on OAuth flows. With `"*"` an attacker could craft `…/api/auth/sign-in/github?callbackURL=https://attacker.example/grab`, the user would sign in normally, then land on the attacker's domain — primed for a phishing follow-up under the visual cover of "I just logged in to Pollinations."

Both protections were silently disabled. This fix re-enables them at the lowest cost: no infrastructure change, no API surface change, just a one-config-line correction.

## Tasks

- [x] Replace wildcard with explicit Pollinations origin allowlist
- [x] Cover prod (`enter`), staging (`staging.enter`), dev (`dev.enter`), gateway (`gen`), main site (`pollinations.ai`), plus any future subdomain via glob
- [x] Keep `localhost:3000` / `127.0.0.1:3000` working for local development per AGENTS.md port conventions
- [x] Document the field's purpose and its relationship to the BYOP redirect-URI allowlist so reviewers don't break it later

## Notes

- This change does **not** affect the BYOP `/authorize` flow's own redirect handling — that's gated separately by per-key `redirectUris[]` allowlists landing in #10524. Two distinct security boundaries, intentionally kept separate.
- No test changes required: the better-auth library handles the validation; we're only adjusting its configuration. Existing auth integration tests against `enter.pollinations.ai` and `localhost:3000` continue to be in-allowlist.
- If a future deploy pattern needs a non-`pollinations.ai` origin (e.g. a custom-domain partner dashboard), it should be added explicitly here rather than reintroducing the wildcard.


voodoohop (approved):
- Closes a real open-redirect in the GitHub OAuth `callbackURL` flow (`/api/auth/sign-in/github?callbackURL=…`) — distinct from BYOP's per-key `redirectUris[]` allowlist, which gates a different flow.
- `https://*.pollinations.ai` glob correctly covers multi-level subdomains (`staging.enter`, `dev.enter`, etc.) — verified against better-auth's wildcard matcher.
- All current consumers stay in-allowlist: prod/staging/dev enter, gen, frontend `config.baseUrl`, tests on `localhost:3000`.
- Nit: PR description slightly overstates the CSRF angle — `SameSite=Lax` cookies + non-credentialed `cors({origin:"*"})` already blunt most of #1. The real win is #2 (callback redirect). Fix itself is correct either way.

LGTM.

PR #10559, opened 2026-04-29
fix(authorize): validate redirect_uri against metadata.appUrl, lock to http(s)

its3li:
Fixes the open redirect on /authorize reported in the BYOP bug bounty.

Changes:
- [safeParseUrl](cci:1://file:///f:/bug-bounty/POLLINATIONS_BUG_BOUNTY/pollinations_updated/enter.pollinations.ai/src/client/routes/authorize.tsx:57:0-70:1) now rejects non-http(s) schemes (`javascript:`, `data:`, `file:`, etc.)
- Added [isRedirectUriAllowed](cci:1://file:///f:/bug-bounty/POLLINATIONS_BUG_BOUNTY/pollinations_updated/enter.pollinations.ai/src/client/routes/authorize.tsx:72:0-92:1): requires `client_id` to be present, then does strict `origin` equality check (`scheme + host + port`) against the pk_ key's `metadata.appUrl` not startsWith/substring
- Without a `client_id` the request is rejected outright (no trusted anchor to compare against)
- [handleAuthorize](cci:1://file:///f:/bug-bounty/POLLINATIONS_BUG_BOUNTY/pollinations_updated/enter.pollinations.ai/src/client/routes/authorize.tsx:325:4-410:5): double-checks origin match before redirecting with the api_key fragment
- [handleDeny](cci:1://file:///f:/bug-bounty/POLLINATIONS_BUG_BOUNTY/pollinations_updated/enter.pollinations.ai/src/client/routes/authorize.tsx:412:4-442:5): only mirrors redirect_uri if the origin check passes; falls back to navigate home otherwise
- Surfaces a clear error message to the user when redirect_uri origin doesn't match

Related to #10550


ElliotEtag:
- Checked this against current `origin/main`.
- The core `client_id` + `redirect_uri` binding is already handled server-side by #10524:
  - `/authorize` passes `redirect_uri` to `/api/app-lookup`.
  - `/api/app-lookup` validates the redirect against the key's registered allowlist via `redirectUriMatchesAllowlist`.
  - Mismatches return `redirect_uri_mismatch` and block attribution/key delivery.
- The remaining scheme/Deny hardening is tracked separately in #10550 and should be handled there.
- Not merging this PR because it also makes `canAuthorize` require `pollenBudget !== null`, but `null` is the normal unlimited-budget state in the key permission UI.
- Closing as superseded.

PR #10561, opened 2026-04-29
feat: add Stripe Portal auto top-up

ElliotEtag:
Adds Stripe Customer Portal billing and auto top-up for paid pack balance.

- Adds billing portal setup/readiness UI and auto-top-up settings on the Pollen tab. The UI uses the shared `PollenPackSlider` and shared 5 pollen threshold constant.
- Adds D1 migration `0024_sparkling_wong.sql` with `stripe_customer_id`, `auto_top_up_*` fields, and the `stripe_auto_top_up_attempt` table.
- After billed pack usage in gen, gen calls enter via Cloudflare service-binding RPC (`ENTER_AUTO_TOP_UP.triggerForUser`) instead of using an admin bearer token. There is no cron fallback in this PR; the enter scheduled handler still only runs tier refill.
- `processAutoTopUpForUser` claims the user row, creates/finalizes/pays a Stripe invoice, records attempts, and relies on `invoice.paid`/failure webhooks for async outcomes. Idempotency is the D1 claim, claim-timestamp Stripe idempotency keys, and unique `stripe_invoice_id`.
- Hardens failure paths: GET billing stays read-only, PATCH validates JSON, disabling does not require `packAmountUsd`, non-charge exits clear `auto_top_up_claimed_at`, and failed attempts are scoped by invoice.

Deploy order: migrate prod D1 -> deploy enter -> deploy gen. Enter must deploy before gen because gen calls the new RPC entrypoint.

Out of scope: refund-induced `pack_balance` debit (#10583).


ElliotEtag:
## Pre-merge checklist

Branch up to date with `main`. Live Stripe PMC IDs filled. **Staging fully validated end-to-end.** Ready for prod rollout.

### Done — infrastructure

- [x] SOPS conflicts resolved in `gen.pollinations.ai/secrets/{prod,staging}.vars.json` — kept rotated Replicate token from #10778 + our `PLN_ENTER_TOKEN`
- [x] Live PMC IDs filled in `enter.pollinations.ai/wrangler.toml` `[env.production]`:
  - `STRIPE_BUY_POLLEN_PMC_ID = pmc_1SrY4O7rcjS3l7trnGXDTuat` (existing live Default — all 15 methods on)
  - `STRIPE_AUTO_TOP_UP_PMC_ID = pmc_1TVU4T7rcjS3l7trqtBuve71` (newly created live PMC, card + Apple Pay + Google Pay only)
- [x] Migration 0024 applied to staging D1
- [x] Enter + Gen staging deployed
- [x] Stripe sandbox webhook verified

### Done — new work landed in this PR

- `revert pollen pack bonus curve to original values` — bonus changes deferred to a separate PR
- `support rich react content in info-tip` — optional `content?: ReactNode` prop
- `surface auto top-up failures with inline warning notice` — dim-red inline notice in the auto-top-up subcontainer below "Manage billing details", `requires_action` and `failed` variants. Persists across re-enable until next successful charge. Backend exposes `lastIssue` from latest non-paid attempt.
- `add exponential backoff for auto top-up retries (1h/4h/24h then disable)` — replaces fixed 1h cooldown. Silent auto-disable after 4 consecutive failures (~29h window before disable, vs ~3h before).
- `rewrite auto top-up tooltip with bullets and paid-balance clarification` — explicit that only paid balance triggers top-up, tier balance is ignored.
- `lower auto top-up minimum to $5` — slider starts at $5
- `allow $5 in auto top-up backend validator` — fixes 400 "Invalid auto top-up pack amount" when frontend sent $5
- `void stripe invoices on auto top-up failure to keep portal clean` — webhook handlers void failed invoices
- `void auto top-up invoice in catch block when stripe pay fails` — also void inline in `processAutoTopUpForUser` catch block, since Stripe doesn't always fire `payment_failed` for SCA-blocked off-session invoices
- `update stripe tests for $5 pack and 4-strike disable threshold` — vitest snapshots updated

### Tested on staging (end-to-end)

| Scenario | Result |
|---|---|
| Buy pollen happy path (Visa 4242) | ✅ Balance credited, no double-credit on webhook retry |
| Auto top-up enable + threshold trigger | ✅ Fires within ms of crossing 5-pollen threshold via gen middleware service binding |
| Successful charge against working card | ✅ Pack credited, attempt row `paid`, `last_attempt_at` cleared |
| SCA failure (`4000 0027 6000 3184`) | ✅ Attempt `failed`, full reason stored, cooldown engaged, invoice voided immediately |
| Hard decline (`tok_chargeCustomerFail`) | ✅ Attempt `failed` with concise "Your card was declined." reason, invoice voided |
| Failure notice UI placement & copy | ✅ Dim red inline notice below "Manage billing details", correct variant per kind |
| Notice persists across disable/re-enable | ✅ Backend queries attempts table directly, not gated on `last_attempt_at` |
| Notice clears on successful charge | ✅ Latest attempt becomes `paid`, query returns null |
| Burst protection — working card | ✅ 5 concurrent → 1 charge fired, balance credited once |
| Burst protection — broken card | ✅ 5 concurrent → 1 attempt row, 1 voided invoice, no Stripe noise |
| Auto-disable after 4 consecutive failures | ✅ `auto_top_up_enabled` flips to 0 silently, no UI noise (per design) |
| Stripe portal stays clean | ✅ Outstanding count = 0 after both SCA and decline failures |
| Reverted bonus values active | ✅ $5 → 7 pollen, $10 → 15, etc. |
| 5-min Stripe idempotency window | ✅ Same charge attempt within 5 min returns same invoice (correct dedup) |

### Test suite

`enter.pollinations.ai`: **349/350 tests pass**. The 1 failure (`account-usage.test.ts > rejects periods outside supported bounds`) is unrelated date-rot — test hardcodes `2026-05-01` as a "future" period, but today is `2026-05-10`. Pre-existing, pre-PR. Worth a follow-up one-liner to compute the date dynamically.

### Production rollout (next)

- [ ] Apply migration 0024 to prod D1
- [ ] Push gen secrets to production (new `PLN_ENTER_TOKEN`)
- [ ] Deploy `enter` + `gen` to production
- [ ] Smoke test with a small ($5) real pack purchase on a maintainer account
- [ ] Enable + immediately disable auto top-up to verify the live Auto Top-Up PMC (`pmc_1TVU4T7rcjS3l7trqtBuve71`) is wired
- [ ] Watch logs for `auto_top_up_last_attempt_at` writes and webhook delivery

### Notes

- Staging is sandbox-only (`wrangler.toml` hardcodes `STRIPE_MODE=sandbox` for `[env.staging]`), so live-mode behavior can only be verified in prod
- Auto top-up trigger has two paths, both validated: request-time (gen middleware) fires within ms of crossing threshold; hourly cron sweep is the failed-charge retry backstop
- The `requires_action` UI variant is dead code in practice — Stripe routes both SCA blocks and hard declines through `payment_intent.payment_failed` for off-session charges. Harmless, can be cleaned up in a follow-up

ElliotEtag:
## Hardening pass — summary of all changes in this PR

This comment captures the full set of correctness, security, and UX hardening applied to auto top-up. Grouped by risk class.

### 1. Anti-fraud: payment evidence verification (highest-severity)

`invoice.paid` also fires for out-of-band marked-paid invoices, customer-credit-balance applications, and below-minimum-charge cases. Crediting on `invoice.paid` alone could grant pollen with no actual charge.

`creditAutoTopUpInvoice` now requires:
- `invoice.status === 'paid'`
- `invoice.amount_paid === expected_amount_cents`
- `invoice.currency === expected_currency`
- Walk `invoice.payments`, find one with `status='paid'` and `payment.type='payment_intent'`
- That invoice payment row's own `amount_paid` and `currency` must also match expectations (closes a credit-balance + tiny PaymentIntent edge case)
- Retrieve the PaymentIntent and assert `status='succeeded'`

Failure marks the attempt `failed` with `verification mismatch: <reason>` — never credits.

Both `invoice.paid` and `invoice.payment_succeeded` route through the same verifier so we don't depend on Stripe webhook subscription updates landing in lockstep with deploy.

### 2. Charge lifecycle: claimed → pending → paid with attempt-scoped idempotency

Migration 0026 added nullable `stripe_invoice_id`, plus `expected_amount_cents` and `expected_currency` columns on `stripe_auto_top_up_attempt`.

- `claimed` row written first via atomic ``INSERT ... SELECT ... WHERE EXISTS(user enabled+threshold) AND NOT EXISTS(open attempt)``.
- Stripe idempotency keys derive from the attempt UUID — no wall-clock dependency, retries map to the same Stripe operations.
- `setAutoTopUpAttemptInvoice` checks `changes === 1` and throws if the claim row vanished, so the catch path can void/delete the just-created invoice instead of proceeding with an untracked one.
- `expireStaleClaimedAttempts` deletes claims older than 5 min — self-heals worker crashes mid-`invoices.create`.

### 3. Credit-path atomicity (the subtle one)

D1 batch only rolls back when a statement *errors* — a 0-row UPDATE is a successful statement.

- Two-statement credit batch: attempt UPDATE pending|requires_action|failed → paid, then user UPDATE gated by ``EXISTS(attempt with status='paid' AND completed_at = our now)``.
- When `userChanges !== 1`: explicitly revert the attempt back to `pending` (gated by ``status='paid' AND completed_at = our now`` so we only undo our own transition), then `console.error` + throw → Stripe retries → next delivery sees pending, re-runs the credit batch, converges. No silent paid-without-credit state.
- Short-circuit when `attempt.status === 'paid'` skips the payment-evidence walk on the redundant second event (halves Stripe API spend per top-up).

### 4. Failure cleanup

- `cleanupFailedAutoTopUpInvoice` retrieves the invoice and acts by status: `draft → invoices.del()`; `open → voidInvoice()` (re-retrieve on `StripeInvalidRequestError`, treat new state); `paid|void|uncollectible → no-op`.
- Wired into `processAutoTopUpForUser` catch block, `invoice.payment_failed` webhook, and the SCA expiry path (24h-old `requires_action` attempts now also void the still-open hosted invoice).
- `invoice.voided` and `invoice.marked_uncollectible` no longer disable auto top-up — user-initiated portal voids shouldn't have hidden side effects on feature state. Controlled via `{ disableAutoTopUp?: boolean }` option (default true).
- Failure path uses `attempt.userId` from the DB row, not invoice metadata (which is mutable from the Stripe Dashboard).
- Worker-crash recovery: when `markAttemptFailedByInvoice` returns null (attempt already in `failed` from a prior crashed delivery), fall back to `SELECT user_id WHERE status='failed'` so the disable side effect still runs on retry. Gated on `status='failed'` so already-paid attempts don't accidentally disable.

### 5. Authentication

- Constant-time bearer compare on `/api/stripe/auto-top-up/trigger`. SHA-256 digest of presented + expected, byte compare via XOR loop. Reject if header is missing/malformed or if `PLN_ENTER_TOKEN` is unset/<32 chars.

### 6. Side-effect hygiene

- `GET /api/stripe/billing` no longer mutates state. Computes `effectivelyEnabled = autoTopUpEnabled && paymentMethod && billingDetailsComplete` in memory; DB never written on read paths. User can fix Stripe and the feature comes back without a re-save.
- Deleted Stripe customer: `retrieveActiveCustomer` returns `null`; `processAutoTopUpForUser` treats null as setup invalid → fail attempt, disable + clear claim, do NOT auto-create a new customer at charge time.
- Webhook `livemode` mismatch returns 200 (with log) instead of 400 — stops Stripe retry storms on permanent config mismatch.

### 7. Performance

- Gen-side threshold gate. `handleBalanceDeduction` now returns `postDeductionPackBalance` (purely additive change to `atomicDeductUserBalance`'s `RETURNING` clause). `track.ts` only fires `triggerAutoTopUp` when post-deduction pack balance is `<= AUTO_TOP_UP_THRESHOLD_POLLEN` — eliminates the per-request cross-service call on the hot generation path.

### 8. Input validation

- Backend `$2` bypass closed via shared `AUTO_TOP_UP_PACK_MIN_USD = 5` / `MAX_USD = 100` constants in `shared/billing/auto-top-up.ts`. UI imports the same constants.
- `PATCH /auto-top-up` is type-strict: `typeof body.enabled !== 'boolean'` → 400, `Number.isFinite` check on `packAmountUsd`, JSON parse wrapped in try/catch returns 400 (not 500).
- Disable path skips pack validation entirely — clients can `{ enabled: false }` without re-supplying pack data.

### Tests

- `enter.pollinations.ai/test/integration/stripe.test.ts`: **51 passing**. New coverage includes:
  - `$2` rejection, malformed JSON, disable-without-pack
  - Mode-mismatch webhook returns 200
  - Out-of-band paid invoice rejection
  - Amount/currency mismatch rejection
  - Missing succeeded PaymentIntent rejection
  - Partial PaymentIntent evidence rejection (per-payment amount/currency check)
  - Both `invoice.paid` and `invoice.payment_succeeded` for the same invoice → single credit
  - Draft invoice deleted (not voided), open invoice voided, race re-retrieved
  - SCA expiry voids the hosted invoice
  - Terminal `invoice.voided` / `invoice.marked_uncollectible` record failure without disabling
  - `payment_failed` retry disables user even when attempt already `failed` (worker-crash recovery)
  - `payment_failed` retry does not disable when attempt already `paid` (false-positive guard)
  - Deleted Stripe customer disables and does not re-create
  - Constant-time auth: missing/malformed/wrong token → 401

- `gen.pollinations.ai`: **7 passing**. Threshold gate, deduction `RETURNING` correctness for tier/pack/paid-only.

- `npm run typecheck` clean both sides. `npx biome check --write` clean.

### Out of scope (P2 follow-ups)

Tagged for a future PR:
- Retry/backoff state machine + 4-strike disable + cron retry sweeper.
- Sanitized `failure_reason` (raw Stripe error messages currently flow through to UI).
- Structured `lastIssue.actionUrl` (UI currently regex-scrapes the SCA URL out of `failure_reason`).
- SCA TTL extension from 24h → 72h (product call).
- Trigger-failure metrics to Tinybird.
- `getBillingOverview` caching (only if Stripe rate-limit / latency becomes a problem).
- Narrowing `shouldDisableAutoTopUpAfterFailure` from "any `StripeInvalidRequestError`" to specific error codes.

PR #10567, opened 2026-04-30
Normalize text cache query params

guus6457:
## Changes Made

- Sort non-auth GET query params alphabetically before hashing them into the text cache key. Previously the params were appended in caller-supplied order, so equivalent URLs with the same set of params in different orders hashed to different keys.
- Added a regression test in `gen.pollinations.ai/test/text-cache.test.ts` that issues two requests with the same params in opposite orders and asserts the second one returns `X-Cache: HIT` with `originHits === 1`.

## Why

This is a free cache-hit-rate improvement on the text endpoint. Concretely, before this change:

```
/text/foo?model=openai-fast&referrer=test  →  cache key A
/text/foo?referrer=test&model=openai-fast  →  cache key B  ← different! origin hit, no reason
```

The two URLs are semantically identical but generated different keys because `URLSearchParams` iteration preserved insertion order. SDKs, frontends, and curl one-liners that don't agree on a canonical param order all paid the cost.

After: both forms hash to the same key, and the second request is a hit.

## Tasks

- [x] Sort filtered params by key before serializing into the cache key
- [x] Regression test that covers the reordering case
- [x] Resolve the rebase against `main` — the test file was relocated from `src/` to `test/` in #10555, git's rename detection put the new test case into the right location automatically

## Notes

- **No behavior change for any single-param request** — sorting a one-element list is a no-op, and the cache key shape stays identical for any URL whose params are already in alphabetical order.
- **No effect on auth or cache-control params** — the existing `EXCLUDED_PARAMS` filter (`key`, `no-cache`) runs before the sort, so excluded params are still excluded; sort only sees the things that already mattered.
- **Validation**: `npm run test --prefix gen.pollinations.ai -- text-cache.test.ts` passes locally.


voodoohop (approved):
Reviewed locally. Focused text-cache test and Biome check pass.

PR #10580, opened 2026-04-30
fix: resolve nova-fast fallback strategy model name

voodoohop:
## Summary
- `nova-fast` uses a Portkey fallback strategy with the model nested in `targets[].override_params.model`
- `resolveModelConfig` only checked top-level `config.model` / `azure-*` / `vertex-*` fields, returning undefined
- New throw at `genericOpenAIClient.ts:145` (added in #10555) then errored every request with `"Model is required"`
- **99% failure rate, 75 distinct users, ~27k 500s/hour** since #10555 merged at 11:45 UTC
- Adds `extractFallbackModel(config)` reading `targets[0].override_params.model` as a final fallback

## Test plan
- [x] `npx vitest run test/text/utils/modelResolver.test.ts` — 5/5 pass (includes new nova-fast assertion)
- [x] `npx tsc --noEmit` clean
- [x] `npx biome check` clean
- [ ] Post-deploy: verify `curl -d '{"model":"nova-fast",...}' /v1/chat/completions` returns 200
- [ ] Post-deploy: confirm Tinybird `model_used` shows `amazon.nova-micro-v1:0` (currently `"undefined"`)

### DIFF
--- a/gen.pollinations.ai/src/text/utils/modelResolver.ts
+++ b/gen.pollinations.ai/src/text/utils/modelResolver.ts
@@ -16,6 +16,15 @@ function modelResolutionError(message: string): ServiceError {
     return error;
 }
 
+function extractFallbackModel(
+    config: Record<string, unknown>,
+): string | undefined {
+    const targets = config.targets as
+        | Array<{ override_params?: { model?: string } }>
+        | undefined;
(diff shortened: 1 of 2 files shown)


voodoohop:
Closing — wrong approach. Will reopen with a simpler fix that just removes the over-eager throw.

PR #10617, opened 2026-05-02
feat: enable image-to-image (reference images) for Grok Imagine and Grok Imagine Pro

### LINKED ISSUE #10569: Enable reference image support for Grok Imagine models
## Feature Request

Enable reference image (image-to-image) support for Grok Imagine and Grok Imagine Pro models, similar to how it already works for NanoBanana models.

## Current Behavior

When selecting NanoBanana models, a "REFERENCE IMAGES" upload section appears, allowing users to provide input images for image-to-image generation. When selecting Grok Imagine or Grok Imagine Pro, this section is hidden.

### Frontend registry (`shared/registry/image.ts`)

Update `inputModalities` for both Grok models from `["text"]` to `["text", "image"]`:

```diff
 "grok-imagine": {
     // ...
-    inputModalities: ["text"],
+    inputModalities: ["text", "image"],
     outputModalities: ["image"],
 },
 "grok-imagine-pro": {
     // ...
-    inputModalities: ["text"],
+    inputModalities: ["text", "image"],
     outputModalities: ["image"],
 },
```

This will cause `hasImageInput` (computed in `useModelList.ts`) to be `true`, which controls the reference image UI visibility in `PlayGenerator.tsx`.

### Backend handler (`image.pollinations.ai/src/models/xaiModel.ts`)

The `callXaiImageAPI` function currently only sends `prompt`, `model`, `n`, and `response_format` to xAI's API. It would need to be updated to:

1. Check if `safeParams.image` is provided
2. Forward the reference image to xAI's API in the expected format (if the xAI image editing endpoint supports it)

### Note

This depends on whether xAI's Grok Imagine API actually supports image inputs. If xAI does support image-to-image, the changes above should be straightforward. If not, this would need to wait for xAI to add that capability.

## Screenshots

**NanoBanana (reference images visible):**
The reference image upload section appears between the prompt and width/height controls.

**Grok Imagine (reference images hidden):**
The reference image section is absent — only prompt and width/height controls are shown.

Circuit-Overtime:
## Changes Made

- **Registry**: marked both `grok-imagine` and `grok-imagine-pro` as accepting image inputs by flipping `inputModalities` from `["text"]` to `["text", "image"]`. The frontend's `hasImageInput` flag — derived from this field — now resolves `true` for these models, which is the single switch that controls visibility of the REFERENCE IMAGES upload section in the Play UI. No other frontend wiring needed.
- **xAI handler**: extended `callXaiImageAPI` to detect reference images via `safeParams.image` and route to xAI's image editing endpoint when present. Concretely:
  - Split the single endpoint constant into `XAI_GENERATE_URL` (`/v1/images/generations`) and `XAI_EDITS_URL` (`/v1/images/edits`); the function now picks one based on whether reference images are supplied.
  - Added `XAI_MAX_REFERENCE_IMAGES = 5` and trim silently rather than 4xx-ing on the 6th — matches xAI's documented per-request cap.
  - Built the JSON body for edits mode with `image: [{url, type: "image_url"}, ...]` per xAI's required schema. URLs from `safeParams.image` are forwarded verbatim; xAI fetches them server-side.
  - Aspect ratio mapping, response_format, n, and the download/result-handling flow are all preserved exactly as in generate mode — edit mode is purely an additive branch with no regressions to the text-only path.
  - Progress copy switches to "Editing with Grok Imagine..." in edit mode so the UI reflects what's actually happening.
- **Documentation**: rewrote the `callXaiImageAPI` doc-comment to call out the most surprising xAI quirk — their `/v1/images/edits` requires `application/json`, **not** `multipart/form-data` like the OpenAI SDK and like our existing Azure Flux Kontext handler. A future contributor reading this file alongside `azureFluxKontextModel.ts` would otherwise reasonably assume xAI follows the same FormData pattern and break the integration.

## Why

The Play UI already has a polished reference image upload component, exposed for NanoBanana and Azure Flux Kontext, but Grok users were stuck with a text-only experience even though xAI's API supports image-to-image. Two coordinated changes (registry → UI gate, handler → wire upload through to xAI) close the gap. xAI's docs were checked directly to confirm the endpoint, body schema, and image-count cap before implementing — none of this is speculative.

## Tasks

- [x] Flip `inputModalities` for `grok-imagine` and `grok-imagine-pro` so the frontend exposes the upload UI
- [x] Branch on `safeParams.image?.length` in the xAI handler and switch endpoint URL
- [x] Build the `image: [{url, type: "image_url"}]` array for edits mode (cap at 5)
- [x] Preserve aspect ratio + response_format + download flow on edit mode (no regression on text-only)
- [x] Update the handler's doc-comment to call out the JSON-not-multipart divergence from OpenAI / our other i2i handlers
- [x] No changes needed to PlayGenerator.tsx, useModelList.ts, or any other consumer — the registry flip is sufficient because `hasImageInput` is the single source of truth

## Notes

- **Trusting caller-supplied URLs**: the handler forwards image URLs straight to xAI rather than downloading and base64-encoding server-side. xAI accepts both forms, and pass-through avoids extra latency, memory pressure on the gen worker, and a redundant fetch. If xAI ever starts rejecting URLs from non-allowlisted hosts, falling back to base64 (via the existing `downloadImageAsBase64` helper) is a small follow-up — no schema work needed.
- **Cost/price entries unchanged**. Edit mode is billed per output image just like generate mode, so the existing `cost`/`price` tables apply unchanged.
- **No migration concerns**. Old callers without `safeParams.image` continue to hit `/v1/images/generations` exactly as before — `isEditMode` is false, the new image field is omitted from the body, and the URL constant resolves to the legacy generations endpoint. Behavior is fully backwards-compatible.
- **The 5-image cap is silent** rather than an error. Reasoning: the upload UI doesn't yet enforce a per-model max (it accepts whatever the user adds), and a 4xx on the 6th image after a successful upload would feel like a worse failure mode than just dropping extras. If we want to surface the cap to users explicitly, it belongs in PlayGenerator's drop-zone copy, not here. Easy follow-up if desired.

Fixes #10569


voodoohop:
Tested the request body shape against xAI's live `/v1/images/edits` and a couple of things need a small adjustment:

- xAI's `image` field is a single `ImageUrl` struct, not a list — accepts either `{url, detail}` or the 2-element tuple `[url, detail]`. The current `[{url, type: "image_url"}]` shape returns 422.
- Edits endpoint takes one reference image, not up to 5 — so `XAI_MAX_REFERENCE_IMAGES` and the `.map()` can drop to a single `safeParams.image[0]`.

Confirmed working:
```json
"image": { "url": "https://...", "detail": "auto" }
```

…with `response_format: "url"` and both `grok-imagine-image` and `-pro` returning 200. Everything else in the PR (registry flip, edit-mode branching, response handling) looks good — just the body shape and the cap to tweak.

Circuit-Overtime:
@voodoohop thank you for the review, I have addressed the changes

voodoohop:
re-tested against live xAI endpoint, all good:

- new `{url, detail}` body shape → 200 on both `grok-imagine-image` and `grok-imagine-image-pro`
- confirmed the array form still returns 422 (matches the doc-comment claim)
- single-image use of `safeParams.image[0]` is correct, `XAI_MAX_REFERENCE_IMAGES` removal is clean
- text-only path unchanged — `requestBody.image` only set when `isEditMode`, endpoint resolves to `XAI_GENERATE_URL`

LGTM, going to enable auto-merge.

PR #10632, opened 2026-05-03
(feat): Polli CLI stable release refactor v1.0.4

Circuit-Overtime:
### `polli auth login` — branded device flow
- New ASCII flower logo derived from `assets/logo.svg`, baked at 40×20 (one-time offline rasterization via ImageMagick → PBM → `█`/space mapping). Rendered in brand purple `#a78bfa` and printed via a new TTY/JSON-gated `printBanner()` helper.
- Verification URL is now a **clickable hyperlink** (OSC-8) via the new `terminal-link` dependency. Modern terminals (iTerm, kitty, recent VS Code) make it cmd-clickable; older terminals see plain text and degrade gracefully.
- New layout: banner → "Open this URL to approve" → "Confirm the code matches" (security framing — surfacing the device-code's anti-phishing purpose instead of just calling it "Your code") → "Waiting for approval…".
- After successful login, prints a **"Try:" hint** with three example commands so new users have somewhere to go without reading `--help`.

image

### `polli auth logout` — quieter, more informative
- Captures the masked key prefix before clearing so the user sees which identity they actually logged out of (handy when juggling multiple accounts).
- Shows the credentials file path (`~/.pollinations/credentials.json`) so they know exactly what got cleared.
- Subtle checkmark + dimmed flavor line; no more shouting "CLEARED!".

### `polli whoami` (and `polli auth status`) — honest balance reporting
- **Bug fix**: the server's `/account/balance` returns the **key's remaining budget** (not the user's account pollen) when the key has a budget set, which device-flow CLI keys do by default. The CLI was mis-labeling that 5 as "pollen" while the user's actual tier balance was ~0.8.
- Now also queries `/account/key` to inspect the key, then labels honestly: `budget:` when the value is the per-key remaining budget (with an explanatory `note` field), `pollen:` when it's the real account total. JSON consumers can detect the difference programmatically via the `note` key.

### `polli whoami` top-level alias
- New top-level `polli whoami` command — alias for `polli auth status`. Reflex command in most CLI ecosystems (`gh whoami` doesn't exist but users try it; `op whoami`, `gcloud auth list`, etc.). Saves new users from digging through `polli auth --help`.

### Local-file image-to-image
- `polli image "make it sunset" --image ./photo.jpg` now works. Local paths are auto-uploaded to `media.pollinations.ai` and the returned public URL is substituted in place before the gen request fires. Mixed URLs+paths in one invocation work too — each entry is resolved independently and order is preserved (i2i providers including the recently shipped Grok-Imagine i2i path care about ordering).
- New `lib/media-upload.ts` helper with friendly framing for the most common error cases: `401` → "Run `polli auth login` or pass --key", `413` → "File too large for media.pollinations.ai (10 MB cap)". Single source of truth so `polli upload` and any future commands can reuse.
- `--image` flag's help text updated to mention local-path support; ditto README example.

### Personality refresh
- `quotes.ts` rewritten to randomize on every access via property getters. Existing call sites unchanged — they just see a different pick per invocation. Pools: 12 login quotes ("The garden is open.", "Buzz buzz. We're listening.", etc.), 11 logout ("Hibernation engaged.", "See you on the next bloom."), 10 version ("Pollen-powered.", "42 isn't the answer this time."). Hitchhiker's-Guide vibe preserved with bee/pollen flavor.

### Plumbing
- `terminal-link@^3.0.0` added to `dependencies`.
- `tsconfig.json` gains `"types": ["node"]` to fix the IDE's "Cannot find name 'process'" diagnostic on files that touch `process.stdin`/`process.stderr`.
- Version bump `0.1.3 → 0.1.4`. README adds a local-file `--image` example.

## Tasks

- [x] ASCII logo from real `assets/logo.svg` + brand-purple banner helper
- [x] Clickable verification URL via `terminal-link` (OSC-8) with graceful fallback
- [x] Login output redesign with security-framed "confirm code" copy
- [x] Post-login "Try:" hint pointing at stable commands
- [x] Logout revamp showing masked previous key and cleared file path
- [x] Fix `whoami` mis-labeling the key's budget as "pollen"; surface the distinction with a `note` field for JSON consumers
- [x] Top-level `polli whoami` alias for `polli auth status`
- [x] Auto-upload local files in `polli image --image <path>` via media.pollinations.ai
- [x] Friendly 401/413 error framing in the upload helper
- [x] Randomize login/logout/version flavor quotes
- [x] Update tsconfig types, bump version, document new behavior in README

## Notes

- **JSON / pipe / non-TTY callers see no decoration** — banner, clickable links, hint, and the new logout layout are all gated through `getOutputMode() === "human"` and `process.stderr.isTTY`. Machine consumers continue to see exactly the structured payloads they saw before.
- **No new direct dependencies in the upload path** — uses Node's built-in `FormData` + `Blob` + `fs.readFileSync`. Engines field already targets Node 20+ which has these natively.
- **Server-side balance fix (separate concern)** — the underlying server behavior where `/account/balance` returns the key's budget instead of the account total when both are set is a UX wart on the API itself, not just the CLI. Fixed honestly in the CLI by labeling the value correctly; the server-side option to mint device-flow keys without a default budget (or to treat `account:usage` as overriding the budget bias) is worth a follow-up.
- **Logo regeneration recipe** is documented in a comment at the top of `lib/branding.ts` so anyone updating the SVG can refresh the ASCII version with a one-liner.
- **Out of scope (next CLI PR)**: interactive `polli chat` REPL, `polli config get|set|unset`, per-model `polli cost`, `polli history`, shell completion, `polli topup`. Each deserves its own scope review.

### DIFF
--- a/packages/polli-cli/README.md
+++ b/packages/polli-cli/README.md
@@ -44,6 +44,7 @@ echo "context" | polli gen text "question"
 
 polli gen image "cyberpunk city at night" --model flux --output city.png
 polli gen image "enhance this" --image https://media.pollinations.ai/abc --model gptimage
+polli gen image "make it sunset" --image ./photo.jpg --model grok-imagine    # local files auto-uploaded
 
 polli gen audio "Hello world" --voice nova --output speech.mp3
 polli gen audio "read it to me" --play                # plays back after saving (blocks until done)
--- a/packages/polli-cli/README.md
+++ b/packages/polli-cli/README.md
@@ -14,7 +14,7 @@ npx @pollinations_ai/cli gen image "a cat in space" --output cat.png
 
 Point your coding agent (Claude Code, Cursor, Windsurf, Codex) at the skill file and it gets the full usage map — flags, stdin conventions, `--json` output shape, error codes, the lot:
 
-> Read https://raw.githubusercontent.com/pollinations/pollinations/main/packages/polli-cli/SKILL.md and follow the instructions to generate media with the `polli` CLI.
+> Read [SKILL.md](https://raw.githubusercontent.com/pollinations/pollinations/main/packages/polli-cli/SKILL.md) and follow the instructions to generate media with the `polli` CLI.
 
 The skill also ships inside the package: `node_modules/@pollinations_ai/cli/SKILL.md`.
 
--- a/packages/polli-cli/package-lock.json
+++ b/packages/polli-cli/package-lock.json
@@ -1,17 +1,18 @@
 {
   "name": "@pollinations_ai/cli",
-  "version": "0.1.3",
+  "version": "0.1.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@pollinations_ai/cli",
-      "version": "0.1.3",
+      "version": "0.1.4",
       "license": "MIT",
       "dependencies": {
         "chalk": "^5.4.1",
         "commander": "^13.1.0",
-        "open": "^11.0.0"
+        "open": "^11.0.0",
+        "terminal-link": "^3.0.0"
       },
       "bin": {
         "polli": "bin/polli.js"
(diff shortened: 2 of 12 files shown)


Circuit-Overtime:
polly review this please and let me know if all good

voodoohop:
- Closing as stale: the proposed CLI 1.0.4 refactor has unresolved review findings, and the CLI and docs have advanced through later releases.

Issue #10629, opened 2026-05-03
API ignores requested width/height parameters (960x640) and returns different image dimensions (1248x832) for flux / z-image models


mikewap83:
## Description

When sending a request to the Pollinations image generation API with explicit `width` and `height` parameters (e.g., `960x640`), the returned image has completely different dimensions (e.g., `1248x832`). This happens with both `flux` and `z-image` models, and also when calling the API directly from console/curl.

## Expected behavior

The returned image should strictly match the requested `width` and `height` parameters.

## Actual behavior

The API ignores the requested dimensions and returns an image with arbitrary (possibly model‑default) resolution and aspect ratio.

## Logs (from ESP32 / Arduino client)
https://gen.pollinations.ai/image/Icicles%20melt%20on%20the%20roof%20first%20drops%20fall%2C%20glitch%20art%2C%20digital%20distortion%2C%20corrupted%20data%2C%20VHS%20error%2C%20intricate%2C%20detailed%20textures%2C%20tactile%20surfaces%2C%20material%20realism?seed=40741021&negative_prompt=blurry%2C%20out%20of%20focus%2C%20ugly%2C%20disfigured%2C%20artifacts%2C%20noise%2C%20text%2C%20signature%2C%20watermark%2C%20poor%20quality%2C%20low%20resolution%2C%20missing%20limbs%2C%20extra%20limbs%2C%20mutated%2C%20distorted%2C%20bad%20anatomy%2C%20poorly%20drawn%20hands%2C%20worst%20quality&model=zimage&enhance=true&nologo=true&quality=high&width=960&height=640&safe=false

## Answer (from ESP32 / Arduino client)
the roof first drops fall, glitch art, digital distortion, corrupted data, VHS error, intricate, detailed textures, tactile surfaces, material realism","width":1248,"height":832,"seed":40741021,"model":"zimage","enhance":true,"negative_prompt":"blurry, out of focus, ugly, disfigured, artifacts, noise, text, signature, watermark, poor quality, low resolution, missing limbs, extra limbs, mutated, distorted, bad anatomy, poorly drawn hands, worst quality","nofeed":false,"safe":false,"quality":"high","image":[],"transparent":false,"reasoning":"balanced","audio":true,"has_nsfw_concept":false,"concept":[],"trackingData":{"actualModel":"zimage","usage":{"completionImageTokens":1,"totalTokenCount":1}}}

[...]

PR #10661, opened 2026-05-04
fix(image): respect user-provided dimensions with bounded scaling

sreerevanth:
## Changes Made

Previously, `gen.pollinations.ai` always pre-scaled image generation requests to ~1MP before forwarding them to the self-hosted backends, which meant explicit `width` and `height` parameters were silently ignored.

This PR adjusts the scaling behavior in `callSelfHostedServer`:

- **Respect user-supplied dimensions when the request is already ≥1MP.** Previously the path always ran through `calculateScaledDimensions`. Now we only invoke the upscaler when the request is below the 1MP target; at-or-above-1MP requests pass through untouched.
- **Clamp on the high end by the backend's actual pixel cap.** Replaces the per-axis `MAX_DIMENSION = 2048` introduced in the first revision with a total-pixel clamp matching `image.pollinations.ai/z-image/server.py`'s `MAX_FINAL_PIXELS = 768*768*4 ≈ 2.36 MP`. A per-axis cap would let 2048×2048 (4.2 MP) through to a backend that rejects it as 422; a total-pixel cap correctly handles non-square shapes too — e.g. a 2048×768 panorama (1.57 MP) is fine, while 4096×4096 is clamped down to fit the 2.36 MP budget while preserving aspect ratio.
- **Extracted `clampToPixelBudget(width, height, maxPixels)`** alongside `calculateScaledDimensions` so the high-end clamp is a small, named helper rather than an inline IIFE. Both functions now live next to each other and share the same shape (`{ scaledWidth, scaledHeight }`).
- **Tidied `data?.[0]?.b64_json`** in the GPT image response check — replaces a triple-conjunction nullish guard with optional-chaining. Equivalent semantics, smaller diff to read.
- **Removed an accidentally-committed `gen.pollinations.ai/pnpm-lock.yaml`.** This repo uses npm with a single root `package-lock.json` (workspaces); the pnpm lockfile slipped in from a local dev session and was 7367 lines of unrelated noise.

## Why this is scoped to "≥1MP requests" only

The PR's goal is **predictability for already-reasonable requests**, not "always honor exactly what the user asked for." Sub-1MP inputs still upscale via the existing `calculateScaledDimensions` heuristic — a deliberate choice so that a caller asking for `256×256` (often a default someone forgot to override) still gets a usable output. Fully respecting small explicit dimensions, which is what issue #10629 is asking about (the reporter sent `960×640` and got `1248×832` upscaled), is **out of scope here** and is being addressed in a separate PR that drops the pre-scale entirely. That PR is the right home for the discussion of "should small explicit dimensions auto-upscale?" — both because the answer affects every backend, not just z-image, and because the change shape there is delete-pre-scale, not add-clamp-logic, so coupling them in one diff would obscure the actual design call.

## Tasks

- [x] Stop forcing ≥1MP requests through the upscaler
- [x] Clamp by total pixels (matches backend's `MAX_FINAL_PIXELS`) instead of per-axis
- [x] Add `clampToPixelBudget` helper, exported alongside `calculateScaledDimensions`
- [x] Remove the accidental `pnpm-lock.yaml`
- [x] Keep the b64_json optional-chaining nicety

## Notes

- **Backend contract this PR aligns with**: `image.pollinations.ai/z-image/server.py` validates `width × height ≤ 768*768*4 = 2,359,296`, honors requested dimensions within that cap, and runs its own SPAN 2× upscaler internally for requests above `MAX_GEN_PIXELS = 768²`. The gen-side clamp here matches the validator's cap exactly (2,359,296), so anything we forward will pass the backend's own check.
- **Aspect ratio is preserved** in both branches by construction — the upscaler uses `sqrt(target/current)`, the clamp uses `sqrt(max/current)`. No risk of squashing.
- **No new dependencies, no changes to external API surface.** Same query-string parameters in, same image bytes out for any request that wasn't already broken.
- **#10629 stays open after this PR merges.** It's tracked in a follow-up PR that drops the pre-scaling step entirely so explicit small dimensions are honored. That's the change the issue's reporter is actually asking for.


Itachi-1824:
polly review please

Circuit-Overtime:
Closing this in favor of #10669, which takes the simpler shape — drop the gen-side pre-scaling entirely and let the z-image backend's own validation + SPAN upscaler handle the contract it's already designed for. That fully closes #10629 (the reporter's `960×640` case is honored exactly), whereas this PR explicitly preserved the upscale for sub-1MP requests by design.

Bundled your `b64_json` optional-chaining cleanup into #10669 with `Co-authored-by:` attribution — it's a clean win that doesn't depend on the dimensions question, and you get credit on the merge commit.

Thanks for the diagnosis and for pushing through the lockfile / per-axis-cap fixes during review — both useful and shaped the conversation that landed on the final design call.

PR #10696, opened 2026-05-05
fix(text): strip cache_control from content parts

### LINKED ISSUE #10668: kimi and kimi-k2.6 models returning PROXY ERROR 404
## Issue
Both `kimi` and `kimi-k2.6` models are returning a 404 error from the proxy layer.

### Error Message
```
PROXY ERROR 404: {"success":false,"error":{"message":"Oh no, there's nothing here.","code":"NOT_FOUND","timestamp":"2026-05-04T11:09:36.487Z"},"status":404} (unk)
```

### Expected Behavior
Models should resolve and route to Fireworks AI backend:
- `kimi` → `accounts/fireworks/models/kimi-k2p5`
- `kimi-k2.6` → `accounts/fireworks/models/kimi-k2p6`

### Registry Configuration
Both models are defined in `shared/registry/text.ts` with:
- Provider: `fireworks`
- Proper portkey configs in `text.pollinations.ai/configs/modelConfigs.ts` and `gen.pollinations.ai/src/text/configs/modelConfigs.ts`

### Reported By
Discord: `h1l0l15`

---
This appears to be a proxy resolution or backend routing issue since the model configs exist but the proxy layer returns 404.

---
**Author:** `h1l0l15` (UID: `924242134951952384`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/889573359111774329/1500817504421937284)

### LINKED ISSUE #10672: gen text gateway: normalize Anthropic-shaped requests on ingest (flatten typed-parts content, strip cache_control)
## Problem

The OpenAI-compatible shim at `gen.pollinations.ai/v1/chat/completions` forwards request bodies to upstream providers without normalizing two shape variations that some providers reject:

1. **Typed-parts `content` arrays** — when a caller sends `messages[i].content` as an array of typed parts (e.g. `[{type: "text", text: "..."}]`), some providers (notably Fireworks-hosted Kimi K2.6) reject with a 400 because they only accept `content` as a plain string for text-only turns. OpenAI itself accepts both shapes; not all upstream providers do.

2. **Anthropic-specific fields like `cache_control`** — Claude Code, the Claude Agent SDK, and any tooling that originates Anthropic-shaped requests attach `cache_control: {type: "ephemeral"}` to content blocks for prompt caching. When those requests get translated to the OpenAI shape (e.g. by [`claude-code-router`](https://github.com/musistudio/claude-code-router)) and forwarded through gen, the `cache_control` field rides along and causes strict providers to 400 with `"Extra inputs are not permitted, field: 'messages[0].content.list[1].cache_control'"`.

## How we hit it

Polly's GitHub Actions workflow ([`pr-issue-assistant.yml`](https://github.com/pollinations/pollinations/blob/main/.github/workflows/pr-issue-assistant.yml)) runs `claude-code-action` behind a local `claude-code-router` proxy that translates Anthropic→OpenAI and forwards to `gen.pollinations.ai/v1/chat/completions` with `kimi-k2.6` as the default model. Every comment-triggered Polly run started failing with the validation errors above as soon as the default model was migrated to `kimi-k2.6`. Hotfix landed in #10671 (switches Polly's default to `glm` / `kimi` K2.5, both of which happen to tolerate the format). Real fix is in the gateway.

Repro from the failing run: <https://github.com/pollinations/pollinations/actions/runs/25307594551>

## Proposed fix

Gateway-side normalization in `gen.pollinations.ai/src/text/` before forwarding upstream:

1. **Flatten text-only content arrays to a single string.** When `messages[i].content` is an array and every element is `{type: "text"}`, collapse it to the concatenated string. Preserve the array form when there's a vision/audio/tool part — those are the cases where the array shape is meaningful.
2. **Strip Anthropic-specific fields recursively.** `cache_control` is the immediate offender; there are likely a handful of others (`citations`, certain `tool_use_id` shapes) worth catching at the same time. A small allowlist of fields per OpenAI spec is more robust than a denylist.
3. **Apply uniformly to all providers**, not per-provider. The point of the shim is that callers can target any model with one shape; per-provider hacks defeat that.

## Why this matters

- **Anthropic-shaped tooling is becoming the default for agents.** Claude Code, Claude Agent SDK, OpenHands, Cline, Aider — all of these emit Anthropic-format requests, often run behind a local translator, often pointed at OpenAI-compatible gateways. If gen rejects those because one upstream provider is strict, gen stops being a useful "single front door" for that ecosystem.
- **Strictness varies by provider.** GLM-4.7 and Kimi K2.5 happen to tolerate the typed-parts format. Kimi K2.6 doesn't. As we add more providers, the surface for these mismatches grows. Normalizing once at ingest costs us one place; doing it per-call at every consumer is N places.
- **It's a real cache hit / cost issue too.** `cache_control` carries through to providers that *do* support prompt caching (e.g. Anthropic itself via Bedrock or Vertex) but for providers that don't, the field should be silently dropped, not error out the whole request. Stripping at the gateway is the right place to make that decision.

## Out of scope

- Implementing prompt caching translation (e.g. mapping Anthropic `cache_control` to provider-specific cache annotations where available). That's a separate, larger feature. This issue is just "tolerate the input shape, don't 400."
- Changes to `claude-code-router` itself. Fixing the upstream client of gen wouldn't help any of the other Anthropic-shaped clients; the gateway is the right boundary.

## Acceptance criteria

- [ ] An Anthropic-format request translated to OpenAI shape (typed-parts content + `cache_control` annotations) sent to `gen.pollinations.ai/v1/chat/completions` against any model in the registry succeeds, regardless of upstream provider strictness.
- [ ] Vision/audio/tool calls with legitimate typed-parts arrays (image_url, input_audio, tool_use, tool_result) are not flattened — only pure-text arrays collapse.
- [ ] Polly's `pr-issue-assistant.yml` can be reverted to default `kimi-k2.6` once the gateway normalization is in place. Closing this issue should mean the workflow can be flipped back.

## Related

- Hotfix PR: #10671
- Failing run: <https://github.com/pollinations/pollinations/actions/runs/25307594551>
- The `claude-code-router` config in [`pr-issue-assistant.yml:100`](https://github.com/pollinations/pollinations/blob/main/.github/workflows/pr-issue-assistant.yml#L100) is what translates Anthropic→OpenAI; the rejection happens *after* that translation, at the gen→upstream-provider hop.

voodoohop:
- Strips `cache_control` from `messages[*].content[*]` before forwarding upstream
- Unblocks Anthropic-shaped clients (Claude Code via claude-code-router) against strict OpenAI-compatible providers — verified Fireworks Kimi K2.6 returns 400 with the field, 200 without
- Narrow: only `cache_control`, only on content-part objects; string content untouched
- 3 unit tests in `messageSanitizer.test.ts`

Refs #10668, #10672. Once deployed, Polly default can be reverted to `kimi-k2.6` (#10671).

### DIFF
--- a/gen.pollinations.ai/src/text/availableModels.ts
+++ b/gen.pollinations.ai/src/text/availableModels.ts
@@ -9,6 +9,7 @@ import { createSystemPromptTransform } from "./transforms/createSystemPromptTran
 import { pipe } from "./transforms/pipe.js";
 import { removeToolsForJsonResponse } from "./transforms/removeToolsForJsonResponse.ts";
 import { sanitizeToolSchemas } from "./transforms/sanitizeToolSchemas.js";
+import { stripCacheControl } from "./transforms/stripCacheControl.js";
 import type { TransformFn } from "./types.js";
 
 interface ModelDefinition {
@@ -57,6 +58,7 @@ const models: ModelDefinition[] = [
     {
         name: "mistral",
         config: portkeyConfig["mistral-small-2503"],
+        transform: stripCacheControl,
     },
     {
         name: "deepseek",
@@ -154,10 +156,12 @@ const models: ModelDefinition[] = [
     {
         name: "kimi",
         config: portkeyConfig["accounts/fireworks/models/kimi-k2p5"],
+        transform: stripCacheControl,
     },
     {
         name: "kimi-k2.6",
         config: portkeyConfig["accounts/fireworks/models/kimi-k2p6"],
+        transform: stripCacheControl,
     },
     {
         name: "gemini-large",
@@ -180,6 +184,7 @@ const models: ModelDefinition[] = [
     {
         name: "glm",
         config: portkeyConfig["accounts/fireworks/models/glm-5p1"],
+        transform: stripCacheControl,
     },
     {
         name: "minimax",
--- a/gen.pollinations.ai/src/text/transforms/messageSanitizer.ts
+++ b/gen.pollinations.ai/src/text/transforms/messageSanitizer.ts
@@ -8,33 +8,7 @@ import type {
 const log = debug("pollinations:transforms:sanitizer");
 
 /**
- * Strips Anthropic-style `cache_control` annotations from typed content parts.
- * Strict OpenAI-compatible providers (e.g. Fireworks Kimi K2.6) reject the
- * field with a 400. Non-array content is returned untouched.
- */
-function stripCacheControlFromContent(content: ChatMessage["content"]) {
-    if (!Array.isArray(content)) return content;
-
-    let changed = false;
-    const next = content.map((part) => {
-        if (!part || typeof part !== "object" || Array.isArray(part)) {
-            return part;
-        }
-        const record = part as Record<string, unknown>;
-        if (!("cache_control" in record)) return part;
-
-        changed = true;
-        const { cache_control: _drop, ...rest } = record;
-        return rest;
-    });
-
-    return changed ? next : content;
-}
-
-/**
- * Transform that sanitizes messages: replaces empty user content with a
- * placeholder, and strips Anthropic-only `cache_control` annotations from
- * typed content parts so strict upstream providers don't 400.
+ * Transform that sanitizes messages by replacing empty user content with placeholder text.
  */
 export function sanitizeMessages(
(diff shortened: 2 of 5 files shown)


voodoohop:
polly please review

PR #10727, opened 2026-05-06
feat: developer earnings dashboard

ElliotEtag:
Adds a new **Earnings** tab in the enter.pollinations.ai dashboard for devs running BYOP apps with markup enabled.

- Three stat cards: **Pollen earned** (with avg markup tag), **Active users** (with "across N apps"), **Top app** (with pollen + distinct user tags)
- Line chart of daily earnings, period picker, per-app filter
- `PeriodPicker` and `MultiSelect` now accept a `theme` prop so the controls can render yellow inside Earnings while staying pink in Usage; `Chart` accepts an optional `paidBarColor` override for the same reason
- Backend: `GET /account/earnings/daily` (per-day per-app for the chart) and `GET /account/earnings/summary` (period-rolled-up for stats), both KV-cached for 1h
- Two new Tinybird pipes: `developer_earnings_daily.pipe`, `developer_earnings_summary.pipe`. Earnings = `total_price - dev_price` filtered to events with `markup_rate > 0` through the dev's `pk_` keys

**Heads-up:** depends on `dev_price` / `markup_rate` columns shipping in `generation_event` from `feat/creator-balance-bucket`. Until that merges, the pipes won't deploy and the section will gracefully show zeros + the chart's "Try again" error overlay. Stat cards always render so the layout is visible.

Sidebar order also updated to: News · Models · Keys · Pollen · Earnings · Usage.


ElliotEtag:
**REMOVE BEFORE MERGE — design mock**

I added a dev-only earnings mock so the wallet + earnings dashboard could be designed against varied data. It is gated on `import.meta.env.DEV` and a `?mockEarnings=<scenario>` URL param, so it cannot reach prod, but the call sites and the file should be stripped before merging.

Files / call sites to remove:
- `enter.pollinations.ai/src/client/lib/mock-earnings.ts` — delete entire file
- `enter.pollinations.ai/src/client/routes/index.tsx` — drop the `getMockEarningsScenario` import + the `mockToday` block in the loader
- `enter.pollinations.ai/src/client/components/usage-analytics/use-earnings-data.ts` — drop the `getMockEarningsScenario`/`getMockEarningsResponse` import + the mock short-circuit in `fetchEarnings`

Each call site is tagged with a `REMOVE BEFORE MERGE` comment — `grep "REMOVE BEFORE MERGE"` from the repo root surfaces all three.

PR #10769, opened 2026-05-08
feat(authorize): ignore query string in redirect_uri match

voodoohop:
## Summary
- Match `redirect_uri` on scheme + host + port + path only; ignore query string.
- Apps round-trip state in query params (e.g. CatGPT's `?prompt=&model=`); strict query equality forced re-registration for every variant.
- Pinning host + path is what blocks confused-deputy attacks — the query adds no security. Mirrors GitHub's behavior; OAuth 2.1/RFC 8252 §8.10 explicitly permits this.
- Fragments still rejected on both sides.

## Test plan
- [x] `npx vitest run test/url-utils.test.ts` (11/11 pass)
- [ ] Verify CatGPT authorize flow with `?prompt=...` query params now succeeds end-to-end.


voodoohop:
Superseded by combined PR

PR #10787, opened 2026-05-08
Updated chat.pollinations UX and UI

CloudCompile:
## Changes Made

- Fix tutorial modal flashing on first open (gated step-change effect with a ref so init no longer triggers the fade-out/in transition).
- Sanitize API errors on the client: 401/402/403/429/5xx are mapped to friendly messages and tagged with `errorType` — raw server messages are no longer surfaced.
- Add BYOP (Bring Your Own Pollen key) flow: in-message button on auth/balance errors opens a modal where users can paste an `sk_…` key (stored in `localStorage`); all API calls prefer it over the env key.
- Constrain Keyboard Shortcuts modal so it always fits on small viewports.
- Fix model dropdown getting clipped by `.input-card` (`overflow: hidden` → `visible`).
- Add top padding to the messages container so the conversation no longer touches the header.
- Move all component CSS into `apps/chat/src/components/styles/` and update imports.
- Lint cleanup: template literal in `markdown.js`, drop unused `supportsVision`, drop assign-in-while in chart-regex loop, replace overlay `<div onClick>` with a button for a11y.

## Checklist

- [x] Tutorial no longer flashes on first open
- [x] 401 surfaces a "Use your own API key (BYOP)" button instead of raw error text
- [x] BYOP key persists across reloads and is used by chat/image/video/audio
- [x] Keyboard Shortcuts modal fits within viewport
- [x] Model dropdown opens above the input without being clipped
- [x] All `apps/chat/src/components/*.css` moved to `components/styles/` with imports updated
- [x] Biome diagnostics from the CI run resolved


Circuit-Overtime (approved):
@CloudCompile Approving all the changes for the merge LGTM!

PR #10811, opened 2026-05-09
fix(gen): surface upstream error codes instead of generic 500s

fisventurous:
stops video model handlers from masking upstream errors as generic 500s

the error path in several video models was hardcoding 500 for any upstream failure, making it impossible to distinguish a content policy rejection (400) from a rate limit (429) or an actual server crash. this surfaces the real status codes downstream.

\- seedance v1: extract BytePlus `error.code` and use it as HTTP status when it's a valid 4xx/5xx
\- seedance v2: richer Replicate error logging - logs full error object and structured `{ message, status }` before rethrowing
\- wan: add `getDashScopeErrorStatus()` classifier that maps error messages to 400 (content filter, validation) or 429 (rate limit) instead of blanket 500
\- xai video: poll failures now return 502 (bad gateway) since it's an upstream provider failure and not a pollinations server error

### DIFF
--- a/gen.pollinations.ai/src/image/models/seedanceV2VideoModel.ts
+++ b/gen.pollinations.ai/src/image/models/seedanceV2VideoModel.ts
@@ -24,11 +24,38 @@ const logError = debug("pollinations:seedance2:error");
 const MODEL = "bytedance/seedance-2.0";
 const TRACKING_LABEL = "seedance-2.0";
 
+// Replicate's Seedance 2.0 accepts a narrower set than our shared aspectRatio
+// enum (which also allows 9:21). Validate at the boundary so users get a clear
+// 400 instead of a Replicate 422 round-trip.
+const SEEDANCE_V2_ASPECT_RATIOS = [
+    "16:9",
+    "4:3",
+    "1:1",
+    "3:4",
+    "9:16",
+    "21:9",
+    "adaptive",
+] as const;
+type SeedanceV2AspectRatio = (typeof SEEDANCE_V2_ASPECT_RATIOS)[number];
+
+export function resolveSeedanceV2AspectRatio(
+    requested: ImageParams["aspectRatio"] | undefined,
+): SeedanceV2AspectRatio {
+    if (!requested) return "16:9";
+    if ((SEEDANCE_V2_ASPECT_RATIOS as readonly string[]).includes(requested)) {
+        return requested as SeedanceV2AspectRatio;
+    }
+    throw new HttpError(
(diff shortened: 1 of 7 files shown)


ElliotEtag:
added 3 follow-up commits, polly please review:

- pass through Replicate HTTP error statuses (429 → 429, other upstream 4xx/5xx → 502) — fixes the 500 fallback in `replicateFetch`
- map BytePlus content-filter ("output may contain sensitive information") and image-fetch errors to 400/422; validated against 30 days of prod error strings
- 15 unit tests covering all three classifiers (replicate http, replicate prediction, byteplus)

ElliotEtag:
addressed polly's findings — 3 more commits:

**fixed in this PR:**
- **P1 log leak** `937e7c2`: dropped duplicate unredacted request body log in seedance v1 (the redacted variant at lines 215-230 already handles it)
- **P2 broken test + P2 over-eager 502 mapping** `f0da421`: `classifyReplicateHttpStatus` now passes through 400/422 (Replicate input validation) so users see actionable 4xx instead of opaque 502; updated existing replicateClient test to match
- **P1 unbounded poll + missing aborted** `11bfad5`: added 5-min poll budget (60×5s) → 504 with prediction id, and 'aborted' joins failed/canceled as terminal status. two new tests cover both paths.

**deferred to separate PR:**
- **P2 aspectRatio mismatch** at seedanceV2VideoModel.ts:79: our schema allows 9:21 but Replicate Seedance 2.0 only accepts 16:9/4:3/1:1/3:4/9:16/21:9/adaptive. legitimate bug but it's a feature/data fix (model-specific input validation), not error handling. happy to open a follow-up issue.

all 28 tests pass, biome + tsc clean. polly please re-review.

ElliotEtag:
you were right, fixed it: `d06ffc7` validates aspectRatio against Replicate's supported set (16:9, 4:3, 1:1, 3:4, 9:16, 21:9, adaptive) at the gateway. 9:21 (allowed by our shared enum but not by Replicate) now returns 400 with a clear list of supported values instead of round-tripping to a 422.

polly please re-review.

fisventurous:
nice, the follow ups are exactly what was needed. the regex classifiers validated against prod logs are way more robust than what I had inline, and the poll budget + aborted handling in the replicate client were blind spots I missed entirely. test coverage is looks good. good edge cases around the byteplus code scheme. the aspect ratio guard on seedance v2 is a nice bonus as this should save a wasted run to replicate. 

everything checks out on my end.

PR #10842, opened 2026-05-10
add support read-only route + speed up tinybird sync to 15min

ElliotEtag:
- new `/api/support/*` routes for live D1 reads to support user accounts/balances (user lookup by id/email/github, apikeys, sessions, accounts, payments, quest payouts, stats)
- bearer auth via new `SUPPORT_READ_TOKEN` secret (per-env), separate from admin token
- sensitive columns never returned: `account.access_token`/`refresh_token`/`id_token`/`password`, `apikey.key`, `session.token`, `verification.value`
- predefined queries only — no raw SQL, no writes, easy to revoke
- d1→tinybird sync cron: `0 * * * *` → `*/15 * * * *` on prod + staging (still fresh enough without compaction risk)

deploy steps:
1. `npm run push-secrets:staging && npm run deploy:staging`
2. test with curl
3. `npm run push-secrets:production && npm run deploy:production`


ElliotEtag:
superseded by the new PR — we decided to skip live D1 access (no backup) and just speed up the tinybird mirror instead

PR #10909, opened 2026-05-13
feat: add generic Replicate passthrough

voodoohop:
- Adds `/v1/replicate/predictions` for time-priced Replicate models.
- Bills runtime through `billingDollars` and hidden `replicate-generic`.
- Rejects tiered Replicate pricing; failed predictions pass through without billing.
- Validated focused tests, typechecks, routing test, and diff check.

### DIFF
--- a/enter.pollinations.ai/observability/datasources/generation_event.datasource
+++ b/enter.pollinations.ai/observability/datasources/generation_event.datasource
@@ -62,6 +62,7 @@ SCHEMA >
 `token_price_completion_image`     Float64  `json:$.tokenPriceCompletionImage` DEFAULT 0,
 `token_price_completion_video_seconds` Float64 `json:$.tokenPriceCompletionVideoSeconds` DEFAULT 0,
 `token_price_completion_video_tokens`  Float64 `json:$.tokenPriceCompletionVideoTokens` DEFAULT 0,
+`token_price_billing_dollars`      Float64  `json:$.tokenPriceBillingDollars` DEFAULT 0,
 
 -- Usage
 `token_count_prompt_text`          UInt32   `json:$.tokenCountPromptText` DEFAULT 0,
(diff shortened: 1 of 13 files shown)


voodoohop:
[claude] reviewed — two follow-ups still open:

**1. non-succeeded → return raw replicate JSON + actual upstream status (currently force-overrides to 502 + returns 200 to client)**

`routes/replicate.ts:162-167` overrides tracking to 502 but the response body still goes out with hono's default 200. The client sees a 200 with a `status: "failed"|"canceled"|"aborted"` body — confusing. Either:
- propagate replicate's terminal status as an actual HTTP code (failed/canceled/aborted → 502, content-filter via `prediction.error` matches like image client's `classifyReplicatePredictionError` → 400), or
- at minimum align the client-visible status code with the tracking override.

**2. succeeded-but-no-predict_time → log + bill \$0 + return 200 (currently logs + overrides to 502 + returns prediction)**

`routes/replicate.ts:174-185` — log warning is in, good. But still overrides to 502 instead of letting the success through unbilled. If replicate gives us a `succeeded` we should return it. Worst case we ate the cost; that's a metrics/alerting concern, not a 4xx/5xx to the user.

**3. smoke test status?**

SDXL 1-step live call against the deployed branch — did `REPLICATE_API_TOKEN` get exported and a real prediction round-trip end-to-end? Want to see the response headers (`x-usage-billing-dollars`, `x-replicate-*`) on a real run before merge.

want me to push fixes for 1+2 as a follow-up commit on this branch?

ElliotEtag:
- Closing this stale, conflicting PR in favor of the current Replicate lifecycle work in #12656.
- The unrestricted Pollen-billed passthrough is not being carried forward: Replicate does not expose an authoritative downstream-inclusive charged amount, and scraped/cached pricing can undercharge when hardware, pricing, or nested model usage changes.
- A safe generic replacement would need BYOK or an audited allowlist with explicit pricing.

PR #10966, opened 2026-05-15
fix(byop): revoke child keys when app key is deleted

ElliotEtag:
Cascade BYOP `sk` child keys when their parent `pk` app key is deleted.

- `/api/account/keys/:id`: delete children before parent so a partial failure never leaves active BYOP keys pointing at a deleted app.
- Better Auth `/api-key/delete`: new plugin captures the deleted key in a before-hook and cleans up children on success.
- `user.delete.before` cascade: drops BYOP children owned by other users that reference this user's app keys (FK cascade only handles same-user rows).
- Dashboard delete now goes through the account route so the cleanup always runs.
- Adds integration tests covering both delete paths.

### DIFF
--- a/enter.pollinations.ai/src/auth.ts
+++ b/enter.pollinations.ai/src/auth.ts
@@ -1,4 +1,8 @@
 import { createApiKeyPlugin } from "@shared/auth/api-key.ts";
+import {
+    deleteByopChildKeysForAppKey,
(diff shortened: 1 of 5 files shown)


voodoohop:
Id close for now. until we have evolved our thoughts a bit more how / if byop keys should be tied to specific apps.

PR #11001, opened 2026-05-16
fix(image): surface real error on upstream 400 instead of "{}"

voodoohop:
## Summary
- `/v1/images/edits` (and `/image/*`) was returning `upstreamBody: "{}"` and a generic `"Image provider error"` whenever an `HttpError` was thrown without a `details.body` (e.g. `nanobanana-2` failing in `vertexAIImageGenerator.buildNoImageDataError`).
- Now the real `HttpError.message` is preserved on both `responseBody` and the user-facing `message`, so clients see something like `"Image provider error: Gemini: <reason>"` instead of an empty `{}`.
- Existing tracked-400 flow (provider body with `message`/`detail`) is unchanged — covered by the existing tests.

## Why
While debugging a `nanobanana-2` image-edit failure, the only signal from the gateway was `upstreamBody: "{}"`. That's not actionable — the underlying Gemini text/finish-reason was already being attached to the `HttpError` but discarded by `throwImageError` and `classifyImageHttpError`.

## Changes
- `gen.pollinations.ai/src/image/handler.ts`
  - `throwImageError` now uses `imageResponseBody(error)` which prefers `details.body`, then `error.message`, then a JSON details fallback — never raw `"{}"` when a message exists.
  - `classifyImageHttpError` falls back to `error.message` when the upstream body can't be parsed for 4xx errors.
- `gen.pollinations.ai/test/generation-vcr.test.ts`
  - Adds a regression case: upstream 400 with empty body now surfaces `"Image provider error: Image backend rejected request with status 400"` rather than `"Image provider error"`.

## Test plan
- [x] `npx vitest run test/generation-vcr.test.ts` — 13/13 pass
- [x] `npx vitest run test/error-observability.test.ts` — 1/1 pass
- [x] `npx biome check` clean on touched files
- [ ] After deploy, verify a failing `nanobanana-2` edit returns the underlying Gemini reason (currently masked as `"{}"`).

### DIFF
--- a/gen.pollinations.ai/src/image/handler.ts
+++ b/gen.pollinations.ai/src/image/handler.ts
@@ -252,7 +252,9 @@ function classifyImageHttpError(error: HttpError): {
         const text = parsed.text || error.message;
         return {
             status: remapUpstreamStatus(error.status),
-            message: text ? `Image provider error: ${text}` : "Image provider error",
+            message: text
+                ? `Image provider error: ${text}`
+                : "Image provider error",
         };
     }
 
--- a/gen.pollinations.ai/src/image/handler.ts
+++ b/gen.pollinations.ai/src/image/handler.ts
@@ -186,7 +186,7 @@ function throwImageError(error: unknown, c: ImageContext): never {
             requestUrl:
                 safeUpstreamUrl(error.upstreamUrl) ?? new URL(c.req.url),
             upstreamStatus: error.status,
-            responseBody: JSON.stringify(error.details || {}),
+            responseBody: imageResponseBody(error),
             cause: error,
         });
     }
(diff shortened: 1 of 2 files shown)


voodoohop:
Deployed to production and re-ran the original failing request.

**Before fix:** `upstreamBody: "{}"`, message "Image provider error"
**After fix:** `message: "Image provider error: Gemini: For information on human anatomy, resources like MedlinePlus from the National Library of Medicine are a reliable source."`

So the underlying `nanobanana-2` failure is **content moderation by Gemini 3.1 Flash** — it treats the "label the parts" prompt on this image as a medical/anatomy query and deflects to MedlinePlus instead of editing. `nanobanana` (Gemini 2.5) is more permissive on the same input. Not an infrastructure bug — just needed the message to surface so we could see it.

voodoohop:
Independent reviews complete:

**Codex CLI review**: ship it. Single concrete concern — `imageResponseBody` now surfaces `HttpError.details.body` / `error.message` to clients via `details.upstreamBody`. Confirmed no token/secret values appear in any `new HttpError(...)` site under `gen.pollinations.ai/src/image/` — the only match for 'token' is the literal string "Failed to get Google Cloud access token" at `veoVideoModel.ts:110`. Provider error bodies (the intended surface) are what we want exposed.

**Independent reviewer**: ship it. Same single concern. Also noted (deliberate trade-off):
- `createAndReturnImages.ts:522` and `azureFluxKontextModel.ts:209` now surface raw upstream errorText
- `fetchUpstream.ts:28` now surfaces `"<Provider> request failed: <body>"` instead of generic "Image provider error" — minor model→provider name disclosure, no secrets

**Coverage**: every image+video model funnels through `handler.ts:throwImageError`. Non-HttpError throws already passed `error.message` through unchanged (handler.ts:193-198). No bypass.

**Happy path**: `parseUpstreamErrorBody` still extracts `detail`/`message`/`error` first; the new `|| error.message` fallback is strictly a no-op when `parsed.text` is truthy.

Pushed format wrap fix for CI biome (line-length). Merging once CI is green.

PR #11016, opened 2026-05-16
fix(gen): redact secrets from text responses

### LINKED ISSUE #8913: Polly should mask/redact API keys in responses
**Problem:**
When users share error logs or code snippets containing their API keys (like in `api_key=sk_...` or `Authorization: Bearer pk_...`), the bot currently quotes them back verbatim in responses. This inadvertently exposes sensitive credentials even if the user later deletes their message.

**Example scenario:**
1. User pastes error: `"Fetching from... api_key=sk_jSnBVLc1kQSJOJSC0y1sYTeBDrYUHHSC"`
2. Bot responds quoting the full URL with key intact
3. User deletes their message, but bot's quote remains

**Expected behavior:**
Bot should auto-detect and mask API keys in its responses:
- `pk_xxx...` → `pk_[MASKED]`
- `sk_xxx...` → `sk_[MASKED]`
- `Bearer sk_...` → `Bearer [MASKED]`

**Implementation notes:**
- Regex pattern: `(pk_|sk_)[a-zA-Z0-9]{20,}`
- Apply masking before outputting any quoted user content
- Also mask in code blocks, URLs, and JSON

**Priority:** Security/privacy issue
**Reporter:** `snapabi33` (Discord)

---
**Author:** `snapabi33` (UID: `957905262428172298`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1479490353743663297/1479825819844087929)

fisventurous:
adds a small egress redaction layer so obvious Pollinations keys and bearer tokens do not leak through text responses, streams, upstream error bodies, or error observability events

changes
\- add a shared secret redaction utility for `sk_`, `pk_`, and `Bearer ...` values
\- redact OpenAI JSON responses, plain text responses, and streamed text chunks
\- redact `UpstreamError` messages, details, upstream bodies, stack/event fields before client responses and Tinybird events
\- add focused regression tests for normal output, split stream chunks, and upstream error leakage

Closes #8913


ElliotEtag:
useful PR — one concern on the streaming redactor that's probably worth addressing before merge:

`createSecretRedactionStream` in `src/utils/secret-redaction.ts` holds back up to `STREAM_TAIL_LENGTH = 4096` chars before emitting anything:

```ts
transform(chunk, controller) {
    pending += decoder.decode(chunk, { stream: true });
    if (pending.length <= STREAM_TAIL_LENGTH) return;
    ...
}
```

for SSE chat streams that's a flat ~4KB first-byte delay — at ~4 chars/token, roughly 1000 tokens of latency before the client sees anything. on slower models that's 5–20+ seconds of dead air, which will make streaming feel broken even though it's technically working.

**quick fix:** drop `STREAM_TAIL_LENGTH` to something like 256. covers any real-world `sk_*`/`pk_*`/`Bearer *` token (patterns are unbounded in regex but bounded in practice) and reduces latency by ~16×.

**cleaner fix:** most chunks contain none of the three anchors (`sk_`, `pk_`, `Bearer `) — detect that and pass them through with zero buffering:

1. find the rightmost index of any anchor in `pending`
2. no anchor → emit the whole buffer (redaction is a no-op), clear `pending`. zero latency.
3. anchor present → emit everything before it, keep only from the anchor onward in `pending`
4. cap `pending` at some sane max (e.g. 1KB) so adversarial input can't grow it unbounded — emit + redact + reset on overflow

normal output then streams with effectively no added latency, only the rare chunk containing an anchor pays a small buffering cost.

**also worth flagging:**
- the secret patterns are unbounded (`{6,}`, `{8,}`). combined with a flat tail, a hostile upstream could grow `pending` indefinitely before flush. bound it explicitly regardless of which fix you pick.
- tests cover correctness for split chunks but not latency. an assertion that a chunk with no secret anchor is emitted immediately (no buffering) would catch any future regression here.

happy to leave the broader redaction approach as-is — it's the streaming latency that's the user-visible issue.

fisventurous:
@ElliotEtag oh, well spotted. I've updated the TransformStream to use the dynamic anchor based lookahead you suggested. it now passes safe chunks with zero latency and bounds the pending buffer to 1KB to stop runaway growth. please do review once again.

fisventurous:
closing this after discussing on discord. came to a conclusion that silently mutating the outbound response stream doesn't really fit the gateway's thin, passive pipe design.

PR #11073, opened 2026-05-17
feat(image): migrate seedream5 + seedance-pro to Replicate; retire seedance

Itachi-1824:
Migrates the two viable Seedream/Seedance models to Replicate, where seedance-2.0 already lives. Consolidates inference on Replicate, consistent error handling, billing tied to where the rest of our infra is.

- `seedream5` → `bytedance/seedream-5-lite` — price stays at **$0.0525/img** (Replicate's $0.035 × our 1.5× markup = existing price, no user impact)
- `seedance-pro` → `bytedance/seedance-1-pro-fast` — switches from per-token to **$0.025/s at 720p**

`seedance` (Lite) is **retired from the registry**. Replicate's `bytedance/seedance-1-lite` endpoint reproducibly fails with `Service is temporarily unavailable (E004) (1cah9wlWR9)` at every resolution (verified 2026-05-18 and 2026-05-19), and BytePlus is being deprecated, so neither path is viable. The model is removed entirely; revisit when Replicate restores Lite.

Both video models locked to 720p for now since the registry only carries a single per-second rate. Tiered 480p/1080p can come back in a follow-up when the registry grows per-resolution support. Model IDs/endpoints/response shapes for the two migrated models unchanged.

Old BytePlus seedance handler (`seedanceVideoModel.ts`) is gone along with its `classifyByteplusError` test. The legacy `seedream` 4.0 / `seedream-pro` 4.5 paths are still BytePlus and untouched — they live in `seedreamModel.ts` and continue to need `BYTEDANCE_API_KEY`.

`scripts/test-replicate-migration.mjs` is a small chained T2I → I2I/I2V harness for eyeballing output (output dir gitignored). Kept in-tree so the next Replicate model add has something to crib from.

The Replicate seedance-pro handler honors the documented schema contract: when clients pass `width`/`height` without `aspectRatio`, derives the closest Seedance-supported ratio (16:9, 4:3, 1:1, 3:4, 9:16, 21:9, 9:21).


ElliotEtag:
@Itachi-1824 — local Replicate test results (ran twice, same outcome):

| Model | Runs | Result |
|---|---|---|
| `bytedance/seedream-5-lite` (seedream5) | 2.0M | ✓ T2I + I2I |
| `bytedance/seedance-1-pro-fast` (seedance-pro) | 1.5M | ✓ T2V + I2V |
| `bytedance/seedance-1-lite` (seedance) | 3.4M | ✗ 100% fail, `E004 (1cah9wlWR9)` |

`seedance-1-lite` reproducibly fails for every input (720p/480p, T2V/I2V) within 1-3s. Same Replicate internal incident ID `1cah9wlWR9` across all attempts. Direct curl to their API has the same result, so it isn't our handler code or account quota — `seedance-1-pro-fast` works on the same token.

Options:
- Ship seedream5 + seedance-pro only; keep `seedance` on BytePlus for now.
- Route `seedance` → `bytedance/seedance-1-pro-fast` under the hood until Lite is restored (cost goes from `\$0.036/s` → `\$0.025/s` and quality changes — silent downgrade).
- Route `seedance` → `bytedance/seedance-1-pro` (the non-fast 480p/1080p variant; not currently in the PR).

Also worth quoting prediction `tev7wzctghrmw0cy7cjvbyrckg` to Replicate support — the consistent `E004` with the same incident ID looks like a known model-side break.

Other items from review (separate from rollout decision): rebase needed (PR is CONFLICTING); `test/video-error-classification.test.ts` deletion also dropped the still-live `resolveSeedanceV2AspectRatio` / `callSeedanceV2API` tests; new seedance handler reads `safeParams.aspectRatio` only — clients passing `width`/`height` without aspectRatio now default to 16:9 instead of the derived ratio the old handler returned.

PR #11088, opened 2026-05-17
fix(image): normalize whitespace + case in model param

### LINKED ISSUE #9468: [Bug] Image generation fails intermittently with "Invalid model" error despite valid model name
### Description

Image generation API returns "Invalid model" error for `zimage` model intermittently. The same request works during daytime (Turkey timezone, ~09:00-18:00) but fails at night (~20:30).

### Environment

- **API Endpoint:** `https://gen.pollinations.ai/image/{prompt}`
- **Integration:** n8n workflow 
- **Timezone:** Europe/Istanbul

### Request Details

```
GET https://gen.pollinations.ai/image/{encoded_prompt}?model=zimage&width=1080&height=1920&nologo=true&enhance=true&private=false
```

### Error Response

```json
{
  "success": false,
  "error": {
    "message": "Query parameter validation failed",
    "code": "BAD_REQUEST",
    "timestamp": "2026-03-22T17:01:47.815Z",
    "details": {
      "name": "ValidationError",
      "formErrors": [],
      "fieldErrors": {
        "model": ["Invalid option: expected one of \"kontext\"|\"nanobanana\"|\"nanobanana-2\"|\"nanobanana-pro\"|\"seedream5\"|\"seedream\"|\"seedream-pro\"|\"gptimage\"|\"gptimage-large\"|\"flux\"|\"zimage\"|\"veo\"|\"seedance\"|\"seedance-pro\"|\"wan\"|\"klein\"|\"p-image\"|\"p-image-edit\"|\"p-video\"|\"nanobanana2\"|\"gpt-image\"|\"gpt-image-1-mini\"|\"gpt-image-1.5\"|\"gpt-image-large\"|\"z-image\"|\"z-image-turbo\"|\"veo-3.1-fast\"|\"video\"|\"wan2.6\"|\"wan-i2v\"|\"flux-klein\"|\"pruna-image\"|\"pruna\"|\"pruna-edit\"|\"pruna-image-edit\"|\"pruna-video\""]
      }
    }
  },
  "status": 400
}
```

### The Problem

The error message says `zimage` is an "Invalid option" but **`zimage` is listed as a valid option** in the same error message's expected values list.

### Observed Behavior

| Time (Turkey) | Result |
|---------------|--------|
| 09:00 - 18:00 | ✅ Works |
| 20:30+ | ❌ Fails with "Invalid model" |

### Steps to Reproduce

1. Send GET request to `https://gen.pollinations.ai/image/{prompt}?model=zimage&width=1080&height=1920`
2. During daytime: Works correctly, returns image
3. During nighttime (~20:30 Turkey time / ~17:30 UTC): Returns 400 error

### Expected Behavior

The API should accept `zimage` as a valid model consistently, regardless of time of day.

### Additional Context

- Same workflow, same parameters, same prompt
- No changes to the request between successful and failed attempts
- Other models also do not works
- The error lists `zimage` as valid but rejects it simultaneously

### Possible Causes

1. Server-side model availability changes at certain times?
2. Load balancing to different servers with different model configurations?
3. Rate limiting that manifests as model validation error?

ElliotEtag:
n8n templates often inject trailing newlines or uppercased strings into the `model` query param. The current image route then 400s with `"Invalid option: expected one of ... zimage ..."` — listing `zimage` as valid while rejecting `"zimage "` or `ZIMAGE`. Reported by users hitting this from n8n at night (time-of-day was a red herring — it's whatever workflow runs at that hour that introduces the whitespace).

- `gen.pollinations.ai/src/schemas/image.ts` — preprocess the model param to trim + lowercase before the enum check. The query validator runs ahead of the `resolveModel` middleware, so this layer needs the fix on its own.
- `shared/registry/registry.ts` — apply the same normalization in `resolveModelName` so JSON POST paths (`/v1/images/generations` etc.) and other categories benefit too.
- New unit tests on `resolveModelName` for whitespace, casing, aliases, and the still-rejects-unknown case.

**Out of scope:** literal surrounding quotes (e.g. `model="zimage"`). Trim/lowercase covers the reported failure modes; quote-stripping would start swallowing user-side template errors and isn't requested.

Verified locally with `wrangler dev`:
- `model=zimage%20`, `model=%0Azimage`, `model=ZIMAGE` now pass query validation (401 = auth, not 400 = validation)
- `model=not-a-real-model` still 400s as expected

213/213 gen tests pass, 191/191 enter alias tests pass.

Fixes #9468

### DIFF
--- a/enter.pollinations.ai/test/aliases.test.ts
+++ b/enter.pollinations.ai/test/aliases.test.ts
@@ -40,6 +40,27 @@ test.for(
     expect(resolved).toBe(shouldResolveTo);
 });
 
+test("resolveModelName tolerates surrounding whitespace", () => {
+    for (const raw of ["zimage ", "  zimage", " zimage\n", "\tzimage\t"]) {
+        expect(resolveModelName(raw)).toBe("zimage");
+    }
+});
+
+test("resolveModelName lower-cases the input", () => {
+    expect(resolveModelName("ZIMAGE")).toBe("zimage");
+    expect(resolveModelName("Flux")).toBe("flux");
+});
+
+test("resolveModelName normalizes aliases too", () => {
+    expect(resolveModelName(" Z-Image ")).toBe("zimage");
+});
+
+test("resolveModelName still throws on unknown models", () => {
+    expect(() => resolveModelName("definitely-not-a-real-model")).toThrow(
+        /Invalid model or alias/,
+    );
+});
+
 test("cost lookup uses the public model name instead of collapsing shared provider ids", () => {
     const usage = {
         promptTextTokens: 1_000_000,
--- a/gen.pollinations.ai/src/schemas/image.ts
+++ b/gen.pollinations.ai/src/schemas/image.ts
@@ -20,7 +20,11 @@ const GenerateImageRequestQueryParamsBaseSchema = z.object({
     // Image model params
     model: z
         .preprocess(
-            (val) => (val === "" ? undefined : val),
+            (val) => {
+                if (typeof val !== "string") return val;
+                const normalized = val.trim().toLowerCase();
+                return normalized === "" ? undefined : normalized;
+            },
             z
                 .enum(VALID_IMAGE_MODELS as unknown as [string, ...string[]])
                 .optional()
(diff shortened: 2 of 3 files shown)


voodoohop:
Closing this because it changes the API behavior in the wrong direction.

- `?model=zimage` is accepted in production.
- The reported validation error is reproduced by malformed values like `zimage%20` or `ZIMAGE`.
- We should keep model names strict instead of silently normalizing misspellings/casing/whitespace server-side.
- A better follow-up would be improving the validation error to show the received value, so hidden whitespace is obvious.

PR #11039, opened 2026-05-17
feat: log request inputs in error events

ElliotEtag:
- Adds `request_inputs` as a JSON blob on existing Tinybird `error_event` rows.
- Captures route params, query params, and request body without per-parameter columns.
- Keeps direct `return c.json(..., 500)` paths unchanged and out of `error_event`.
- Extends `error_event` retention and `recent_server_errors` default lookback from 3h to 24h.
- Requires Tinybird validation before merge: `tb --cloud deploy --check --wait`.

PR #11111, opened 2026-05-18
fix(gen): capture upstream body when chat response fails schema validation

voodoohop:
## Summary
- Wrap chat-completion response parse in try/catch and re-throw as `UpstreamError` so the raw upstream body is logged via `error_event.upstream_body` when validation fails
- Currently a plain `ZodError` is thrown and the malformed JSON is discarded — we get 22+ daily 500s on `mistral-large` with `usage:null` and no way to see what Portkey/Azure actually returned
- Schema NOT loosened: we still reject responses without proper usage (billing depends on it); this PR just gives us debug visibility into the root cause

## Background

Investigation of Mistral errors (#11039 logging) surfaced a pattern: `mistral-large` + `response_format: json_object` + non-streaming returns ZodError "usage: expected object, received null". Direct Azure + direct Portkey replays of the failing payload both return proper usage objects. Production fails with empty `upstream_body` because the Zod error path doesn't capture it.

## Test plan
- [x] `npx vitest run` (220/220 pass)
- [x] `npx tsc --noEmit` clean on gen + enter
- [ ] After deploy: confirm next `usage:null` 500 has populated `upstream_body` in Tinybird `error_event` so we can finally see what Portkey is returning

### DIFF
--- a/gen.pollinations.ai/src/routes/proxy.ts
+++ b/gen.pollinations.ai/src/routes/proxy.ts
@@ -136,13 +136,23 @@ const chatCompletionHandlers = factory.createHandlers(
         // add content filter headers if not streaming
         let contentFilterHeaders = {};
         if (!c.var.track.streamRequested) {
-            const responseJson = await response.clone().json();
-            const parsedResponse = CreateChatCompletionResponseSchema.parse(
-                responseJson,
-                { reportInput: true },
-            );
-            contentFilterHeaders =
-                contentFilterResultsToHeaders(parsedResponse);
+            const responseText = await response.clone().text();
+            try {
+                const parsedResponse = CreateChatCompletionResponseSchema.parse(
+                    JSON.parse(responseText),
+                    { reportInput: true },
+                );
+                contentFilterHeaders =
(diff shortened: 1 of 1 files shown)


voodoohop:
Reviewed from a code-review stance.

No blockers found.

- Schema remains strict: `usage: null` is still rejected, so we do not serve unbillable successful responses.
- The change is scoped to the non-streaming chat response validation path and converts JSON/schema parse failures into `UpstreamError`.
- That means the next `usage:null` failure should populate `error_event.upstream_body` with the raw upstream response, using the existing 16KB truncation in `error.ts`.
- Client-facing status changes from plain 500 to 502 for this upstream contract failure, which matches the existing `UpstreamError` semantics.

Local checks passed:
- `npx vitest run test/error-observability.test.ts test/openai-schema.test.ts`
- `npx tsc --noEmit` in `gen.pollinations.ai`
- `npx biome check gen.pollinations.ai/src/routes/proxy.ts`

I tried to submit an approval, but GitHub rejected it because this authenticated account is the PR author.

PR #11131, opened 2026-05-18
fix(gen): return 400 when upstream rejects unsupported modality

voodoohop:
## Summary
- Before: clients sending \`image_url\` to text-only models (e.g. \`deepseek\`) got back a 502 because we remapped OpenRouter's 404 'no endpoints found that support image input' as a gateway error
- After: those user-input errors return 400 so the client sees them as their bug, not ours

## Detection
With \`request_inputs\` logging now in place, we can confirm 100+ daily errors of this shape, dominated by one user sending images to deepseek

## Test plan
- [x] \`npx tsc --noEmit\` clean
- [x] \`npx vitest run\` (220/220 pass)

### DIFF
--- a/gen.pollinations.ai/src/error.ts
+++ b/gen.pollinations.ai/src/error.ts
@@ -60,13 +60,26 @@ export async function ensureUpstreamOk(
     const rawMessage =
         extractUpstreamMessage(responseBody) ||
         getDefaultErrorMessage(response.status);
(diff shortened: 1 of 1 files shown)


voodoohop:
Superseded by #11139, which validates image input against the model registry before calling the provider. That is cleaner than parsing the upstream error string for this case.

PR #11141, opened 2026-05-18
fix(registry): correct text-model inputModalities to match real upstream support

voodoohop:
## Summary
Audit via live test requests to gen.pollinations.ai exposed 5 mismatches between our registry's \`inputModalities\` and what the upstream provider actually accepts. Important prerequisite for #11139 (which uses this field to validate requests before forwarding).

## Corrections

| Model | Before | After | Evidence (1×1 PNG test) |
|---|---|---|---|
| \`openai-audio\` | \`["text","image","audio"]\` | \`["text","audio"]\` | Azure: \"This model does not support image_url content\" |
| \`openai-audio-large\` | \`["text","image","audio"]\` | \`["text","audio"]\` | Azure: same error |
| \`perplexity-fast\` | \`["text"]\` | \`["text","image"]\` | 200 OK: \"Black\" |
| \`perplexity-reasoning\` | \`["text"]\` | \`["text","image"]\` | 200 OK |
| \`nova\` (Nova 2 Lite) | \`["text"]\` | \`["text","image"]\` | 200 OK: \"Black.\" |

## Models verified correct (no change)

Text-only as marked: \`qwen-coder\`, \`qwen-coder-large\`, \`qwen-safety\`, \`deepseek\`, \`deepseek-pro\`, \`glm\`, \`llama\`, \`minimax\`, \`nova-fast\`. (Some return 200 but explicitly state they cannot see images; some 400 from upstream.)

\`midijourney\` / \`midijourney-large\` are specialized music-composition wrappers around Claude. The underlying Claude model accepts images, but the product is intentionally text-only — left unchanged.

## Test plan
- [x] \`npx tsc --noEmit\` clean on both gen and enter
- [x] \`npx vitest run\` (223/223 pass)
- [x] biome clean


voodoohop:
Verified the reviewer's question empirically in prod. Nova 2 Lite gateway behavior:

- `nova` + `video_url` content part → 200 OK upstream, but when asked honestly the model says "As an AI text-based model, I don't have the capability to view or receive files, including videos." — content is silently dropped before reaching the model
- `nova` + `input_audio` content part → 200 OK, model says "I cannot hear any audio you might be referring to" — same silent drop

Root cause: our chat transform pipeline (`imageUrlToBase64Transform.ts`) only handles `image_url` parts; `video_url` and `input_audio` are forwarded as-is, and Bedrock's OpenAI-compatible adapter drops what it doesn't understand. `gemini` rejects video URLs with a clean 400 instead, but only because Vertex hard-validates URL reachability.

Confirms the conservative scope is correct — `inputModalities` should reflect what our chat path end-to-end supports, not raw vendor capability. Updated the skill doc with this distinction (last commit). When we do want to expose video/audio for these models, we'd need to add the transform + verify round-trip first, then update the registry.

PR #11169, opened 2026-05-19
lock down staging signups to github id allowlist

### LINKED ISSUE #11137: 🔒 Lock down staging signup to prevent test-card abuse
## Problem
`staging.enter.pollinations.ai` is publicly accessible. Anyone can:

1. Sign in via GitHub OAuth and create an account
2. Use Stripe test card `4242 4242 4242 4242` on staging checkout → sandbox webhook fires → pollen credited
3. Mint a staging API key, hit `staging.gen.pollinations.ai` → **real OpenAI / Google / etc. costs billed to us** (gen-staging holds the production provider keys)

Stripe side is fake money; upstream provider bill is real.

## Recommended fix (cheapest)
Email domain allowlist in the auth callback, gated by environment:

```ts
if (c.env.ENVIRONMENT === 'staging') {
  const allowed = ['myceli.ai', 'pollinations.ai'];
  if (!allowed.some(d => email.endsWith('@' + d))) {
    return new Response('staging is invite-only', { status: 403 });
  }
}
```

- ~5 lines in `enter.pollinations.ai/src/auth.ts` (Better Auth callback or similar)
- Prod path untouched
- New team members just need a `@myceli.ai` / `@pollinations.ai` email
- No infra dependency

## Alternatives
- **GitHub username allowlist** — same code shape, env var of usernames, requires redeploy to add people
- **Cloudflare Access on `staging.*` domains** — zero code, ~15 min in CF dashboard, blocks unauthenticated traffic before the worker. Needs CF service token added to Stripe webhook config though
- **Disable signup entirely on staging** — pre-create accounts via admin route. Most secure, most friction

## Related
- #11127 (general staging follow-ups)
- #11128 (enter staging worker isolation)
- #11135 (gen staging worker isolation)

ElliotEtag:
Closes #11137.

`staging.enter.pollinations.ai` was publicly accessible — anyone could sign in via GitHub OAuth, mint an API key, and hit `staging.gen.pollinations.ai` which holds the **production** provider keys. Sandbox Stripe = fake money in, real OpenAI/Google/etc. bills out.

This gates GitHub OAuth sign-in on staging to an env-var allowlist of numeric GitHub IDs (`STAGING_ALLOWED_GITHUB_IDS`). Two Better Auth `databaseHooks`:
- `user.create.before` blocks new signups whose `githubId` isn't in the set
- `session.create.before` blocks every login (covers users that already exist in staging D1)

Both throw `APIError("FORBIDDEN")`, which Better Auth redirects to `/error?error=staging_is_invite-only`. Plugin is a no-op outside staging.

GitHub **IDs**, not usernames — usernames are mutable (silent lockout if someone renames). Not the DB user ID either — that only exists post-signup, so it can't gate a fresh OAuth callback.

Existing staging D1 already cleaned: 4 external dev accounts deleted + cascades. Remaining 5 users match the allowlist.

Initial allowlist:
- `36901823` ElliotEtag
- `5099901` voodoohop
- `235942848` thomashmyceli
- `248917639` fouchyelliot-rgb
- `241978997` pollen-router

Production untouched.


voodoohop:
Pushed a merge of latest `main` plus three fixes (PR was blocking before):

- **gen staging allowlist (blocker):** `gen.pollinations.ai/wrangler.toml` set `ENVIRONMENT="staging"` but never defined `STAGING_ALLOWED_GITHUB_IDS`. Since the request-time gate now runs inside `authenticateApiKeyRequest` with gen's own `c.env`, the empty allowlist made it fail closed and 403 *every* staging-gen request — including the allowlisted devs, on the worker that holds prod provider keys. Added the var to `[env.staging.vars]` + the bindings type.
- **rebase:** resolved conflicts with main in `enter` `auth.ts` (kept `authAdditionalFields`), `wrangler.toml` (kept `STRIPE_PMC` + allowlist), and regenerated `worker-bindings.d.ts`.
- **error page:** `error.tsx` now shows a staging-specific message for `staging_is_invite-only` instead of the generic error.

Verified: enter + gen `typecheck` pass, biome clean. Squash auto-merge enabled.

Open follow-up (separate issue, non-blocking): confirm whether staging-gen bills anonymous/keyless requests — those skip `authenticateApiKeyRequest` entirely and aren't gated.

PR #11172, opened 2026-05-19
feat: country-routed checkout (phase 1 backend)

ElliotEtag:
Routes Stripe Checkout sessions by buyer cohort based on `CF-IPCountry` (Cloudflare-set, not spoofable), but keeps pricing USD-native and lets Stripe Adaptive Pricing handle buyer-currency localization.

Earlier iterations of this branch tried per-cohort integration currencies (EUR/INR/GBP) backed by frankfurter.dev FX so we could unlock UPI, Pix, and mainland Alipay as currency-locked payment methods. The cost (FX cache, multi-currency settlement plumbing, extra Tinybird modeling) wasn't justified by the short-term volume, so checkout collapsed back to USD + Adaptive Pricing in 3c2c05125. The cohort enum stays as the hook point for any future currency-routed checkout and for analytics.

**What this PR ships**

- USD-native `price_data` (`unit_amount: pack.amountUsd * 100`) with `adaptive_pricing: { enabled: true }` on every checkout session — Stripe localizes presentment per buyer where supported (see `enter.pollinations.ai/src/routes/stripe.ts` lines 107 and 121)
- `getCohortFromCountry(cf-ipcountry)` resolves a `CheckoutCohort` enum (USD / BR / APAC_ALIPAY / EU_CORE / INDIA / UK). The cohort id is written to `session.metadata` + `payment_intent.metadata` only — Stripe stores it for free, no schema commitment downstream. It does not influence pricing, PMC, or currency, and is not forwarded to Tinybird (geographic signals already on `stripe_event` — `card_country` from Radar and buyer billing-address country via the stored Stripe `payload` JSON — cover the analytics need)
- MO routes to USD default rather than APAC_ALIPAY — a 5,000-charge live audit found 99.8% of MO billing-country charges were US-issued cards (card-testing fingerprint). Covered by `currency-router.test.ts`
- Pollen grants stay USD-anchored (`5+1 = 6 pollen` regardless of buyer-displayed currency)
- Single `STRIPE_PMC` env var per environment; Stripe filters methods by buyer (currency, country) at runtime. Sandbox + Live both point at the repurposed account Default PMC (renamed `top_up`)
- Legacy `/checkout/<USD-amount>` URLs removed — all first-party callers and `/api/stripe/products` expose `packKey` (`p2`..`p100`)
- Auto-top-up paths in `stripe-billing.ts` are USD-only (non-USD invoices rejected with `currency mismatch`)

**Explicitly not in this PR (was in earlier drafts)**

- Per-cohort integration currencies, frankfurter.dev FX cache (`fx-cache.ts`, `fx-cache.test.ts`, `shared/test/mocks/frankfurter.ts`), and matching settlement plumbing — all removed in 3c2c05125
- UPI / mainland Alipay / Pix via currency-locked sessions — those require a currency-routed checkout and are deferred

**Verification**

- Deployed to `staging.enter.pollinations.ai` (sandbox Stripe). Real-IP-verified end-to-end across BR / EU / APAC / IN / UK / US — every cohort resolves to a USD session, with Adaptive Pricing surfacing local-currency presentment in EU / APAC test cases
- Tests green: `test/currency-router.test.ts` (41 cases), `test/pollen-packs.test.ts`, `test/integration/stripe.test.ts` (62 cases incl. cohort metadata + packKey flows)
- Dashboard panels in `provisioning/dashboards/pollen-flow.json` normalized to handle currency mixing in historical data — Adaptive Pricing settles in multiple currencies even when integration currency is USD

**Follow-up**

#11333 — dashboard accuracy gap if promo codes are ever issued on non-USD checkouts (low priority, no promo codes today)


ElliotEtag:
Pushed 582bfce3d addressing the review:

- **Finding 2 — async_payment_failed:** now forwards cohort + payment_methods_offered to Tinybird (was log-only). New integration test `POST /api/webhooks/stripe emits checkout.session.async_payment_failed to Tinybird with cohort`. Matters specifically because BR Pix and EU SEPA are the delayed-payment methods this rollout adds.
- **Finding 3 — biome on generated d.ts:** root cause was a typo in `biome.jsonc` ignore list (`worker-bindins.d.ts`). Fixed; the generated file is now correctly ignored.
- **Finding 4 — stale plan:** branch name, worker version, test count (118/118), and Tinybird cohort-column status all refreshed.

**Finding 1 — prod PMC config (`STRIPE_PMC_*` in production wrangler.toml):** intentionally deferred. Tracked as Stage D in the pre-merge checklist — live PMCs need to be created in the production Stripe dashboard first (separate from sandbox). The worker fails closed (HTTP 500 + log) if any cohort's PMC env is missing, so the worst case if this reached prod prematurely is hard checkout failure, not silent mis-routing. Will be filled in alongside the live PMC setup before promoting to prod.

PR #11227, opened 2026-05-21
feat: add Gemini dynamic billing rules

voodoohop:
WIP draft.

- Adds a model-level cost override hook while keeping price multiplier logic centralized.
- Adds reusable Gemini billing for search grounding and the >200k prompt-token tier.
- Threads generic response output into tracking so registry billing can read grounding metadata.
- Covers the new billing paths with registry tests and a local smoke check.


ElliotEtag:
Reviewed the billing hook — core design is right: the optional `calculateCost` per model with `linearCost` injected keeps token math centralized and is fully backwards-compatible. A few things worth addressing before merge:

**1. Surface fees as named line items.** `extraCost` only lands in `totalCost`, so the search fee is invisible in the per-type breakdown — no `x-usage-*` header, no `generation_event` column, and `sum(line items) ≠ totalPrice`. For a billing change that makes provider-invoice reconciliation hard. Suggest emitting it as its own cost key (e.g. `searchFee`).

**2. Signal on zero-detection.** If grounding metadata is missing (non-JSON content-type, stream without `groundingMetadata`, provider quirk), it silently bills 0 → under-charge with no trace. A log line when search ran but query count = 0 makes the leak observable.

**3. Collapse the per-model calculators into one declarative mechanism.** `calculateGeminiSearchQueryCost` / `…GroundedPromptCost` / `…31ProCost` are three bespoke functions; every new provider adds another. They're the same formula — `count × unitCost` added to the total — with a different `count` (per-query for Gemini 3, flat-if-grounded for 2.5). Suggest a declarative `surcharges: [{ label, unitCost, count }]` on the model + a small central counter dispatch (`always` | `ifGrounded` | `geminiQueries`), keeping `calculateCost` as the escape hatch for rate-tier cases (the >200k tier). One path for all fixed per-request fees.

Context: we're adding Perplexity Sonar (#11358), which also carries a per-request search fee. It maps onto this as `count: "always"` (Sonar always searches — verified live, citations on every response incl. "2+2") with no output parsing needed. Worth designing the mechanism once here so Perplexity and future providers reuse it instead of adding calculator #4.

One decision to make explicit: whether the fee should be marked up by `priceMultiplier` (currently it is, via `totalPrice = totalCost × multiplier`) or billed at cost.

**Next steps:** (1) name the fee line item, (2) add the zero-detection log, (3) generalize to declarative surcharges. Happy to take the surcharge refactor and wire Perplexity onto it.

voodoohop:
Superseded by #11683 — same Gemini coverage reimplemented as declarative billing rules, plus Perplexity request fees. Closing.

PR #11237, opened 2026-05-22
fix(enter): repair four pre-existing test failures

voodoohop:
- Preserve query string in `/api/docs` -> `/docs` redirect (was dropping `?format=json` after #11130)
- Compute future period dynamically in `account-usage` bounds test (hardcoded `2026-05-01` is now in the past)
- Regenerate `effective-prices` snapshot for the gemini-search rate reset in #11226
- Remove obsolete `aliases` cost-collision test (data eliminated by #11226; `resolveModelName` tests still cover the invariant)

All 196 tests in the four affected files pass.


miguelsaldanha1213-byte:
"Plants can grow in dirty water, so don't give up on growing because of criticism. Look at them and keep growing."

PR #11271, opened 2026-05-24
fix(proxy): pass WebSocket 101 responses through unchanged

voodoohop:
## Summary

- WebSocket upgrade responses through `*.pollinations.ai` were failing with HTTP 500 at the edge.
- Root cause: the proxy worker wraps the upstream response in `new Response(body, init)`, which drops Cloudflare's `webSocket` field — the handshake never completes.
- Fix: detect `Upgrade: websocket` requests and return the upstream response verbatim when the status is 101.

## Test plan
- [x] `npm run deploy:staging` from `pollinations-myceli-proxy/` (version `73bd8ff1`)
- [x] WS upgrade through `wss://staging.gen.pollinations.ai/v1/realtime?model=gpt-realtime-2` succeeds (HTTP 101, full round-trip to OpenAI realtime confirmed)
- [x] Non-WS traffic through the proxy unchanged

Unblocks `/v1/realtime` (and any future WS endpoint) through the public proxy.

### DIFF
--- a/pollinations-myceli-proxy/src/index.ts
+++ b/pollinations-myceli-proxy/src/index.ts
@@ -37,6 +37,9 @@ export default {
         // Override Host so the upstream Worker's routing/SNI is correct.
         headers.set("Host", upstreamHost);
 
+        const isWebSocketUpgrade =
+            req.headers.get("Upgrade")?.toLowerCase() === "websocket";
+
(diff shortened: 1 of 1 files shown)


voodoohop:
Folded into #11272 — the proxy WS pass-through is required for the realtime route to work through the public host, so they ship together.

PR #11308, opened 2026-05-27
feat: add lightweight media catalog

voodoohop:
- Adds R2-backed catalog sidecars for owner media, public app galleries, and public remix children.\n- Exposes key id, user id, and BYOP app attribution from /account/key for server-stamped indexes.\n- Saves CatGPT generated memes through media.pollinations.ai and reads user/public galleries.\n- Stamps voice-edit remixes with parent hashes and shows a public remixes strip.\n\nTests:\n- media.pollinations.ai: npm run test\n- enter.pollinations.ai: npx vitest run test/account-key.test.ts\n- root: npx biome check --write touched JS/TS/CSS files\n- root: git diff --check; node --check apps/catgpt/ai.js apps/catgpt/script.js

### DIFF
--- a/apps/catgpt/ai.js
+++ b/apps/catgpt/ai.js
@@ -9,6 +9,8 @@ const SELFIE_CATGPT = "https://media.pollinations.ai/657d58ee4c9c22d7";
 const AUTH_KEY = "catgpt_api_key";
 const APP_KEY = "pk_uWjreBEkxFAhjDHo";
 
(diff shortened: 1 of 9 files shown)


voodoohop:
Superseded by #11309, which is based on main and brings in only apps/voice-edit/remix.html and apps/voice-edit/walkthrough.html from the pointer-editor branch.

PR #11313, opened 2026-05-27
feat: add sidecar media catalog

voodoohop:
- Adds R2 sidecar media catalog indexes for owner media, app galleries, tags, and remix children.
- Stamps owner and app attribution only from verified /account/key identity.
- Wires CatGPT, voice-edit remix/walkthrough, Chat, AI Dungeon Master, Virtual Makeup, and Product Packaging uploads.
- Keeps public listings free of owner/API-key fields; private media stays out of public indexes.
- Tests cover app spoofing, lineage, tags, visibility, and account key identity.


voodoohop:
Closing as superseded by the newer tag-only media catalog direction in #11330. Keeping the newer/reference drafts open for comparison.

PR #11311, opened 2026-05-27
feat(media): lightweight catalog and app integrations

voodoohop:
- Adds an R2-backed media catalog index for owner, verified app, public tags, and public remix children.
- Exposes server-attested key/user/app identity from /account/key.
- Wires CatGPT, voice-edit remix/walkthrough, AI Dungeon Master, Product Packaging Designer, Virtual Makeup, Chat, SlidePainter, and Pollicraft into catalog uploads where auth/media flow already exists.
- Keeps app identity server-stamped; unverified app tags remain queryable but are not trusted for attribution.

Validation:
- npx biome check --write apps/chat/src/utils/api.js apps/slidepainter/src/services/pollinations.ts
- node --check apps/chat/src/utils/api.js
- Pollicraft module script syntax check
- apps/chat: npm run build
- apps/slidepainter: npm run build
- media.pollinations.ai: npm run test
- enter.pollinations.ai: npm run decrypt-vars && npx vitest run test/account-key.test.ts
- app builds: ai-dungeon-master, product-packaging-designer, virtual-makeup
- node syntax checks: CatGPT JS, voice-edit remix/walkthrough module scripts

PR #11310, opened 2026-05-27
feat: add media history catalog

voodoohop:
- Adds a D1 media catalog for uploads with owner, verified app, tags, and parent-child history.
- Exposes verified user/app key identity from /account/key for server-attested app attribution.
- Adds /me/media, /apps/:appKeyId/media, and /:hash/children endpoints.
- Wires CatGPT, voice-edit remix/walkthrough, AI Dungeon Master, Virtual Makeup, and Product Packaging uploads into the catalog.
- Requires provisioning a real MEDIA_DB D1 binding before deploy; apps need registered pk_ app keys for verified app attribution.

PR #11306, opened 2026-05-27
feat(media): server-attested provenance + remix lineage

voodoohop:
Smallest change that lets client-only apps (CatGPT, voice-edit) have a clean public gallery, a personal "my media" list, and a remix-lineage tree, without new infra.

Mechanism: media.pollinations.ai writes tiny JSON index entries under stable R2 prefixes on every upload. Three new read endpoints `list({prefix})` over those. Identity facets (`owner`, `app`) are stamped from the verified `/account/key` response and are not readable from request params.

Why R2 prefix, not D1/KV: only query needed is "list children of X" / "list by owner" / "list by app". No joins, no random tag lookup. Same shape works on S3 (ListObjectsV2) unchanged.

Changes:
- `enter.pollinations.ai/src/routes/account.ts` — `/account/key` now also returns `userId`, `appId`, `appName` (additive, backwards-compatible)
- `media.pollinations.ai/src/index.ts` — upload accepts optional `parent` + `tags`; writes index entries; new endpoints `GET /:hash/children`, `GET /me/media`, `GET /apps/:app/media`
- 11 new tests covering the trust boundary, lineage round-trip, listing isolation, tag normalization, and the legacy-key-without-userId case

Storage layout (R2):
```
catalog/v1/items/<hash>.json
catalog/v1/by-owner/<userId>/<revTs>-<hash>.json
catalog/v1/by-app/<appId>/<revTs>-<hash>.json
lineage/v1/by-parent/<parent>/<revTs>-<child>.json
tags/v1/<userTag>/<revTs>-<hash>.json
```

`revTs` is `MAX_SAFE_INTEGER - Date.now()` so lex-sort is newest-first.

Not in this PR:
- Frontend integration (voice-edit branch strip, CatGPT gallery) — separate
- Retention/billing for index entries — defer to #10832 work
- Closure tables, edit-verb classification, abuse pipelines — deferred until shape is in production
- Cross-app moderation/takedown — separate concern

Tests: 17/17 pass under cloudflare/vitest-pool-workers.

PR #11321, opened 2026-05-27
feat(media): minimal catalog — opt-in, tag-only, no privacy lie

voodoohop:
The simplest catalog that still solves communal galleries and lineage. Drops `visibility`, `parents[]`, and `relationship` from the other drafts (#11306–#11313). Lineage and visibility become *conventions* on top of tags.

## Three premises

1. **The hash is the capability.** `GET /:hash` is public, period — anyone who has the hash gets the bytes. There is no "private" mode and the catalog doesn't pretend otherwise. *Don't share the hash if you don't want the bytes seen.* This is honest about what content-addressed storage actually provides; the other drafts' `visibility: \"private\"` was advisory metadata only and never enforced anywhere.

2. **Catalog inclusion is opt-in.** An upload appears in `/tags/:tag` only if the uploader explicitly tagged it. Untagged uploads are visible only via `/me/media` (the owner's own inbox).

3. **Server-attests owner and app from the verified key.** Never from request params. The adversarial test confirms it.

## What's in the catalog entry

- `owner`, `app`, `appName` — server-stamped from `/account/key`
- `tags[]` — sanitized, ≤8
- `prompt`, `model` — optional, for gallery cards
- `contentType`, `size`, `createdAt`
- *nothing else* — no `visibility`, no `parents`, no `relationship`

## Endpoint surface — two new routes

- `GET /me/media?limit=&cursor=` — authed; your own uploads, newest first
- `GET /tags/:tag?app=<keyId>&limit=&cursor=` — public; tagged uploads, optionally scoped to one app's namespace

That's it.

## How conventions cover what the other drafts modelled with fields

| Use case | Convention |
|---|---|
| App's public gallery | App tags uploads with `tag=<app>` (e.g. `tag=catgpt`) → `GET /tags/catgpt` |
| Lineage (parent → child) | Tag the child with `tag=parent:<hash>` → `GET /tags/parent:<hash>` lists the children |
| Multi-parent (pollicraft: water + fire → steam) | `tag=parent:<water>`, `tag=parent:<fire>`, `tag=recipe:water+fire` |
| Per-app namespace (avoid \"recipe:\" collisions across apps) | Server stamps the verified app; query `?app=<keyId>` reads from the per-app prefix |
| Typed relationship (\"edit\" vs \"combine\") | Tag with `kind:edit` or just use the parent tag and let the app infer |
| Multiple BYOP keys per app | Server stamps the verified `byopClientKeyId`, app filters its own gallery |

## What this PR *isn't*

- **Not a privacy system.** If you need bytes to be unreachable, content-addressed storage isn't the tool. Don't share the hash.
- **Not a lineage graph.** No first-class parent/child, no recursive ancestry queries. Apps that want lineage encode it as `parent:<hash>` tags; the catalog treats those like any other tag.
- **Not opinionated about app workflows.** No `relationship` enum, no `visibility` enum, no `source` enum. Apps own their semantics.

## CatGPT integration as proof

One app wired in to show the convention works: every meme is tagged `catgpt` at upload (so it appears in /tags/catgpt — the community gallery); when the meme was generated from a user portrait, it's *also* tagged `parent:<hash>` so the portrait's remixes are discoverable via `/tags/parent:<hash>`. No first-class lineage needed.

The other apps from prior PRs (voice-edit, pollicraft, ai-dungeon, etc.) port cleanly using this same convention. They don't need any new catalog fields — they just emit the tags they care about.

## /account/key additions (server-attested)

- `userId` (nullable)
- `keyId`
- `byopClientKeyId` (nullable)
- `byopClientName` (nullable)
- `byopClientUserId` (nullable)

## Tests — 14/14 pass

- baseline upload/retrieve/dedupe (6)
- untagged upload → in /me/media, not in any /tags
- tagged upload → in /tags/:tag (opt-in)
- **adversarial**: request-supplied `owner`, `app`, `byopClientKeyId` cannot override the verified identity
- per-app tag namespace isolates `recipe:water+fire` across CatGPT and Voice Edit
- `/me/media` scoped to caller only (Alice cannot see Bob's uploads)
- invalid tag rejected by `/tags/:tag`
- malformed tags silently dropped at upload
- **lineage-via-tag convention** works end-to-end (parent → child via `tag=parent:<hash>`)

## Compared to siblings

- #11306–#11310: each added some combination of `visibility` / `parents[]` / `relationship`. This PR drops all three. Net change vs. main is much smaller.
- #11311–#11313 (ensembles): each ~4500–4900 LoC across 16–18 files. This PR is ~780 LoC across 6 files.

### DIFF
--- a/apps/catgpt/ai.js
+++ b/apps/catgpt/ai.js
@@ -183,3 +183,59 @@ export async function handleImageUpload(file, notify) {
         }
     }
 }
+
+// Re-upload a generated meme blob into the media catalog. Tagging it with
+// `catgpt` opts it in to the public /tags/catgpt listing. The returned URL
+// is keyless (unlike the gen URL which embeds the user's key), so it's
+// safe to render in galleries and to persist in localStorage.
+//
+// Lineage convention (no first-class lineage in the catalog): if this meme
+// was generated from a user-uploaded portrait, we also stamp
+// `parent:<hash>`. The community can list a portrait's remixes via
+// /tags/parent:<hash>.
+export async function uploadGeneratedMeme(blob, { prompt, parentHash } = {}) {
+    const key = getStoredApiKey();
+    if (!key) return null;
+    try {
+        const form = new FormData();
+        form.append(
+            "file",
+            new File([blob], `catgpt-${Date.now()}.png`, { type: blob.type }),
+        );
+        form.append("tag", "catgpt");
+        if (parentHash) form.append("tag", `parent:${parentHash}`);
+        if (prompt) form.append("prompt", prompt);
+        const res = await fetch(MEDIA_UPLOAD, {
+            method: "POST",
+            headers: { Authorization: `Bearer ${key}` },
+            body: form,
+        });
+        if (!res.ok) return null;
+        return (await res.json()).url || null;
+    } catch (err) {
+        console.warn("catalog upload failed:", err);
+        return null;
+    }
+}
+
+const MEDIA_HASH_RE = /^https:\/\/media\.pollinations\.ai\/([a-f0-9]{16})$/i;
+export function extractMediaHash(url) {
+    if (!url) return null;
+    const m = MEDIA_HASH_RE.exec(url);
+    return m ? m[1].toLowerCase() : null;
+}
+
+// Public community gallery — anyone, no auth needed. Items here are opt-in:
+// every CatGPT meme is tagged `catgpt` at upload time.
+export async function fetchCatgptGallery(limit = 12) {
+    try {
+        const res = await fetch(
+            `https://media.pollinations.ai/tags/catgpt?limit=${limit}`,
+        );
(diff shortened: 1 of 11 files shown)


voodoohop:
Matches the media-catalog direction from the Jun 17 1:1:

- Premise 3 server-stamps `owner` and `app` from the verified `/account/key` — metadata records who uploaded and from which app, never request params.
- `GET /me/media` already exposes the owner's own uploads; `GET /tags/:tag?app=<keyId>` covers the per-app slice.

Forward note:
- The obvious next read is per-owner ("show all media from user X"). The opt-in tag model here is the right base for that.

No new scope for this PR. Server-attested owner/app plus opt-in tags is the correct foundation.

voodoohop:
Closing — superseded by #12277. The core idea here carried over: tags gate public visibility, untagged uploads stay private.

PR #11380, opened 2026-05-29
Restrict BYOP redirect URI schemes

its3li:
## Summary
- restrict BYOP redirect URIs to `https://` by default, with `http://` allowed only for loopback hosts
- apply the same redirect-scheme policy in allowlist matching, app-key creation, and metadata updates
- add regression coverage for app creation, metadata updates, stale stored metadata, app lookup, and secret-key minting

## Testing
- `XDG_CONFIG_HOME=F:\tmp\xdg-config npx vitest run test/url-utils.test.ts test/integration/api-keys.test.ts test/integration/account-keys.test.ts`
- `XDG_CONFIG_HOME=F:\tmp\xdg-config npm run typecheck`

##  TODO before merge

- [ ] **CI green** — all four checks currently `queued`, wait for them to complete.
- [ ] **Migration audit (prod D1).** Confirm no existing publishable keys have registered redirect URIs that this PR will retroactively reject. Run against prod:
  ```sql
  SELECT id, name, metadata
  FROM apikey
  WHERE metadata LIKE '%redirectUris%'
    AND metadata NOT LIKE '%"https://%'
    AND metadata NOT LIKE '%"http://127.0.0.1%'
    AND metadata NOT LIKE '%"http://localhost%'
    AND metadata NOT LIKE '%"http://[::1]%';
```

> If 0 rows: safe to merge as-is. If >0 rows: list owners and decide — notify + grace period, grandfather via flag, or force-update.
- [ ]  Custom URL schemes. Confirm no registered native app relies on a private-use scheme (com.example.app://). RFC 8252 §7.1 allows these; this PR rejects them. Same query as above, looking for redirectUris values that don't start with http.


voodoohop:
- Reworked to **registration-only** scheme enforcement (per discussion):
  - Removed the https/loopback re-checks from the auth-flow matcher (incoming URI + stored entries) and from the sk_ mint path — the allowlist match is the only gate the auth flow applies.
  - Kept the https-only/loopback gate at registration (create, metadata update, Zod schema) — the only place a bad scheme can enter an app keys allowlist.
  - Non-registered-app redirects (no `client_id`) keep working with any scheme (e.g. `file://`), unchanged from main.
  - Dropped the `legacy-unsafe-app` test — no fallback handling for old/force-written data.
- This also makes the merge TODOs moot: nothing retroactively rejects already-registered URIs, so no prod-D1 migration audit needed.
- Tests: url-utils + api-keys + account-keys all green (59 passing); typecheck + biome clean.
- Pre-existing double-validation on the metadata-update route tracked separately in #11451 (out of scope here).

PR #11392, opened 2026-05-29
Feature/sse balance updates 

Amine-SGM:
Implemented real-time balance updates using Server-Sent Events (SSE). The backend now streams updates every 30 seconds if the user's balance changes, and the frontend reacts accordingly, removing the need for polling or explicit router invalidations on balance changes.


voodoohop:
@Amine-SGM the PR is touching some unrelated files. I'm also wondering if balance needs to be streamed. Polling is cheap in this case and doing it every few seconds should be enough for displayed balances. I'm closing this for now, but feel free to comment and we can circle back.

PR #11426, opened 2026-05-30
Fix quest payout idempotency

its3li:
## Summary
- move quest payout idempotency from quest/PR/user/role scope to quest/user/role scope
- add a shared payout-key helper so the behavior is importable and testable
- guard the CLI entrypoint so importing the helper path does not execute the command

## Tests
- `npx vitest run test/quest-grant-pollen.test.ts`


enter.pollinations.ai/src/tier-progression/shared/quest-payout-key.ts
```diff
@@ -0,0 +1,7 @@
+export function buildQuestPayoutKey(
```
voodoohop: id say this function is too small to be split into its own file. just inline it

voodoohop: i dont think it even needs to be a function its so short

enter.pollinations.ai/src/tier-progression/shared/quest-grant-pollen.ts
```diff
@@ -135,4 +136,9 @@ const grantCommand = command({
     },
 });
 
-run([grantCommand]);
+if (
```
voodoohop: this change seems unrelated to what the PR does. I don't think it's necessary as the file is only ever run as a script. id just keep the PR to changing the key. The thing I like seeing most as a reviewer in the age of AI-assisted code commits is minimalism. the less lines changed the better

PR #11431, opened 2026-05-31
fix(gen-text): strip reasoning_effort for non-reasoning grok via transform

ElliotEtag:
- Strips `reasoning_effort` only for `grok` / `grok-4-20-non-reasoning`; prod repro showed 500 with the param and 200 without it.
- Preserves `reasoning_effort` for reasoning Grok variants (`grok-large`, `grok-4.3`).
- Composes with `stripCacheControl` from #11512 so the newer Grok Azure fix is not regressed.
- Adds transform and model-wiring tests for the stripped and preserved paths.

Verification:
- `npm run test -- --run test/text/transforms/stripReasoningEffort.test.ts test/text/transforms/stripCacheControl.test.ts`
- `npx tsc --noEmit --project gen.pollinations.ai/tsconfig.json`

Addresses #11167
Addresses #11089


voodoohop:
Opened #11531 as an alternative implementation using the per-model `transform:` system (a `stripReasoningEffort` transform on the `grok` entry, mirroring `stripCacheControl`) instead of the handler-level registry-wide strip. Same fix, scoped to the one broken deployment; doesn't touch the Gemini path. Whichever direction fits — close the other.

voodoohop:
Reworked to use the per-model `transform:` system instead of the handler-level registry-wide strip (thomash):

- Reverted `handler.ts` back to main; added a `stripReasoningEffort` transform on the `grok` entry only — same pattern as `stripCacheControl`.
- Scoped to the broken deployment: `grok-large` / `grok-4.3` (`reasoning: true`) keep accepting `reasoning_effort`; Gemini path untouched (the old approach keyed on `reasoning: true` registry-wide, which also stripped Gemini's user value and relied on the thinking transform re-injecting it).
- Coverage is identical — all grok traffic runs through `generateTextPortkey` where transforms apply.
- Swapped the test to `stripReasoningEffort.test.ts` (unit + wiring). 6/6 pass, biome + tsc clean.

Net diff is now 3 files, +78/−0.

PR #11440, opened 2026-05-31
Harden remote media downloads

its3li:
## Summary
- add a shared remote-media helper for URL validation, redirect rejection, content-length checks, and bounded reads
- apply the helper to embedding media fetches and image reference downloads
- add regression coverage for private URL blocking and redirect-disabled image fetches
- update Seedream tests to mock a realistic image content type for reference downloads

## Tests
- direct TS runtime check of `downloadUserImage` private URL blocking and `redirect: manual` fetch behavior

Not run locally: full `gen.pollinations.ai` worker Vitest targets are currently blocked on this machine because `marked` is declared but missing from `node_modules`, and reinstalling dependencies failed with `ENOSPC`.


voodoohop (commented):
- The size/streaming hardening is useful.
- Concern: rejecting every 3xx breaks valid media URLs behind CDN, http-to-https, and presigned redirect flows.
- Please follow a small bounded number of redirects and re-validate the final response instead of blanket rejecting redirects.

PR #11443, opened 2026-05-31
Cap realtime session spend

its3li:
## Summary
- derive a realtime session spend limit from the caller's usable balance and finite API key budget
- clamp realtime response output token settings before forwarding client events upstream
- close and settle sessions when observed usage reaches the authorized spend limit, while capping the settled charge to that limit
- add a regression test for a low-budget realtime session that previously could overdraw both user balance and key budget

## Tests
- npm test -- --run test/realtime.test.ts
- npm run typecheck --workspace pollinations-gen
- npx biome check gen.pollinations.ai/src/routes/realtime.ts gen.pollinations.ai/test/realtime.test.ts


voodoohop (commented):
@its3li although i think the idea is good, i wouldnt implement it this way. our billing logic lives in another part of the codebase. it's acceptable that balances go negative for the moment. so i'd ignore it for now. if we see that user's balances go very negative i'd revisit it.

PR #11348, opened 2026-05-28
feat: pollen-first pricing + localized currency on the top-up slider

ElliotEtag:
Implements #11346: show pack prices in Pollen and flip the displayed price to the buyer's local currency.

- Slider now leads with the Pollen amount; a per-buyer local price (e.g. `≈ €9.30`) appears below it, derived from Stripe FX Quotes mid-market rates with the country from `CF-IPCountry`. Unmapped countries fall back to the USD price. Checkout stays USD-native + Adaptive Pricing, so the slider price is a labelled `≈` estimate.
- Learns the real Adaptive Pricing markup per currency from completed purchases (EMA in KV), recorded off the credit path via `waitUntil`; 4% cold-start default until the first sale in a currency. Mid-market rate cached hourly.
- Dev-only `?country=` override to preview localized pricing on localhost (ignored in production).
- Pollen-frames the remaining stray `$` copy in `README.md` and the polli-cli skill.

Out of scope (separate PR): pack discount pricing and field/logic renames. The bonus pricing model is unchanged.

Checks: `npx biome check`, `npx vitest run test/fx-quotes.test.ts test/pollen-packs.test.ts test/integration/stripe.test.ts` (110 pass).

Fixes #11346

PR #11445, opened 2026-05-31
Simplify pollen packs: USD display + round amounts

voodoohop:
Small follow-up on top of #11348. Two simplifications to the pollen pack UX:

- **USD-only display**: drop the local-currency estimate plumbing (FX Quotes, country map, KV cache, `/localized-prices`). Slider shows exact USD; Stripe Adaptive Pricing still localizes the real charge at checkout. Research found no measured conversion lift from a pre-redirect estimate, and it risked drifting from the actual charge.
- **Round pollen via discount model**: grant = nominal USD (2/5/10/20/50/100) instead of bonus grants (6/13/28…); the discount lives in the charged price (`priceCents`, ladder 0/15/20/25/30/35%). Checkout + auto top-up both charge and verify the discounted price.

Slider now reads: pick pollen → see discounted USD + "% off". Stripe shows local currency at checkout.

Verified: biome clean, `tsc --noEmit` clean, 66 tests pass.

Refs #11348


voodoohop:
- Current checkout already follows this USD-native plus Stripe Adaptive Pricing direction. Closing the stale follow-up.

PR #11499, opened 2026-06-01
fix: preflight duration-based media charges

its3li:
## Summary

- Use requested duration as a preflight price floor for duration-billed media models
- Apply the floor to both user balance and finite API key budget checks
- Cover GET media generation plus OpenAI-compatible audio speech requests

## Why

The existing generation gate used the model historical average price before provider work starts. Duration-billed requests can be much more expensive than that average, for example nova-reel with duration=120 bills 120 * 0.08 = 9.6 pollen. This made a low balance or finite API key budget pass preflight and go negative after final billing.

This keeps the historical average as the fallback, but raises the required preflight amount when the current request has a duration-based price.

## Test plan

- npx vitest run src/byop-markup.test.ts
- npm run typecheck
- npx biome check gen.pollinations.ai/src/utils/generation-access.ts gen.pollinations.ai/src/routes/audio.ts gen.pollinations.ai/src/byop-markup.test.ts

### DIFF
--- a/gen.pollinations.ai/src/byop-markup.test.ts
+++ b/gen.pollinations.ai/src/byop-markup.test.ts
@@ -14,7 +14,10 @@ import {
 import { sql } from "drizzle-orm";
 import { drizzle } from "drizzle-orm/d1";
 import { describe, expect, it } from "vitest";
-import { checkBalance } from "@/utils/generation-access.ts";
+import {
+    checkBalance,
+    getDurationBasedEstimatedPrice,
+} from "@/utils/generation-access.ts";
 
 const db = drizzle(env.DB);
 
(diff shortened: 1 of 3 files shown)


voodoohop (commented):
- Legit fix: preflight should account for requested duration so long media cannot pass with a near-empty balance.
- Please resolve the current merge conflict.
- Before merge, avoid forking Nova Reel duration rounding; export/reuse the production duration helper so preflight and charge use the same seconds calculation.

voodoohop:
- Closing for now as a product decision.
- Negative balances are allowed as final accounting: many requests, especially text, cannot be priced exactly before upstream work completes.
- Once a balance is negative, further requests should be blocked; this just avoids conservative preflight estimates rejecting work before the real cost is known.

PR #11515, opened 2026-06-01
fix: block native api key redirect metadata

its3li:
Summary: Blocks native Better Auth api-key create/update routes from accepting redirect metadata directly. This keeps BYOP redirect allowlists on the Pollinations API-key routes where scheme and loopback validation runs. Testing: full api-keys integration file passed, targeted native redirect metadata test passed, git diff check passed. Note: local typecheck still fails on existing frontend UI package resolution and React DTS issues before this change.


voodoohop:
Thanks @its3li 🙏 Pushed a follow-up that simplifies this: instead of filtering specific redirect metadata fields, it just blocks the native `/api/auth/api-key/create` + `/update` routes entirely (405). Nothing legit uses them over HTTP — the frontend and our own helper go through `/api/api-keys` — so it closes the same `javascript:` bypass with ~27 fewer lines and no field-allowlist to maintain. Merging.

its3li:
> Thanks @its3li 🙏 Pushed a follow-up that simplifies this: instead of filtering specific redirect metadata fields, it just blocks the native `/api/auth/api-key/create` + `/update` routes entirely (405). Nothing legit uses them over HTTP — the frontend and our own helper go through `/api/api-keys` — so it closes the same `javascript:` bypass with ~27 fewer lines and no field-allowlist to maintain. Merging.

Thanks, makes sense. Blocking the native Better Auth API key create/update routes directly is cleaner than maintaining a redirect metadata allowlist, especially if all legitimate key creation flows go through /api/api-keys.

PR #11522, opened 2026-06-01
fix: prevent auto top-up webhook double credit

its3li:
## Summary
- Prevent concurrent duplicate Stripe auto top-up invoice deliveries from crediting pack balance twice.
- Make the credit update depend on the attempt still being pending/failed, then mark the attempt paid in the same D1 batch.
- Add a focused regression for two concurrent `creditAutoTopUpInvoice` calls for the same paid invoice.

## Context
This is separate from the older checkout-session idempotency fix in #10624. Auto top-up invoices use `stripe_auto_top_up_attempt`; the previous code could let two concurrent handlers both read `pending`, then the second handler could still increment `pack_balance` when the first handler set `status = paid` with the same millisecond `completed_at` marker.

## Checks
- `npx vitest run test/integration/stripe.test.ts -t "creditAutoTopUpInvoice credits concurrent duplicate invoices once"`
- `git diff --check`
- `npx biome check --formatter-enabled=false enter.pollinations.ai/src/utils/stripe-billing.ts enter.pollinations.ai/test/integration/stripe.test.ts`

Notes:
- `npx tsc --noEmit` still fails on existing frontend `@pollinations/ui` resolution / implicit-any errors outside this change.
- Full Stripe webhook-route tests are blocked in this local checkout because `npm run decrypt-vars` fails with missing `sops`, so `STRIPE_WEBHOOK_SECRET` is not available.

### DIFF
--- a/enter.pollinations.ai/src/utils/stripe-billing.ts
+++ b/enter.pollinations.ai/src/utils/stripe-billing.ts
@@ -601,23 +601,7 @@ export async function creditAutoTopUpInvoice(
     }
 
     const now = Date.now();
-    const [attemptUpdate] = await env.DB.batch([
-        env.DB.prepare(
-            `UPDATE stripe_auto_top_up_attempt
-                SET status = ?,
-                    completed_at = ?,
-                    updated_at = ?,
-                    failure_reason = NULL
-                WHERE stripe_invoice_id = ?
-                    AND status IN (?, ?)`,
-        ).bind(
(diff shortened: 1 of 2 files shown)


voodoohop:
- Closing for now as too marginal for the extra churn.
- The current webhook path already has an idempotency guard around the auto top-up attempt status/completed_at.
- A duplicate credit would require duplicate handlers to race in the same millisecond, so this is hardening rather than a necessary fix.

PR #11536, opened 2026-06-01
fix: make Polly API GitHub issues read-only

its3li:
## Summary

- Block `github_issue` write actions when Polly is running in HTTP API mode
- Update the API-mode prompt/tool text to describe GitHub issue access as read-only

## Why

Polly API mode forwards the caller's Pollinations API key for model usage, but GitHub tools still execute with Polly's GitHub credentials. API users should not be able to make Polly create/edit/delete/comment on GitHub issues through that bot identity.

This is separate from #11534: that PR removes `github_custom` raw GitHub API access from API mode; this PR blocks the remaining `github_issue` write path.

## Test plan

- `python -m py_compile apps/polly/src/services/github.py apps/polly/src/constants.py`
- `git diff --check`

Could not run `ruff` or an import-level handler smoke test locally because this Python environment is missing Polly dev dependencies (`ruff`, `ada_url`).


voodoohop:
Thanks @its3li, but closing this — letting API users instruct Polly to create issues/comments is intended, so making it read-only removes a feature we want.

The abuse case is also already contained: the target repo is hardcoded (`config.py` `default_repo`), API mode runs `is_admin=False` with admin actions (close/edit/label/assign) already blocked, and issues are attributed. What's genuinely missing is rate-limiting / cost / stronger per-key attribution — that's the right follow-up if abuse shows up, not blocking writes entirely.

its3li:
> Thanks @its3li, but closing this — letting API users instruct Polly to create issues/comments is intended, so making it read-only removes a feature we want.
> 
> The abuse case is also already contained: the target repo is hardcoded (`config.py` `default_repo`), API mode runs `is_admin=False` with admin actions (close/edit/label/assign) already blocked, and issues are attributed. What's genuinely missing is rate-limiting / cost / stronger per-key attribution — that's the right follow-up if abuse shows up, not blocking writes entirely.

Understood. ThanKS

PR #11538, opened 2026-06-01
fix(media): force active content downloads

its3li:
## Summary
- keep media uploads working, but force scriptable document types like HTML/SVG/XML to download instead of rendering inline
- add nosniff on served media and CSP sandbox headers for active document responses
- cover GET and HEAD responses with a regression test

## Why
media.pollinations.ai is documented as image/audio/video storage, but uploaded active document types were served inline from the public Pollinations media origin.

## Test plan
- npm --prefix media.pollinations.ai test
- npx tsc --noEmit -p media.pollinations.ai/tsconfig.json


voodoohop:
- Closing for now as a product decision, not because the hardening is wrong.
- Forcing active content to download would remove the useful small-HTML hosting behavior on media.pollinations.ai.
- If we later lock media down to assets only, we should do that explicitly, ideally with a separate hosting surface for HTML.

PR #11541, opened 2026-06-01
fix: preflight transcription duration billing

its3li:
## Summary
- require transcription requests to reserve the maximum accepted transcription duration before calling the upstream provider
- enforce the same maximum against provider-reported transcription duration
- add regression coverage for finite API-key budgets, paid-only pack balance checks, and over-limit provider durations

## Why
The transcription route previously used the normal generation preflight before the uploaded audio duration was known. Actual billing is based on provider-reported audio seconds after the upstream transcription work has already completed, so a low-balance or finite-budget API key could pass preflight and then be charged more than it was allowed to spend.

This is separate from #11499: that PR covers request-provided duration for media/audio generation paths. This route learns duration only after the transcription provider responds.

## Tests
- `cd gen.pollinations.ai && npm run typecheck`
- `cd gen.pollinations.ai && npx vitest run test/transcription-param-parsing.test.ts`

### DIFF
--- a/gen.pollinations.ai/src/routes/audio.ts
+++ b/gen.pollinations.ai/src/routes/audio.ts
@@ -1,12 +1,17 @@
 import type { Logger } from "@logtape/logtape";
+import { createBalanceCheckResult } from "@shared/billing/balance.ts";
+import { canCoverEstimatedCharge } from "@shared/billing/bucket-selection.ts";
 import {
     type AudioModelName,
     ELEVENLABS_VOICES,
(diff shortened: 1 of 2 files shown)


voodoohop:
- Closing for now as a product decision.
- Negative balances are allowed as final accounting: many requests, especially text, cannot be priced exactly before upstream work completes.
- Once a balance is negative, further requests should be blocked; this just avoids conservative preflight estimates rejecting work before the real cost is known.

PR #11542, opened 2026-06-01
fix: enforce api key budget deductions atomically

its3li:
## Summary
- require finite API key budgets to have enough remaining pollen at deduction time
- keep unlimited API key budgets (`NULL`) untouched
- add regression coverage for budget exhaustion during atomic deduction

## Why
API key budgets are checked before generation starts, but final billing is deducted after the response. The deduction query only required `pollen_balance IS NOT NULL`, so a finite-budget key could be decremented below zero if the remaining budget was already spent by another request between preflight and final billing.

Adding the remaining-budget predicate makes the spend limit atomic at the ledger update boundary instead of relying only on the earlier preflight snapshot.

## Tests
- `cd gen.pollinations.ai && npm run typecheck`
- `cd gen.pollinations.ai && npx vitest run test/billing-deduction.test.ts`

### DIFF
--- a/gen.pollinations.ai/test/billing-deduction.test.ts
+++ b/gen.pollinations.ai/test/billing-deduction.test.ts
@@ -1,9 +1,16 @@
 import { env } from "cloudflare:test";
 import { getUserBalance } from "@shared/billing/balance.ts";
-import { atomicDeductUserBalance } from "@shared/billing/deduction.ts";
+import {
+    atomicDeductApiKeyBalance,
+    atomicDeductUserBalance,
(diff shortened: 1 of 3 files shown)


voodoohop:
- Holding off on this one.
- Current negative budget behavior is more honest accounting: the key spent past its cap, then future requests stop.
- This PR prevents the negative ledger entry, but because billing happens after the response, the race loser may still get a response and then fail settlement, creating possible free usage.
- Better path: keep internal ledger truthful and clamp/display exhausted budgets as 0, or design a real pre-generation reservation system if we want hard caps.

PR #11546, opened 2026-06-01
fix: require profile scope for device userinfo

its3li:
## Summary
- require `account:profile` when `/api/device/userinfo` is called with an API key
- keep session-cookie access unchanged for the browser/device approval flow
- add regression tests for unscoped and profile-scoped API keys

## Security impact
Medium. `/api/account/profile` already hides name/email unless an API key has `account:profile`, but `/api/device/userinfo` returned OIDC-style profile data, including email, for any valid API key. This keeps userinfo aligned with the existing account profile permission boundary.

## Testing
- `npx vitest run test/device-auth.test.ts`

Note: `npm test --workspace pollinations-enter -- device-auth.test.ts` could not run in this local environment because `sops` is not installed; it fails before Vitest during `decrypt-vars`.


voodoohop:
- Resolved by #12165: device/OAuth userinfo now gates name and email behind `account:profile`. Closing this older branch.

PR #11562, opened 2026-06-02
add minimax m3, move gemma 4 26b to openrouter

ElliotEtag:
- Adds **MiniMax M3** (`minimax-m3`, aliases `minimax3`/`minimax-3`) via OpenRouter — text+image input, tools, reasoning, 1M context. Verified upstream: tool calls, image input, reasoning tokens, prompt caching all work.
- M3 priced at 1x markup using the current OpenRouter rate ($0.30/M in, $1.20/M out, $0.06/M cache). This is a temporary 50% promo; base is $0.60/$2.40 — flagged in a code comment to revisit when the promo ends.
- Moves **Gemma 4 26B-A4B** from DeepInfra → OpenRouter (same SKU, cheaper: $0.06/$0.33 vs $0.07/$0.34), keeps 1x markup, non-paid.
- Checked AWS Bedrock + Fireworks first: M3 isn't on either (Bedrock tops out at M2.5); Gemma 4 is OpenRouter-only (Bedrock has Gemma 3, Fireworks is on-demand only). **Voyage 4 dropped** — not served by Fireworks, Bedrock, or OpenRouter (would need Voyage's native API).
- Tests: aliases, pricing-data, usage-headers, model-permissions pass; effective-prices snapshot updated. Local gateway lists both with correct pricing/modalities. Live local generation not run (no seeded local key in this clone); upstream responses verified directly.

### DIFF
--- a/enter.pollinations.ai/test/__snapshots__/effective-prices.snapshot.test.ts.snap
+++ b/enter.pollinations.ai/test/__snapshots__/effective-prices.snapshot.test.ts.snap
@@ -270,12 +270,12 @@ exports[`effective cost and price per model — snapshot 1`] = `
   },
   "gemma": {
     "cost": {
-      "completionTextTokens": 3.4e-7,
-      "promptTextTokens": 7e-8,
+      "completionTextTokens": 3.3e-7,
+      "promptTextTokens": 6e-8,
     },
     "price": {
-      "completionTextTokens": 3.4e-7,
-      "promptTextTokens": 7e-8,
+      "completionTextTokens": 3.3e-7,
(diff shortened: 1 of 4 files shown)


ElliotEtag:
Live local end-to-end now verified through the gateway (the auth gap is fixed in #11564 via `npm run seed:local`):
- `minimax-m3` → real generation on localhost:8788 (finish=stop, usage incl. cached + reasoning tokens)
- `gemma` → routes via OpenRouter `google/gemma-4-26b-a4b-it`, correct x-usage headers
- bogus token → 401 (auth enforced)

PR #11609, opened 2026-06-04
feat: route gemini to Airforce with Vertex fallback

voodoohop:
- Routes public `gemini` (gemini-3-flash) to api.airforce, falling back to existing GCP/Vertex on failure
- Same modelId on both targets → billing/pricing unchanged regardless of which serves
- Restores the Portkey `strategy`/`targets` fallback in `portkeyUtils.ts` (revert of `04894e386`, shipped in prod Jan–Jun 2026)
- Drops the `code_execution` default-injection from the gemini transform: Airforce returns a phantom tool_call for it; explicit code_execution requests fall back to Vertex via `on_status_codes` 400
- `on_status_codes` includes 402 (Airforce out of balance) — the real production fallback trigger; failed attempts aren't billed, so falling back is always safe
- Adds `AIRFORCE_API_KEY` to gen secrets (dev/staging/prod) + a unit test for the fallback header shape

Validated end-to-end against the live Portkey gateway: happy path serves Airforce, forced failure falls through to Vertex.


sharktide (approved):
i think it looks good, if you've tested it and it works fine then i think we can go for it

PR #11640, opened 2026-06-05
fix: harden tier update D1 execution

its3li:
## Summary
- pass wrangler D1 commands through execFileSync argv instead of a shell-built command string
- SQL-quote tier update string values consistently before sending them to D1

## Verification
- npx biome check --write enter.pollinations.ai/src/tier-progression/shared/tier-update-user.ts
- npx tsc --noEmit --pretty false --skipLibCheck src/tier-progression/shared/tier-update-user.ts
- npx tsx src/tier-progression/shared/tier-update-user.ts --help


ElliotEtag:
Solid hardening on its own — `execFileSync` + SQL-quoting is the right call. But we're deprecating user tiers: #11928 deletes `tier-update-user.ts` entirely (along with the drip and tier-progression flows). So we don't want to land changes to this surface. Closing in favor of the tier removal — thanks @its3li.

PR #11682, opened 2026-06-08
billing design 01: keep calculateCost with catalog notes

voodoohop:
- Keeps the current calculateCost hook design.
- Adds Gemini 3 search billing to gemini-search-fast and gemini-search-large.
- Adds pricing_notes to model-info so catalogs can explain dynamic fees.

Validation:
- npx biome check --write shared/registry/registry.ts shared/registry/model-info.ts shared/registry/text.ts enter.pollinations.ai/test/pricing-data.test.ts
- npx vitest run test/pricing-data.test.ts test/aliases.test.ts

### DIFF
--- a/enter.pollinations.ai/test/pricing-data.test.ts
+++ b/enter.pollinations.ai/test/pricing-data.test.ts
@@ -215,6 +215,48 @@ test("Gemini dynamic billing notes are exposed in model catalog metadata", () =>
     );
 });
 
+test("Perplexity request search fees are added by the shared cost path", () => {
+    const usage = {
+        promptTextTokens: 1_000_000,
+        completionTextTokens: 1_000_000,
+    };
+    const cases = [
+        ["perplexity-fast", 2.005],
+        ["perplexity-deep", 2.012],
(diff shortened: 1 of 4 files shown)


ElliotEtag:
**Design 01 of 3 — verdict.** Full cross-PR comparison: https://github.com/pollinations/pollinations/pull/11683#issuecomment-4660911647

Ranking: best **ship-now / YAGNI** option, but not the long-term base.

- Smallest change; cleanly retires the Perplexity `priceMultiplier` fee hack.
- `pricing_notes` is free text that isn't derived from the billed value → it will drift from the actual charge.
- Metadata is asymmetric: Perplexity fee is structured/numeric, Gemini fee is prose only — weak as a public contract.

The one thing I'd change: `pricing_notes` ships publicly (full `ModelInfo` is serialized on `/models`), so if this merges as a fix, omit the field until there's a real consumer.

Verified locally: passes 212/212 focused tests (`pricing-data` + `aliases`).

voodoohop:
Closing in favor of #11683 — the declarative design was selected after comparing all three approaches.

PR #11683, opened 2026-06-08
billing design 03: declarative tiers and adjustments

voodoohop:
Declarative search-fee billing for Gemini and Perplexity. Supersedes #11227 (and design drafts #11682, #11684).

- Replaces Gemini custom cost calculators with declarative `billing: { tiers, adjustments }` registry rules — Gemini 2.5: $35/1k grounded prompts, Gemini 3.x: $14/1k search queries, long-context tier for gemini-large
- Adds Perplexity per-request search fees ($5/$12/$14/$14 per 1k by model) — closes ~$660/month under-billing (30-day Tinybird data)
- Bills provider-reported `usage.cost.request_cost` when present (streams today); malformed cost data fails loudly (502 on non-stream) instead of billing a fallback; absent data bills the static fee
- Sanitizes non-stream public usage — no `usage.cost`/`search_context_size` leak to clients
- Fix: moderation telemetry kept with tracking overrides (`x-moderation-*` event fields were dropped for non-stream chat)
- Always use registry-selected Perplexity `search_context_size` (caller override was unreachable dead code)

Validation:
- enter: `npx vitest run test/pricing-data.test.ts test/aliases.test.ts` — 217 passed
- gen: `npx vitest run test/generation-vcr.test.ts` — 16 passed (incl. streaming provider-cost + moderation e2e), `npm run typecheck` clean
- biome clean on changed files

Deploy: safe before pollinations/gateway#7 (static fee covers non-stream). Pin bump tracked in #11703.


ElliotEtag:
> Comparison of the three billing-design PRs (#11682 · #11684 · #11683). This PR (#11683, design 03) is my recommended base. Full write-up below; the other two PRs link here.

# Billing Design PR Comparison

Date: 2026-06-09

Scope:
- [#11682](https://github.com/pollinations/pollinations/pull/11682) - billing design 01: keep `calculateCost` with catalog notes
- [#11684](https://github.com/pollinations/pollinations/pull/11684) - billing design 02: documented billing policy functors
- [#11683](https://github.com/pollinations/pollinations/pull/11683) - billing design 03: declarative tiers and adjustments

## Short Answer

My ranking:

1. Best long-term base: **#11683**
2. Best immediate/YAGNI fallback: **a trimmed #11682**
3. Design I would avoid: **#11684**

The practical recommendation is:

- Do not merge #11684.
- Prefer a fourth option built from #11683: adopt the declarative billing engine internally, add a narrow escape hatch for non-declarative cases, and gate the public `billing` field until there is a real consumer and a stable schema.
- If the goal is only to bill Gemini search correctly now, use a simplified #11682-style change and avoid publishing a new metadata contract.

This separates two decisions that should stay independent:

- Internal billing representation: imperative / policy functor / declarative.
- Public `/models` metadata: nothing / prose / structured data.

The strongest path is not "choose #11683 and publish a contract now." It is "use #11683's drift-resistant internal model, but do not freeze a public API shape prematurely."

## Why This Is Mostly An Architecture Decision

All three PRs target the same billing problem and assert the same billing outputs:

- `gemini-search-fast`: `1.778`
- `gemini-search-large`: `10.528`
- `perplexity-fast`: `2.005`
- `perplexity-deep`: `2.012`
- `perplexity`: `18.014`
- `perplexity-reasoning`: `10.014`

These are asserted by the branch tests, not yet independently observed by running the suites locally. Before any merge decision, this parity should be confirmed on the actual branches.

So the main decision is not about the intended final numbers. It is about:

- how dynamic billing rules are represented,
- whether model metadata can drift away from actual billing behavior,
- and what public API shape Pollinations wants to expose.

The metadata point matters because `gen.pollinations.ai/src/routes/proxy.ts` returns full `ModelInfo` objects from the detailed model endpoints (`/models`, `/text/models`, etc.). That means new fields like `pricing_notes`, `billing_policy`, or `billing` become visible to clients immediately unless serialization explicitly omits them.

## Additional Pricing Policy Note

Every dynamically billed model touched by these PRs currently has `priceMultiplier: 1`:

- Gemini grounded/search models
- Perplexity models

That means adjustment-specific multiplier behavior in #11683 is a latent capability, not a present-day win. It is still useful, but it is not exercised by a real model today.

This hides a product/pricing question that should be answered explicitly before any adjusted model gets `priceMultiplier !== 1`:

- Should provider search/request fees be passed through at cost?
- Or should they be marked up like token usage?

#11683 has the mechanism to answer this via per-adjustment `priceMultiplier`, but the policy should be deliberate, not accidental.

## PR #11682: Keep `calculateCost` With Catalog Notes

This is the smallest change. It keeps the existing imperative `calculateCost` hook and adds:

- `searchFeePerThousandRequests` for Perplexity request fees
- `pricingNotes` / `pricing_notes` for catalog display

Strengths:

- Lowest blast radius.
- Easiest to review.
- Fixes the current Perplexity fee workaround cleanly.
- Fits the YAGNI rule if the immediate goal is just correct billing.

Concerns:

- It creates two billing metadata styles: structured numeric Perplexity fee, but Gemini fee described only in prose.
- `pricing_notes` can drift from the actual calculator because the note text is not derived from the billed value.
- Free-text billing notes are weak as a public API contract.

My take:

#11682 is the best ship-now option, but I would avoid exposing `pricing_notes` unless the team is comfortable supporting that public field.

## PR #11684: Billing Policy Functors

This wraps the existing dynamic calculators in named policy objects:

```ts
{
  id,
  description,
  adjustments,
  calculateCost
}
```

Strengths:

- More structured than #11682.
- Keeps arbitrary calculation flexibility.
- Provides policy IDs and adjustment metadata for catalogs.

Concerns:

- It has a dual source of truth: the real charge is still inside `calculateCost`, while `adjustments` are descriptive metadata.
- The registry auto-applies only `when: "always"` adjustments. Grounded/search adjustments still depend on the policy's custom function.
- That means metadata can claim one billing rule while the calculator charges another.
- Because this metadata is exposed publicly, drift becomes an API trust problem.

My take:

I would rank this lowest. It looks like a compromise, but it institutionalizes the exact billing drift risk the metadata is supposed to solve.

## PR #11683: Declarative Tiers And Adjustments

This replaces custom Gemini calculators with data-driven rules:

- `billing.tiers` for conditional cost definitions, such as long-context Gemini rates
- `billing.adjustments` for per-unit fees, such as search queries or request fees

The registry interprets those rules for both billing and metadata.

Strengths:

- Single source of truth internally: the same declarative data can drive billing and metadata.
- Better fit for machine-readable billing metadata.
- Easier to audit: model definitions declare fees as data.
- Adds the useful ungrounded Gemini search test.
- Allows adjustment-specific `priceMultiplier`, which could express "pass through provider fee at cost" versus "mark up like tokens."

Concerns:

- Largest change.
- Moves provider-specific Gemini output parsing into generic registry code.
- The public `count` values currently expose internal implementation names like `geminiWebSearchQueries`.
- A new metered unit still requires editing the central registry enum.
- The adjustment-specific multiplier path is not exercised by a real adjusted model today, because all dynamically billed models currently have `priceMultiplier: 1`.
- More architecture than strictly needed if the only goal is shipping the Gemini fee.

My take:

#11683 is the best architectural direction, but I would revise it before merge:

- Keep provider-specific parsing in provider-specific modules, not the generic registry.
- Gate or omit public `billing` serialization until there is a real consumer.
- If/when `billing` is exposed, map internal counters to public, stable terms before exposing them in `/models`.
- Add a synthetic test for `priceMultiplier !== 1` plus an adjustment.
- Decide whether provider search/request fees are passed through at cost or marked up.

## Fourth Option: #11683 Engine, Gated Public Contract

This is the strongest path if the team is willing to do a small design pass:

- Use #11683's declarative `billing.tiers` and `billing.adjustments` internally.
- Do not serialize `billing` into `ModelInfo` yet, or hide it behind an explicit gate, until a pricing page/docs/dashboard consumer exists.
- Keep provider-specific counters and output parsing outside generic registry code where possible.
- Add a narrow custom-calculator escape hatch for truly non-declarative billing cases.
- If the escape hatch is used, do not publish structured billing metadata for that model unless tests prove the metadata and computed charge match.

This keeps the main benefit of #11683: drift-proof internal billing. It also avoids freezing a public API schema before the consumer requirements are known.

The escape hatch recovers #11684's one genuine merit: arbitrary billing flexibility. The important constraint is that the escape hatch must not recreate #11684's dual-source-of-truth problem.

## Recommendation

Use this decision rule:

- If this is a billing hotfix: merge a trimmed #11682-style change, with no speculative public metadata.
- If this is a billing architecture decision: continue from #11683, but gate public metadata and add an escape hatch.
- Drop #11684.

My preferred path is the fourth option: #11683's declarative engine internally, no public `billing` contract yet, explicit fee-markup policy, and tests covering the multiplier edge case.

## Required Checks Before Merge

Before choosing or merging a final PR:

1. Run the focused tests on each branch:

```bash
npx vitest run test/pricing-data.test.ts test/aliases.test.ts
```

2. Treat passing tests as a precondition, not a follow-up. The current parity is asserted by test expectations; it still needs to be observed on the branches.

3. Diff `/models` and `/text/models` output for the three branches, and for the fourth-option variant if implemented.

4. Identify the actual metadata consumer:

- pricing page
- docs
- dashboard
- external clients

5. Pressure-test one next billing case:

- another provider with flat request fees,
- a model with both token tiers and request adjustments,
- or an image/video fee that does not fit Gemini/Perplexity assumptions.

6. For #11683 specifically, add a synthetic test covering `priceMultiplier !== 1` plus an adjustment, because no current adjusted model exercises that path.

7. Decide and document the pricing policy for provider fees:

- pass through at cost,
- mark up with the model multiplier,
- or choose per adjustment.


---

**Verified locally** — required checks #1–#2 are done, not pending: branches 01 / 02 / 03 each pass the focused suites (`pricing-data.test.ts` + `aliases.test.ts`) at **212/212**, so the billing parity above is observed, not just asserted. #11683's `calculatePrice` rewrite also passes the full existing price catalog with no regression (every adjusted model is `priceMultiplier: 1`).

PR #11684, opened 2026-06-08
billing design 02: documented billing policy functors

voodoohop:
- Replaces bare model-level calculateCost hooks with named billingPolicy objects.
- Exposes billing_policy id and description through model-info for pricing catalogs.
- Adds Gemini 3 search billing to gemini-search-fast and gemini-search-large.

Validation:
- npx biome check --write shared/registry/registry.ts shared/registry/gemini-billing.ts shared/registry/text.ts shared/registry/model-info.ts enter.pollinations.ai/test/pricing-data.test.ts
- npx vitest run test/pricing-data.test.ts test/aliases.test.ts

### DIFF
--- a/enter.pollinations.ai/test/pricing-data.test.ts
+++ b/enter.pollinations.ai/test/pricing-data.test.ts
@@ -177,6 +177,16 @@ test("Gemini grounding cost is added by family billing rules", () => {
         groundedOutput,
     );
     const gemini3Cost = calculateCost("gemini", usage, groundedOutput);
+    const geminiSearchFastCost = calculateCost(
+        "gemini-search-fast",
+        usage,
+        groundedOutput,
+    );
+    const geminiSearchLargeCost = calculateCost(
+        "gemini-search-large",
+        usage,
+        groundedOutput,
+    );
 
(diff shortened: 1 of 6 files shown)


ElliotEtag:
**Design 02 of 3 — verdict.** Full cross-PR comparison: https://github.com/pollinations/pollinations/pull/11683#issuecomment-4660911647

Ranking: the one I'd **avoid** (lowest of the three).

- Dual source of truth: the real charge stays inside the imperative `calculateCost`, while the published `adjustments` are descriptive only.
- The registry auto-applies only `when: "always"` adjustments; grounded/search fees still run through the custom function — so the metadata on `/models` can claim one rule while the calculator charges another.
- That institutionalizes the exact drift the structured metadata is meant to prevent, and it's public.

Better path: keep the good part (named policy `id` + `description`) and fold it into #11683's declarative model, which derives metadata from the same data that bills.

Verified locally: passes 212/212 focused tests (`pricing-data` + `aliases`).

voodoohop:
Closing in favor of #11683 — the declarative design was selected after comparing all three approaches.

PR #11686, opened 2026-06-08
feat: add Google social login

ElliotEtag:
## Summary
- Adds Google as a first-class Better Auth social provider alongside GitHub.
- Removes trusted social providers and enables encrypted OAuth token storage.
- Advertises enabled social providers from the backend so the UI only shows configured providers.
- Moves staging access from GitHub ID allowlist to verified-email allowlist for both providers.
- Adds callback/linking tests for verified same-email linking, unverified no-link, Google-only signup, provider advertisement, and staging gating.

## Notes
- The composite unique index and durable auth rate limiting remain follow-up items, not part of this PR.
- Local auth trusted origins remain on `localhost:3000`; no `3001` exception is included.

## Checks
- `npx vitest run test/staging-access.test.ts test/social-sign-in.test.ts`
- `npm run typecheck`


sharktide (approved):
LGTM I think we can merge. I'm not sure if it is configured on the backend side however and I didn't go in too deep in this

PR #11705, opened 2026-06-09
feat: route claude-opus-4-6 to myceli bedrock account

ElliotEtag:
- Adds per-account Bedrock credential selection (`default` vs `myceli`) via new `AWS_BEDROCK_MYCELI_*` env vars
- Routes `claude-opus-4-6` to the new Myceli prod Bedrock account (514585225061); it's the lowest-traffic activated model and fits the account's quota
- Sonnet/Haiku and Opus 4.7/4.8 stay on the default account (301235909293) until their quota is raised / activated there
- No fallback: missing Myceli creds fail loudly, so a misrouted model is visible rather than silently served from the old account
- Adds `wrangler.toml` test placeholders + binding types; ignores `.superpowers/`


ElliotEtag:
Closing at maintainer request. This stale Bedrock account-routing draft is no longer part of the active model plan.

PR #11770, opened 2026-06-13
Fix n go

stormdede515-eng:
## Summary

- System prompt anchor: model verifies claims against history before accepting them — prevents sycophantic collapse on false references
- Phrase detection (phraseDetector.js): scans outgoing messages for "you told me" / "you said" patterns, checks against conversation history, surfaces violet integrity note when claim is unverifiable
- Integrity hash store (integrityStore.js): djb2 hashes AI responses for within-session tamper detection
- Fault attribution on every context drop: user (wrong model), model (ignored capability), platform (pipeline loss)
- Fixes false positive drops: model-no-vision only fires when registry explicitly rejects images
- Dismiss buttons (×) on all note blocks
- Removes over-eager image-in-prior-turn advisory

Co-authored-by: stormdede515-eng <stormdede515@gmail.com>


ElliotEtag:
Thanks for the substantial effort here. After review, we're going to close this — not over code quality in the details, but because most of the surface area doesn't fit the app's direction, and the parts that don't fit are the bulk of the PR:

- **Integrity hash store** — `storeHash` is wired up but `verifyHash`/`clearHash` are never called, so the hashes are written and never read. The "tamper detection" has no effect as shipped.
- **Phrase detector** — surfacing a user-facing "could not be verified" note based on keyword overlap produces false positives on legitimate references (e.g. a user saying "you recommended Postgres" when the model said "PostgreSQL" gets flagged). We don't want to show users an accusation we can't stand behind.
- **System-prompt injection** — appending a fixed anchor to every conversation (and injecting a default persona when the user set none) is a product change, not a bug fix. That needs its own discussion before it ships to everyone.
- **`contextLogger._entries`** grows unbounded for the page lifetime (`flush()` is never called), and `received` entries are stored but never surfaced.

The one piece worth resubmitting on its own: the **localStorage quota-exceeded warning** in `storage.js`. Silent failure when image attachments overflow the quota is a real bug, and a small focused PR for just that would be welcome. The genuine dropped-attachment logging (no recoverable data) could ride along with it.

Happy to review a scoped follow-up.

PR #11817, opened 2026-06-15
[codex] fix: remove model list sorting controls

voodoohop:
- Removes clickable sorting controls from the Enter model list table headers.
- Keeps the model list on the existing default 1 pollen ordering.
- Validated with `npx biome check --write enter.pollinations.ai/frontend/src/components/models/model-table.tsx` and `npx tsc --noEmit -p frontend/src/tsconfig.json`.


sharktide:
why?

PR #11843, opened 2026-06-16
feat: log whether a Portkey fallback target served the request

voodoohop:
Records, per text request, whether Portkey served from a non-primary fallback target (e.g. Airforce→Vertex on the public `gemini` model). Today nothing captures this — `model_provider_used` is the static registry provider, and the gateway's `x-portkey-*` headers are discarded in `genericOpenAIClient` before tracking sees them.

## Approach (mirrors the existing `x-model-used` flow)
- `genericOpenAIClient` reads `x-portkey-last-used-option-index` off the gateway response (stream + non-stream) → `ChatCompletion.fallbackTarget`
- `handler.ts` re-emits it as the internal `x-fallback-target` worker header (in `usageHeaders` + explicitly in `sendTextStreamResponse`, which bypasses `usageHeaders`)
- `track.ts` parses `"config.targets[N]"` → `fallbackUsed` (true when N > 0), recorded before the cache/billing branch so streaming and non-streaming both work
- New `fallback_used` Bool on the `generation_event` datasource + `fallbackUsed` on the event schema

Verified against the live gateway: happy path returns `config.targets[0]` (`x-portkey-provider: openai`), forced fallback returns `config.targets[1]` (`vertex-ai`). The header value is the string `config.targets[N]`, not a bare int — parser handles that.

## Why not reuse `model_provider_used`
Semantically tempting, but it's consumed by ~45 sites (4 Grafana economics dashboards, alerting rules, a billing test) that assume the static registry provider. Notably `daily-spending.json` filters paid-inference by `model_provider_used IN ('google',...)`; flipping it to `airforce` would silently move billed Gemini out of the paid bucket. A dedicated bool is lower-risk and leaves all of that untouched.

## Query
```sql
SELECT countIf(fallback_used)/count() AS fallback_rate
FROM generation_event WHERE model_requested = 'gemini'
```

## Tests
- `genericOpenAIClient`: captures `fallbackTarget` on stream + non-stream; undefined when header absent
- `tracking-observability`: `fallbackUsed` true (`config.targets[1]`), false (`config.targets[0]`), false (absent)

## Tinybird deploy (manual, not in this PR)
`fallback_used` column ADD is non-destructive (backfills existing rows as `false`). Deploy to **both** workspaces, staging first: `tb --cloud deploy --wait` from `enter.pollinations.ai/observability` (override `TB_TOKEN` for staging). Validate with `--check` first.

### DIFF
--- a/enter.pollinations.ai/observability/datasources/generation_event.datasource
+++ b/enter.pollinations.ai/observability/datasources/generation_event.datasource
@@ -48,6 +48,7 @@ SCHEMA >
 `resolved_model_requested`  LowCardinality(String) `json:$.resolvedModelRequested` DEFAULT 'undefined',
 `model_used`                LowCardinality(String) `json:$.modelUsed` DEFAULT 'undefined',
 `model_provider_used`       LowCardinality(String) `json:$.modelProviderUsed` DEFAULT 'undefined',
+`fallback_used`             Bool                   `json:$.fallbackUsed` DEFAULT false,
 `is_billed_usage`           Bool                   `json:$.isBilledUsage`,
 
 -- Pricing
--- a/gen.pollinations.ai/src/middleware/track.ts
+++ b/gen.pollinations.ai/src/middleware/track.ts
@@ -33,6 +33,7 @@ import {
     type UsagePrice,
 } from "@shared/registry/registry.ts";
 import {
+    FALLBACK_TARGET_HEADER,
     openaiUsageToUsage,
     parseUsageHeaders,
 } from "@shared/registry/usage-headers.ts";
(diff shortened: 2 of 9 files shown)


voodoohop:
Good catch on the body leak — fixed in 26646e297.

- `fallbackTarget` was enumerable on the returned `ChatCompletion`, so `sendOpenAIResponse`'s `JSON.stringify({ ...completion })` leaked `"fallbackTarget":"config.targets[1]"` into the `/v1/chat/completions` body (OpenAI-incompatible).
- Fix: attach it as **non-enumerable** metadata at the source (`genericOpenAIClient`), so it never serializes into any `{ ...completion }` body while staying readable for the `x-fallback-target` header. Chose this over stripping at the serializer so the invariant lives at the attachment point and can't re-leak via a future serialization site.
- On your open question: the **header** is intentional/public (consistent with `x-model-used`, `x-usage-*`); only the **body** field was accidental.
- Added a regression test asserting `fallbackTarget` is non-enumerable and absent from the serialized body.

PR #11849, opened 2026-06-17
feat: add reward grants quest floor

voodoohop:
- Adds a source-agnostic D1 reward_grants table and shared grantReward() helper for idempotent balance credits.
- Adds /api/account/quests to return generic quest grants plus legacy quest_payout_credits rows.
- Dual-writes GitHub code quest payouts into reward_grants while preserving the existing quest_payout_credits guard.
- Adds focused coverage for idempotent grants and legacy quest history.

Validation:
- npm run decrypt-vars && npx vitest run test/reward-grants.test.ts
- npm run typecheck


voodoohop:
Superseded by #11856, which reconciles the atomic reward ledger, live account history, quest evaluator, catalog, and snapshot accounting into one replacement PR.

PR #11864, opened 2026-06-17
fix(community-models): format titles as name by @owner

fisventurous:
fixes the display name for community models where the description was overriding the title and preventing a clean "model by @user" format


fisventurous:
@voodoohop as reported [here](https://discord.com/channels/885844321461485618/1451926412771594312/1516689960768503898)

PR #11866, opened 2026-06-17
fix(community-models): Text Wrap Model Names in the UI

sharktide:
forces model names to text wrap in the model names list


sharktide:
oops wrong file

PR #11567, opened 2026-06-02
feat: add community text endpoints

voodoohop:
- Adds Enter dashboard/API registration for OpenAI-compatible community text endpoints.
- Adds the Drizzle/D1 `community_endpoint` table and stores bearer tokens encrypted.
- Exposes `community/{username}/{model-id}` IDs with description and per-1M-token Pollen pricing.
- Lists published community models for everyone in `/models`, `/text/models`, `/v1/models`, and the dashboard Models page.
- Routes community models through normal text call paths: `/v1/chat/completions`, `/text`, and `/text/{prompt}`.
- Supports community streaming, requests upstream usage, and passes upstream response model names through unchanged.
- Charges community usage through the normal deduction path using upstream-reported usage and dynamic model pricing.
- Keeps community runtime lookup in a small gen module; generic handlers only attach the resolved endpoint.
- Sorts dashboard community model prices from the dynamic model-list data instead of the static registry.
- Merges latest `origin/main` into the branch.

Validation:
- `git diff --check`
- `npx biome check --write enter.pollinations.ai/vite.config.ts`
- Enter: `npx vitest run test/community-endpoint-openai.test.ts test/pricing-data.test.ts`
- Enter: `npm run build:frontend`
- Enter: `npm run typecheck` still fails on broad dashboard/UI declaration required-prop errors unrelated to the merge resolution.
- Gen: `npx vitest run src/community-endpoints.test.ts test/model-permissions.test.ts`
- Gen: `npx tsc --noEmit`
- Local: `http://localhost:8788/v1/models` and `/text/models` list community models with normal text endpoints.

PR #11867, opened 2026-06-17
feat(community-endpoints): Text Wrap Model List Items

sharktide:
Text wraps model list items to allow for longer model names

cc @voodoohop


sharktide:
cc #11567

PR #11869, opened 2026-06-17
fix(gen-image): weight image LB by last latency

voodoohop:
- Routes image requests weighted by `1/lastMs` instead of uniform random, so a fast 4090 gets a larger share than a slow 3090 (which drowned at an equal split → 524 timeouts)
- Gen worker measures `/generate` latency itself and stores the last value on the registry entry; no worker/backend changes
- Latency writes throttled to 20s/server (KV ~1 write/sec/key); recorded on success only so timeouts don't poison the weight
- Unmeasured servers use a neutral default so they still get sampled
- Weighted-random (not always-pick-fastest) to avoid herding requests onto one server between updates
- Adds unit tests for `chooseWeightedServer` + `recordLatency`

### DIFF
--- a/gen.pollinations.ai/src/image/availableServers.ts
+++ b/gen.pollinations.ai/src/image/availableServers.ts
@@ -176,7 +176,7 @@ export function __resetLatencyStateForTests(): void {
     recentWrites.clear();
 }
 
-export const fetchFromLeastBusyServer = async (
+export const fetchFromWeightedServer = async (
     type: ServerType = "flux",
     options: RequestInit,
 ): Promise<Response> => {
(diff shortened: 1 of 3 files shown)


voodoohop:
Reviewed with 3 independent subagents (correctness, algorithm, simplicity). Verdict: algorithm sound (weighting by 1/lastMs gives the correct throughput-proportional split; weighted-random avoids herding). Fixed the real findings:

- **`await recordLatency`** instead of fire-and-forget — a floating promise can be cancelled when the Worker invocation completes, which would silently drop the KV write. Cost stays bounded: recordLatency hits the in-memory throttle and does no KV I/O on all but ~1 request per 20s per server.
- **Renamed** `fetchFromLeastBusyServer` → `fetchFromWeightedServer` (it never did least-busy).
- **Added tests**: invalid latency (0/neg/NaN) ignored, write after throttle expiry.

Not changed (deliberate): KV read-modify-write has no CAS, but a lost latency sample just means one stale weight for ≤20s and self-corrects — not worth CAS complexity. "Last latency" outlier sensitivity is inherent and was an explicit design choice over a moving average. The per-isolate throttle is pre-existing (#11134) and only risks extra KV writes under many concurrent isolates, not correctness.

PR #11878, opened 2026-06-17
feat: flat 0.01/hr pollen drip + drop estimate-based negative prevention

voodoohop:
- Set every tier's hourly pollen drip to **0.01** (`microbe` stays 0) in `shared/tier-config.ts`; the refill SQL already caps at the per-tier value, so this is the only source change for the drip.
- `checkBalance` (gen): drop the model-cost **estimate** logic (`getModelStats`/`getEstimatedPrice`/`canCoverEstimatedCharge`). Non-paid models are now allowed whenever the user has a positive balance (`tierBalance > 0` OR `packBalance > 0`) — no per-request negative prevention. A single request can push the balance negative; the hourly drip recovers it.
- Paid-only models still require positive **pack** balance; the API-key budget gate now blocks only when the key's `pollenBalance <= 0`.
- Remove the now-dead `gen.pollinations.ai/src/utils/model-stats.ts` (`getEstimatedPrice`). `canCoverEstimatedCharge` is kept — still used by the frontend "can afford this model" hint.
- Update `byop-markup` and `tier-system` tests for the positive-balance gate / new drip values.

Tests: gen `byop-markup` 16/16, enter `tier-system` 8/8, gen `tsc --noEmit` clean, biome clean.


ElliotEtag:
We're removing the hourly Pollen drip entirely (#11928 — "Removes hourly Pollen drip"), so the flat 0.01/hr drip and drip-based negative recovery here are superseded. Closing in favor of the tier/drip deprecation.

PR #11885, opened 2026-06-18
fix(polly): extract image_url parts from openai content arrays

fisventurous:
fixes polly's HTTP API endpoint silently dropping images from openai-format content arrays

the `/v1/chat/completions` endpoint parsed list-type `content` for text parts but discarded `image_url` entries entirely. other providers (openai, claude, gemini) handle the same image fine because they receive the full content array. polly's API was the only path stripping them.

changes
\- extract `image_url` parts from list-type message content alongside text
\- merge extracted URLs with the explicit `image_urls` field before passing to `process_with_tools`
\- discord path was unaffected (uses `extract_media_urls` on attachments separately)

ref #11669


sharktide (approved):
LGTM

PR #11948, opened 2026-06-22
feat: add ElevenLabs SFX + Multilingual v2; correct ElevenLabs TTS pricing

ElliotEtag:
Adds two ElevenLabs audio models and corrects ElevenLabs TTS pricing across the board.

### New models
- **`eleven-sfx`** — text → sound effects (`eleven_text_to_sound_v2`). New `generateSoundEffect()` handler on the audio dispatch. `GET /audio/{text}?model=eleven-sfx` and `POST /v1/audio/speech`. Supports `duration` (0.5–30s), `loop`, `prompt_influence`. Aliases: `sfx`, `sound-effects`.
- **`eleven-multilingual-v2`** — stable, lifelike TTS (`eleven_multilingual_v2`, 29 languages); registry-only (flows through `generateSpeech`). Aliases: `multilingual-v2`, `eleven-v2`.

### Pricing — corrected to measured cost
Measured actual credit consumption on our Scale account (the API discounts TTS to ~0.6× the UI credit rate). Now matches elevenlabs.io/pricing/api:
- `eleven_v3`: $0.166 → **$0.10/1k** (0.602 cr/char measured) — *fixes a live ~66% overcharge*
- `eleven_flash_v2_5`: $0.083 → **$0.05/1k** (0.301 cr/char) — *live fix*
- `eleven-multilingual-v2`: $0.166 → **$0.10/1k** (0.601 cr/char)
- `eleven-sfx`: **$0.004/s** (24 cr/s measured), one exact per-second rate; caps at 30s = $0.12, matching ElevenLabs' "per generation" price
- All `priceMultiplier: 1` (at-cost). No `output_format` for SFX — it would make the byte-based duration billing vary by format (variable pricing).

### Other fixes (from review)
- TTS input cap 4096 → 10000 (schema + `generateSpeech` guard) for multilingual's 10k limit.
- SFX `duration` floor lowered (schema min 3 → 0.5).
- `polli-cli` SKILL.md: added SFX + multilingual examples.

### Verification
Empirical (live ElevenLabs): generation for both models, `loop`/`prompt_influence`/short-duration SFX, 5.6k-char multilingual, usage headers, Tinybird billing (multilingual 5625 chars → $0.5625 at the new rate), alias resolution, field-parity. Full suite green: gen 404/404, typecheck clean, price snapshot updated (exactly v3/flash/multilingual repriced).

### DIFF
--- a/enter.pollinations.ai/test/__snapshots__/effective-prices.snapshot.test.ts.snap
+++ b/enter.pollinations.ai/test/__snapshots__/effective-prices.snapshot.test.ts.snap
@@ -114,10 +114,10 @@ exports[`effective cost and price per model — snapshot 1`] = `
   },
   "eleven-multilingual-v2": {
     "cost": {
-      "completionAudioTokens": 0.000166,
+      "completionAudioTokens": 0.0001,
     },
     "price": {
-      "completionAudioTokens": 0.000166,
+      "completionAudioTokens": 0.0001,
     },
   },
   "eleven-sfx": {
@@ -130,18 +130,18 @@ exports[`effective cost and price per model — snapshot 1`] = `
   },
   "elevenflash": {
     "cost": {
-      "completionAudioTokens": 0.000083,
+      "completionAudioTokens": 0.00005,
     },
     "price": {
-      "completionAudioTokens": 0.000083,
+      "completionAudioTokens": 0.00005,
     },
   },
   "elevenlabs": {
     "cost": {
-      "completionAudioTokens": 0.000166,
+      "completionAudioTokens": 0.0001,
     },
     "price": {
-      "completionAudioTokens": 0.000166,
+      "completionAudioTokens": 0.0001,
     },
   },
   "elevenmusic": {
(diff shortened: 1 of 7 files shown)


ElliotEtag:
Both review findings fixed in 02f95d3:

- **SFX `response_format`** — now rejects non-mp3 with 400 instead of silently returning mp3. Not wired through to ElevenLabs `output_format` on purpose: the per-second price is derived from the 128 kbps MP3 byte rate, so honoring other formats would require per-format billing math (variable pricing). Default `mp3` passes unchanged.
- **GET `prompt_influence`** — added a zod `.pipe(z.number().min(0).max(1))` so `NaN`/out-of-range 400 like the POST schema. Applied the same guard to GET `duration`, which had the identical unvalidated `parseFloat` bug.
- Tests: new `test/elevenlabs-sfx.test.ts` (body forwarding, duration clamp, response_format reject/allow) + a GET-validation regression test in `index.test.ts` asserting 400 fires before any ElevenLabs call.

Full suite green (409), typecheck clean.

PR #11956, opened 2026-06-23
fix: warn user when localStorage quota is exceeded

stormdede515-eng:
Silent failure when image attachments overflow the quota causes data loss on reload with no indication to the user.


voodoohop (approved):
Approved after updating to latest main and wiring the existing chat toast helper so the quota warning is visible.

PR #12017, opened 2026-06-26
docs: add DeepSeek Python example to APIDOCS.md

jamalbot204:
Hi! I added a Python code example to `APIDOCS.md` showing how to call the new DeepSeek model using a system prompt with the OpenAI SDK. This will help developers easily integrate and anchor the model's identity.


jamalbot204:
.

voodoohop:
Thanks for contributing. Closing because:
- the added ```python fence is never closed, which would break rendering of the rest of APIDOCS.md
- a generic Python OpenAI-SDK example with the same call shape already exists directly above the insertion point (only delta is `model="deepseek"`)
- docs policy is minimal examples — model IDs are listed via `/v1/models` rather than per-model snippets

PR #12024, opened 2026-06-26
[codex] cleanup dashboard news runtime fetch

voodoohop:
- Removes the dashboard-side GitHub highlights fetch, markdown parser, and dynamic news card renderer.
- Keeps the pinned announcements and FAQ in place.
- Keeps the GitHub link available from the announcements section for full news history.

Validation:
- `npx biome check --write enter.pollinations.ai/frontend/src/components/news-faq/news-banner.tsx enter.pollinations.ai/frontend/src/components/news-faq/news-faq.tsx`
- `git diff --check`

Note: local worktree dependency install hit disk-space limits earlier, so validation used symlinks to the existing installed dependency tree and removed them before commit.


voodoohop:
Closing after quality pass. This removes dashboard runtime news behavior, which is more of a product/content decision than a systemic cleanup. The code is simpler, but this should be re-proposed only after confirming the News section is no longer needed.

PR #12026, opened 2026-06-26
refactor: simplify dashboard hash routing

voodoohop:
- Adds a shared dashboard hash parser for common aliases and authenticated-only dashboard hashes.
- Updates signed-in and signed-out routes to use the shared parser while preserving existing fallback behavior.
- Keeps OAuth callback hash forwarding untouched and preserves `#buy-pollen`, `#keys`, `#models`, activity aliases, and FAQ anchors.

Checks:
- `npx --prefix /Users/thomash/Documents/GitHub/pollinations biome check --write <changed files>`
- `npm run typecheck --workspace pollinations-enter --ignore-scripts`


voodoohop:
Closing after stricter cleanup audit. This refactor centralizes hash parsing, but it adds more code than it removes (+67/-23) and does not simplify the system enough to justify a standalone PR. If we revisit dashboard routing, it should be a broader route model cleanup with net complexity reduction.

PR #12032, opened 2026-06-27
[codex] cleanup stale dead-code finder script

voodoohop:
- Removes the unreferenced regex-based Enter dead-code finder script.
- Avoids keeping a stale analysis tool whose output is noisy against the current codebase.

Validation:
- `rg "find-dead-code" -n . --glob '!node_modules/**' --glob '!**/.git/**'` returned no references after deletion
- `git diff --check`

Note: validation used temporary symlinks to the existing local dependency tree and removed them before commit.


voodoohop:
Closing after quality pass. The stale regex dead-code finder is removable, but this PR is too tactical to count as one of the larger systemic cleanup branches. We can fold this deletion into a broader Enter scripts/tooling cleanup later.

PR #12051, opened 2026-06-27
Update README.md

shaheertariq89:


sharktide:
I'm sorry, but this is not a valid change and looks like an attempt to abuse the quest system. The diff is simply a remove line, which doesn't make a difference. Additionally, the way that this pull request is structured is with no useful title (signals you don't have any experience making pull requests) and does not reflect the experience you boast on your github profile. 

I think this should be closed. cc @voodoohop 

Also if my comments are a bit mean its just because at this point I am fed up with useless pull requests

PR #12062, opened 2026-06-28
fix: convert progressive JPEG to baseline for Fireworks Flux

### LINKED ISSUE #11582: Change default JPEG format from progressive to baseline
Dear [Pollinations.ai](https://pollinations.ai/) team,
I'm using your API endpoint https://gen.pollinations.ai/image/... with the flux model to generate images for embedded devices (ESP32).

The problem:
Your API returns progressive JPEG by default. Progressive JPEGs cannot be properly decoded on microcontrollers like ESP32 because:

Lightweight JPEG decoders (JPEGDEC, TJpg_Decoder) do NOT support progressive JPEGs

They only decode the first (DC) layer, resulting in an image 8 times smaller with severe block artifacts

The image becomes completely unusable on embedded displays

What I see:

text
JPEG size: 768 x 512 (progressive, 8 layers)
Decoded result: 60×40 blocks
What I need:
Please change the default output format from progressive to baseline JPEG.

Baseline JPEG (FF C0 marker) is the standard format supported by all lightweight decoders. Progressive JPEG (FF C2 marker) is not supported and breaks compatibility with embedded devices.

Why this matters:
Progressive JPEG offers no advantage for most use cases

It breaks compatibility with ESP32, Arduino, and other microcontrollers

Many developers are using your API for DIY display projects

There is currently no workaround on the client side

Request:
Please make baseline JPEG the default output format for all models, including flux.

No new parameters needed — just change the default behavior.

Thank you for your consideration!

YoannDev90:
Cleaned-up version of #12050:

- Add missing `isProgressiveJpeg` import (TS error)
- English comments (was Russian)
- Remove `forceBaseline` scope creep from `convertToJpeg`/`resizeForGptImage`
- `any` → typed `ImageOutputOptions & Record<string, unknown>`

Co-authored-by: mikewap83 <mikewap83@users.noreply.github.com>

Fix https://github.com/pollinations/pollinations/issues/11582


YoannDev90:
#12050

voodoohop:
Thanks for the cleanup work @YoannDev90 (and @mikewap83 for the original investigation — the ESP32 progressive-JPEG decode issue is real and well-described).

Closing this after verifying the approach on the actual Cloudflare edge:

- The Workers Images binding (`images.input().output()`) has **no progressive/baseline control**. `progressive: false` isn't a real option — it's silently dropped, so the conversion is a no-op in production.
- `format: "baseline-jpeg"` (and bare `"jpeg"`) are **rejected by the binding** with `IMAGES_TRANSFORM_ERROR 9432: invalid output format` — those values only work on the `/cdn-cgi/image/` URL path and `cf.image` fetch options, not the binding.
- `format: "image/jpeg"` always emits **progressive** on the prod edge (verified via a throwaway probe worker).

So there's no simple binding-level fix; forcing baseline would require routing image encoding through a separate mechanism, which is more backend complexity than we want for embedded-client support.

For the ESP32 use case, the reliable options are decoding the progressive JPEG client-side, or fetching through a `/cdn-cgi/image/?format=baseline-jpeg` proxy (that path *does* support baseline). Closing #11582 as well.

PR #12113, opened 2026-06-29
Fx inability to scroll on mobile on Authorization page

### LINKED ISSUE #12000: Issue: Cannot scroll in BYOP authorization page in mobile
I think this is mostly self explanatory, but if you can’t scroll down you can’t press the authorize button

aeosntw:
Remove `useScrollLock` call for non-mobile devices.
Maintains status quo for desktop and wider screens.

Fixes [#12000](https://github.com/pollinations/pollinations/issues/12000)


sharktide (commented):
@aeosntw does useScrollLock accept an argument? Because right now with this change we are doing 

```ts
useScrollLock(True)
```
or
```ts
useScrollLock(False)
```

When the original implementation was simply doing 
```ts
useScrollLock() // <-- No arguments
```

If ``useScrollLock`` doesn't accept a boolean argument, it should be changed

PR #12145, opened 2026-07-01
ci: update Node.js version from 20 to 22 across workflows and Dockerfiles

diegoalgg88:
**Scope:** Mechanical Node.js version bump (20 → 22) across CI/CD infrastructure.


voodoohop:
Closing — this went stale against main: three of the touched workflows (app-check-registration, app-upgrade-tier, tier-progression-spore-to-seed) no longer exist, and our deploy workflows have since standardized on Node 24. Also bumping `engines` in the published @pollinations/sdk and @pollinations/mcp from >=18 to >=22 would be a breaking constraint for downstream consumers with no functional need.

Thanks for the effort — if you want to pick up the residual (a few non-deploy workflows still on Node 20), a small focused PR against current main is welcome.

PR #12147, opened 2026-07-01
feat(polli-cli): add core lib modules, tests, and build infrastructure

diegoalgg88:
**Scope:** Foundation layer for polli-cli — all new library modules, comprehensive test suite, and build configuration.


voodoohop:
Closing — this bundles retry-by-default wrappers, Zod response-validation rails, a backward-compat credential path, and silent catch-and-return fallbacks, all of which go against how we build (one fetch, throw on error; no defensive rails; no back-compat shims). It's also branched from before #12274, so it deletes the quests/my-models commands that are on main. Small, single-concern polli-cli PRs against current main are very welcome.

PR #12114, opened 2026-06-29
Implement robustness, validation, and UX improvements in polli-cli

diegoalgg88:
### Description:

This Pull Request introduces significant improvements to the `polli-cli` to enhance its robustness, maintainability, and extensibility. A key addition is the implementation of an **MCP (Model Context Protocol) server**, enabling AI agents to programmatically interact with Pollinations' capabilities.

**Key Changes:**

* **New Feature: MCP Server Support:**
* Implemented an MCP server allowing integration with AI agents/assistants.
* Exposes standard tools: `gen_image`, `gen_text`, `gen_audio`, `gen_video`, `transcribe`, `list_models`, `get_usage`, and `upload_file`.
* Supports both `stdio` and `SSE` transport modes for flexible integration.

* **Robustness & Validation:**
* Implemented Zod schemas for strict input validation and API response typing.
* Added retry mechanisms with backoff for transient errors.
* Improved local validation for models and parameters before API calls.

* **UX Improvements:**
* Integrated progress indicators/spinners for long-running operations.
* Enhanced CLI help documentation with concrete examples.

* **Architecture & Maintainability:**
* Set up a testing foundation (Unit tests) using Vitest.
* Implemented a local cache for model metadata to reduce latency.
* Added support for persistent configuration (extending credential management).

* **Security:**
* Refactored hardcoded tokens and improved secure storage handling.

**Testing:**
Added comprehensive unit tests covering the new `lib` modules (API, validation, retry, etc.) and MCP handlers in `src/tests/unit/`.

These changes professionalize the CLI, making it a reliable tool for both manual CLI usage and automated AI agent workflows via MCP.

PR #12149, opened 2026-07-01
feat(polli-cli): improve commands with validation, i18n, spinners, and output formats

diegoalgg88:
**Scope:** Enhance all existing polli-cli commands with the new foundation modules.


voodoohop:
Closing together with #12147 — this diff fully contains #12147's files, so the split from #12114 didn't actually decouple the stack, and the same issues apply (retry loops, AbortSignal.timeout, silent config fallbacks, plus i18n/keytar/caching infra a dev CLI doesn't need yet). One focused PR per concern against current main is the way in.

PR #12157, opened 2026-07-01
Add model search bar and true generation-throughput (tokens/sec) metric

tomdacatto:
## Summary

**Dashboard: model search bar**
- Adds a search input to the Models section (`enter.pollinations.ai`) that filters by name, display name, description, and brand.
- Tabs stay pinned to the full catalog — searching never hides or auto-switches tabs, only the active tab's contents filter. Shows a "no results" message when a filter matches nothing in the current tab.

**Model Monitor: tokens/sec ("Speed") column**
- Adds a sortable "Speed" column to `apps/model-monitor` showing completion tokens per second of actual generation time, with a tooltip explaining the calculation.
- This required a real fix, not just a UI column: the naive version (completion tokens ÷ total `response_time`) is wrong for streaming text. `track.ts`'s `endTime` is captured as soon as `next()` resolves, which happens once the upstream response *headers* arrive — before the stream body is read. So `response_time` reads as near-instant for streamed requests, and dividing by it would inflate throughput the same way a cache hit does.
- Fixed by timestamping the first and last SSE chunk inside the existing stream-parsing loop in `gen.pollinations.ai`'s track middleware, and threading a new `streamDurationMs` field through to the Tinybird event. `model_health.pipe` now uses that pure generation window for streamed responses, falling back to `response_time` only for non-streamed (fully buffered) responses, and excludes cache hits from both sums (a cache hit returns full token counts in near-zero time and would otherwise dominate the ratio).
- Added a new test (`gen.pollinations.ai/test/tracking-observability.test.ts`) that builds a real SSE stream with delayed chunks and asserts `streamDurationMs` reflects the actual elapsed generation time, not the near-instant `response_time`. Full `gen.pollinations.ai` suite (439 tests) passes.

**Model Monitor: TTFT (time to first token) column**
- Adds a sortable "TTFT" column showing p50 time-to-first-token, with p95 in a per-row tooltip.
- Reuses `response_time` (already ≈ time-to-first-byte for streams, per the fix above) rather than adding new instrumentation. Scoped to the streaming-text cohort via `stream_duration_ms IS NOT NULL`, and excludes cache hits and non-streamed requests so it isn't diluted by full round-trip latency for non-streaming traffic.
- Percentiles are computed live at query time from raw `generation_event` rows, same as the existing Avg/P95 latency columns, and automatically respect the existing window selector (5m/1h/4h/24h/7d) — no separate per-hour/day storage was needed.

## Notes for reviewers

- Both `stream_duration_ms` (new datasource column) and the `model_health.pipe` changes need to be redeployed to Tinybird before the Speed and TTFT columns show real numbers — until then they render `-` gracefully (verified live against the current production catalog/pipe).
- `enter.pollinations.ai`'s own test suite requires `sops`-decrypted secrets I didn't have access to in this environment, so those tests weren't run locally — but my only change there is the datasource/pipe SQL text (no app logic), and it typechecks clean.

## Test plan
- [x] `gen.pollinations.ai` full test suite (439 tests, including new streaming-duration test)
- [x] `tsc --noEmit` clean on `gen.pollinations.ai` and `enter.pollinations.ai`
- [x] `biome check` clean on all changed files
- [x] Search bar verified live in a browser against the real model catalog (filtering, tab-pinning, no-results state)
- [x] Model Monitor Speed and TTFT columns verified live against the current production pipe (render `-` gracefully pending pipe redeploy) with working sort and tooltips
- [ ] Tinybird pipe redeploy (needs someone with deploy access) before Speed/TTFT columns show live numbers


YoannDev90:
Maybe also add TTFT (Time to First Token). Would that also be stored as p90, p95, etc., with values per hour, per day, etc. ?

tomdacatto:
> Maybe also add TTFT (Time to First Token). Would that also be stored as p90, p95, etc., with values per hour, per day, etc. ?

Il add that also

tomdacatto:
Splitting this into two PRs per request:
- #12286 — model search bar (frontend only)
- #12287 — generation-throughput (Speed + TTFT), where the `stream_duration_ms` datasource question is now being discussed

Will close this one once both are confirmed to cover everything here.

voodoohop:
Closing as completed-by-split — the search bar landed via #12286 and the throughput/TTFT half is under review in #12287. Nice work getting the first half in 🌱

PR #12171, opened 2026-07-02
feat: route flux to self-hosted Vast.ai 5090 pool with Fireworks fallback

### LINKED ISSUE #11582: Change default JPEG format from progressive to baseline

voodoohop:
- Adds `callFluxWithFallback`: flux tries the self-hosted GPU pool first, any pool error (incl. empty pool) falls back to Fireworks unchanged
- `callSelfHostedServer` gains a `poolType` param (zimage default unchanged)
- `nunchaku/server.py`: env-parametrized `QUANT_MODEL_PATH` / `MAX_PIXELS`; `QUEUE_LIMIT` (default 10) sheds load with 503 → gateway falls back to Fireworks
- `setup-vast.sh`: one-command Vast provisioning incl. Cloudflare tunnel (`TUNNEL_NAME=...`); workers MUST be tunneled — Workers `fetch()` can't reach raw-IP:port origins
- `GPU_INSTANCES.md`: live Vast fleet, provision one-liner, gotchas
- Tests: pool-hit / pool-empty / pool-error fallback (3 passing)

**Already deployed to prod** (2026-07-02 11:14 UTC) — serving from instance 43575766; queue-cap verified shedding at US peak.

Incident note: a polly-suggested `AbortSignal.timeout` on the pool fetch passed vitest but broke every pool fetch in deployed workerd (silent 100% Fireworks fallback for ~1.5h). Reverted with warning comment — do not re-add; local workers-pool ≠ prod workerd.

Pilot benchmark (1×5090 FP4, $0.43/hr): 1,893 img/hr sustained, p50 2.25s @ full 1024², baseline JPEG (fixes #11582), FP4/FP8 same-seed A/B at parity.

Context: Fireworks flux spend $2,149/mo (June); 2×5090 target ≈ $620/mo + failover. Addresses #12136 fallout.


voodoohop:
polly

voodoohop:
Review items addressed in 03194fa00:

- Pool-leg timeout added: 30s `AbortSignal.timeout` (not 15s — at-capacity queueing p95 is ~5s and can reach 10-15s at peak; 30s still catches wedged workers without spurious mass-fallback)
- Fallback catch now logs the full error object incl. stack, not just `.message`
- `server.py` startup guard caps `MAX_PIXELS` to 810k on <30GB-VRAM GPUs (4090 protection)
- `str(900*900)` → `"810000"`

On the Fireworks mock format: `callFireworksFluxSchnellAPI` requests `Accept: image/jpeg` and reads `response.arrayBuffer()` directly — raw JPEG bytes is the real response shape (fireworksFluxModel.ts:56-77), so the mock matches production behavior. No JSON envelope involved.

Keeping catch-all in v1: a fallback that also covers unexpected error types is the safer failure mode for image serving; the stack logging makes regressions visible.

PR #12172, opened 2026-07-02
Fix #10807: Float Precision Bug Resolved via Micro-Pollen Integer Math

### LINKED ISSUE #10807: Float precision bug: pollen calculations use float64 instead of int64, causing precision loss
## Summary
The billing system uses JavaScript `number` (float64) for pollen balance calculations instead of integers or exact decimals. This causes precision loss on every transaction, effectively over-charging users with large balances.

## Problem
Float64 cannot exactly represent many decimal values. For example:
- `0.1 + 0.2 = 0.30000000000000004` in float64
- At scale: 200 pollen balance → deduction of `0.000000001` may actually deduct `~0.000007` due to precision errors
- Small balances (1 pollen) have better relative precision than large balances (200 pollen)

## Evidence from codebase
Found in `enter.pollinations.ai/test/integration/tier-system.test.ts`:
```javascript
expect(balance.tierBalance).toBeCloseTo(0.01, 10);
expect(balance.packBalance).toBeCloseTo(-0.015, 10);
```
The use of `toBeCloseTo()` with 10 decimal precision instead of `toBe()` confirms float arithmetic is in use.

## Impact
Users with large pack balances (>100 pollen) may lose precision on every deduction compared to users with small balances. This is **invisible to users** but compounds over many requests.

## Solution
Convert pollen to **integer base units** (e.g., 1 pollen = 1,000,000 micro-pollen) for all database storage and calculations. Only convert to float for display formatting.

## Files affected
- `shared/billing/deduction.ts` — `atomicDeductUserBalance()` uses `number` type
- `enter.pollinations.ai/src/utils/balance-deduction.ts` — same issue
- Database schema — `tierBalance`, `packBalance` likely stored as float

/cc @marcosfrgames08 for discovery

---
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1500964478936809575/1502688746699620472)

ZevrinBuilds:
> **Fixes Issue #10807**
> 
> **What I changed:**
> * Created a `pollenMath.ts` utility that enforces exact `1_000_000` multiplier logic.
> * Updated `atomicDeductUserBalance`, `atomicDeductApiKeyBalance`, and `atomicCreditUserBalance`.
> * Instead of risking a heavy database migration to change the D1 `real` columns, this PR uses an elegant SQL boundary fix: it multiplies and `ROUND()`s the database floats into precise integers *inside* the SQL query, executes the arithmetic safely in integer space, and divides it back into a clean float before saving. 
> * This instantly prevents `0.1 + 0.2 = 0.300000004` drift without breaking any existing table schemas.


MarcosFRG:
👌

voodoohop:
Closing after a careful read — the integer math doesn't eliminate float drift: the result is divided back by 1e8 and stored into a REAL column, so binary representation error returns on write. Main already quantizes every ledger amount via roundPollenLedgerAmount (shared/billing/precision.ts) before deduction, which covers the practical drift in #10807. Thanks for digging into the billing path.

PR #12048, opened 2026-06-27
feat: add 3D model generation — Trellis 2 (inferenceport) + Rodin 2.5 (fal.ai)

tomdacatto:
## Summary

- Adds `GET /3d/{prompt}` blocking endpoint and `POST /3d/generations` + `GET /3d/jobs/{job_id}` async job API for 3D model generation
- Two models: **Trellis 2** (inferenceport, sync API) and **Hyper3D Rodin 2.5** (fal.ai)
- Reuses `generate.image` event type and `completionImageTokens` for billing — no Tinybird schema changes

## Models

| Model ID | Provider | Price | Input |
|---|---|---|---|
| `trellis-2-low` (default) | inferenceport | $0.24/gen | image |
| `trellis-2-medium` | inferenceport | $0.29/gen | image |
| `trellis-2-high` | inferenceport | $0.35/gen | image |
| `hyper3d-rodin` | fal-ai/hyper3d/rodin/v2.5/fast | $0.40/gen | image or text |

## Implementation notes

- Trellis 2 uses inferenceport's **sync API** (`?sync=true`) on the blocking endpoint — single request, no polling loop
- Async job API (`POST /3d/generations` + `GET /3d/jobs/{id}`) uses inferenceport's async job path for trellis-2 and fal.ai's queue for Rodin
- `model/gltf-binary` content-type support added to `track.ts` billing guard (was silently skipping all 3D billing)
- Live API test confirmed: inferenceport output is nested under `data[0]` (not top-level), model name is `"trellis2"` (no hyphen)
- Completed job results stored in R2; re-polls served from cache with `X-Cache: HIT` to prevent double-billing
- Removed the `format` param (obj/usdz/fbx were never actually produced by either provider — both only return GLB); trimmed to what's real per YAGNI
- Added a `seed` param: passed through to Rodin (fal confirms support); accepted but not forwarded for Trellis 2 (inferenceport support unconfirmed) — still forces a fresh generation via the media-cache key even when unused by the provider
- Rodin text-only requests now route to fal's dedicated `text-to-3d/fast` endpoint instead of the image endpoint (which 422s without `image_urls`)
- Rodin output content-type pinned to `model/gltf-binary` regardless of fal's reported `content_type`, matching Trellis 2 and keeping `model3dCache` working for both

## Test plan

- [ ] `GET /3d/{prompt}?model=trellis-2-low&image=<url>` returns GLB with correct billing
[...]

PR #12186, opened 2026-07-02
fix: render 3D model prices in dashboard

voodoohop:
3D models showed up in the dashboard's new 3D tab but with an **empty price cell**.

### Cause
`modelPriceFromCatalog` has no `type === "3d"` branch, so 3D models fell through to the final `return` (the text-model formatter), which reads `promptTextTokens`/`completionTextTokens` — fields a 3D model doesn't have. `priceLines` drops empty lines → `prices: []` → blank cell. (The `perRequest`/`perImagePrice` 3d block from #12048's history was dropped on rebase, and returned fields that don't exist on `ModelPrice` anyway.)

### Fix
- Reuse the **existing image flat-rate branch** — 3D stores its flat per-generation fee in `completionImageTokens` exactly like flat image models, so it renders the correct `/gen` line with **no new price type, unit, or config**.
- Remove the dead/unreachable `"3d"` block.
- Add a pricing-data test asserting each 3D model yields one flat per-request line.

Sort price + per-Pollen already worked (`completionImageTokens` is in `OUTPUT_PRICE_FIELDS`).

Follow-up to #12048.

### DIFF
--- a/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
+++ b/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
@@ -258,7 +258,10 @@ function modelPriceFromCatalog(model: ApiModelInfo): ModelPrice | null {
         };
     }
 
-    if (price.type === "image") {
+    // 3D generations are flat per-generation charges stored in
+    // completionImageTokens (same convention as image models), so they render
+    // through the image branch: no prompt-token prices → flat "/gen" line.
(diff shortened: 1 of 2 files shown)


voodoohop:
Superseded by #12188 (merged), which fixes the same 3D price-cell bug and additionally adds a cube fallback icon + Microsoft/Deemos brand logos. Closing in favor of that.

PR #12193, opened 2026-07-03
fix: correct stale reactivate instructions on disabled model card

voodoohop:
- The dashboard's "edit, test, then save to reactivate" message is stale — that flow was intentionally removed in #12154 to stop silent reactivation on unrelated saves
- There's currently no self-serve reactivate path at all; owners need a maintainer to flip it back on
- Points owners at that instead of a broken instruction


voodoohop:
Superseded by #13527, which replaced the old disabled state with a self-service Hide / Relist flow. Thanks — this correctly identified the stale instruction at the time.

PR #12251, opened 2026-07-06
chore: small cleanup fixes

sixfingerdev:
- Remove unused parameter in sleep() utility
- Remove empty <track> elements causing browser warnings
- Add missing aria-label to copy button for accessibility
- Clean up stale TODO comment in star history function


voodoohop:
Closing — the aria-label duplicates the existing title attribute, the other two changes are comment churn, and the PR body describes a <track> cleanup that isn't in the diff. Nothing here changes behavior.

PR #12252, opened 2026-07-06
feat: D1-backed media catalog (tags, galleries) on media.pollinations.ai

voodoohop:
- Adds `media_item`/`media_tag` tables to **enter's D1** (`shared/db/media-catalog.ts`, migration `0035`) — same DB gen already binds, so catalog rows join with `user` and PR 2 can wire `?tag=` on gen URLs with a direct insert
- `/account/key` response gains server-attested `userId`/`keyId`/`byopClientKeyId` — media stamps ownership from these, never from request params (spoof-proof, tested)
- `POST /upload`: optional `tag`/`tags`/`prompt`/`model` via query, multipart form, or JSON body; strict validation (lowercase slug tags, max 8, 400 on invalid); every user-owned upload is cataloged, metadata on userless keys → 400
- `GET /me/media?tag=&limit=&cursor=` — personal library (403 for keys without a user); `GET /tags/:tag` — public gallery; newest-first keyset pagination, no owner fields exposed
- Catalog D1 write awaited inline — failure = 500, no silent drops
- Tests: 15 pass (7 pre-existing + 8 new: spoofing, tag validation, tag merge on re-upload, pagination); migrations applied via enter's drizzle dir in vitest-pool-workers

Deploy notes:
- Migration must be applied via enter's `npm run migrate:staging` / `migrate:production` **before** deploying the media worker
- Workers deploy on `production` branch push, not `main`
- `media.pollinations.ai` is not an npm workspace member; its duplicate `drizzle-orm` copy causes tsc nominal-type noise against `shared/db` (runtime + vitest unaffected; CI has no tsc gate) — follow-up candidate: add media to root workspaces

Design context: follows the tag-only public-discovery thesis from #11321 (API redesigned for simplicity). Likes/reactions and gen `?tag=` wiring are follow-up PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12256, opened 2026-07-06
feat: opt-in ?tag= cataloging for image/video generations

voodoohop:
Stacked on #12252 (base = `feat/media-catalog-d1`; retargets to main when that merges). No new migration — uses the media_item/media_tag tables from #12252.

- Adds `mediaCatalog` middleware to gen's `/image` + `/video` chain, registered before `imageCache` so its post-`next()` hook fires on BOTH cache hits and fresh generations — retagging an already-cached image works
- `GET /image/{prompt}?...&tag=sunset` (repeated `tag` or comma `tags`) by a user-owned key upserts a `kind: "generation"` catalog row keyed on (owner, canonical locator) and merges tags; the tagged generation then appears in media.pollinations.ai `/me/media` and `/tags/:tag`
- Only tagged generations write to D1 — untagged requests take a zero-cost early return (no per-generation firehose)
- Fail fast before GPU cost: invalid tag → 400 naming it; tags on anonymous/userless keys → 400
- D1 write in `waitUntil` with error logging — same async post-response pattern as billing/tracking in `track.ts`
- `tag`/`tags` added to `EXCLUDED_PARAMS` so tagging never changes cache identity (unit-tested); zod schemas already strip them before the generation backend (asserted in tests)
- Canonical locator example: `https://gen.pollinations.ai/image/sunset%20over%20hills?height=720&model=flux&seed=42&width=1280` (constant host, sorted params, catalog/cache-control params stripped)
- `shared/media-catalog.ts`: tag normalization + generation upsert helpers (media worker keeps its own copy for now — unifying is a follow-up to keep this stack conflict-free)
- Tests: 26 pass in the two touched files, incl. 6 new (D1 row assertions via drizzle, cache-hit retag merge with single upstream call, 400 paths, untagged no-op)

### DIFF
--- a/gen.pollinations.ai/src/middleware/media-catalog.ts
+++ b/gen.pollinations.ai/src/middleware/media-catalog.ts
@@ -102,6 +102,14 @@ export const mediaCatalog = createMiddleware<MediaCatalogEnv>(
             throw error;
         }
 
+        // `?tags=` (or only empty/whitespace values) normalizes to nothing —
+        // treat it like no tags at all rather than requiring auth and
+        // writing a tagless catalog row.
+        if (tags.length === 0) {
+            await next();
+            return;
+        }
+
         const ownerUserId = c.var.auth?.user?.id;
         if (!ownerUserId) {
             return c.json(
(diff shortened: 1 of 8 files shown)


voodoohop:
Superseded by #12277 — the media-catalog stack was consolidated into a single PR (this branch's commits are contained there verbatim). Branch kept until #12277 merges.

PR #12257, opened 2026-07-06
feat: generic reactions for the media catalog (likes and beyond)

voodoohop:
Stacked on #12252 (base = `feat/media-catalog-d1`; retargets to main when that merges). Sibling of #12256.

- Adds `media_reaction` table (migration `0036`): generic reaction system, not likes-specific — open slug vocabulary per kind (`like`, `heart`, `bookmark`, ...), each user gives an item at most one of each kind, cascade FKs to `media_item` and `user`
- `PUT`/`DELETE /media/:id/reactions/:reaction` — idempotent both ways, returns `{reaction, reacted, count}`; 400 invalid kind (slug, max 32 chars), 401/403/404; `:id` is the catalog item id from `/me/media` / `/tags/:tag`
- List items gain `reactions: {like: 3, bookmark: 1}` (always, `{}` when none) and `myReactions: ["like"]` for authenticated callers; `/tags/:tag` auth is now optional — no key stays public, an invalid key fails fast with 401 (no silent anonymous fallback)
- All per-page reaction data batched: one grouped-count query + one my-reactions query per page
- CORS gains PUT/DELETE for browser clients
- Tests: 22 pass (15 pre-existing + 7 new: like/unlike idempotency, two likers, multi-kind coexistence, 404/400/401/403 matrix)

Deploy note: migration 0036 must be applied via enter's `npm run migrate:staging` / `migrate:production` before deploying the media worker (stacks on 0035 from #12252).

### DIFF
--- a/enter.pollinations.ai/drizzle/0036_media_reactions.sql
+++ b/enter.pollinations.ai/drizzle/0036_media_reactions.sql
@@ -1,0 +1,10 @@
+CREATE TABLE `media_reaction` (
+	`item_id` text NOT NULL,
+	`user_id` text NOT NULL,
+	`reaction` text NOT NULL,
+	`created_at` integer NOT NULL,
+	FOREIGN KEY (`item_id`) REFERENCES `media_item`(`id`) ON UPDATE no action ON DELETE cascade,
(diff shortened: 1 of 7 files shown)


voodoohop:
Superseded by #12277 — the media-catalog stack was consolidated into a single PR (this branch's commits are contained there verbatim). Branch kept until #12277 merges.

PR #12274, opened 2026-07-07
[FEAT]: Polli-cli add rewards **kwargs

Circuit-Overtime:
## Changes Made

- Lists earned quest rewards and flags which are ready to claim (--pending/--claimed). Reuses the existing GET /api/quests/rewards ledger via a small enter() API helper. 
- Fixed the router to address all quests via the `quest` runnable 
- Updated the CLI to the new patch version `0.1.6`

> Read-only — claiming still requires a dashboard session by design.


voodoohop:
Thanks for this! Tested it on a live account and it works 🎉

The quest CLI was already a bit complex before.  we've got `quests`, `quests mine`, and now `quests rewards`, and they overlap a lot. `quests mine` already returns each reward with `claimedAt` (`null` = claimable), so `rewards` is basically re-fetching data we already have.

Came to the conclusion that **one `quests` command is enough**:

- Just `polli quests` — always account-aware (like every other command), no `mine`/`rewards` split
- Show claim state right in the status column: `open` / `claimable` / `claimed` / `coming`
- Filters: `--open --claimable --claimed --coming-soon`

Ends up removing more than it adds. Want to take a crack at it, or would you like a hand?

Circuit-Overtime:
you're right, the split was redundant. Collapsed it into a single account-aware polli quests with the claim state right in the status column (open / claimable / claimed / coming) and --open --claimable --claimed --coming-soon filters. Dropped mine, rewards, and the extra enter() helper, and updated SKILL.md. Heads-up: quests now requires a key like the other account commands (no more keyless catalog) felt right for consistency, but flagging it. Pushed the update 🙌

Circuit-Overtime:
polly give me a review

PR #12297, opened 2026-07-08
chore: enable gen worker observability

voodoohop:
- Enables production Workers observability for gen at 100% sampling.
- Keeps the production edge rate limiter removed; `EDGE_RATE_LIMITER` remains only in staging/test.
- Already deployed directly to production as Worker version `45455499-7cf5-417d-85ae-d28aef314f75`.
- Validated with production build, Wrangler dry-run, deployed binding list, and `/models` smoke test.

### DIFF
--- a/gen.pollinations.ai/wrangler.toml
+++ b/gen.pollinations.ai/wrangler.toml
@@ -60,6 +60,10 @@ run_worker_first = ["/", "/manifest.webmanifest"]
 # Production routes
 [env.production]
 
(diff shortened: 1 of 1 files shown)


voodoohop:
- Production has since been redeployed through GitHub Actions without this observability block.
- Cloudflare script settings now report `observability: null` and `logpush: false`.
- Closing to avoid re-enabling 100% Workers Logs sampling and unnecessary log-ingestion cost.

PR #12307, opened 2026-07-09
fix: dashboard models loading & readme link

### LINKED ISSUE #12289: Model catalog not loading on enter.pollinations.ai dashboard (multiple users affected)
## Problem

The full model catalog fails to load on the [enter.pollinations.ai](https://enter.pollinations.ai) dashboard. User's own models load fine (~1s), but the complete model listing (all available models with pricing, permissions, etc.) never appears.

## Evidence

- Multiple users reporting the same behavior (not isolated to one account)
- User's own models load quickly → the issue is **not** client-side connectivity
- Suggests the endpoint serving the full catalog (likely `gen.pollinations.ai/models` or `/v1/models` via `getOrderedVisibleModelEntries()`) is failing or timing out server-side
- The user's own models likely come from a different, lighter endpoint (`/account/models` or similar) which works fine

## Suspected cause

The full model catalog endpoint may be:
- Timing out when assembling the complete registry with pricing/permissions
- Failing in the permission filtering step (`filterEntriesByPermissions`)
- Returning a 500 or hanging on the backend

## Reproduction

1. Log into enter.pollinations.ai
2. Observe: user's own models load, full catalog does not
3. Check dev console (F12 → Network) for the status of the `/models` or `/v1/models` request

## Context

Reported during a period of platform instability following a national outage in Cuba — but since multiple users outside that context also report it, this appears to be a server-side issue, not client-side.

---
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1510747614641914159/1524137458844041316)

ikkhan0:
This PR resolves two issues:
1. **Major Bug Fix**: Wraps community database query in `loadGenerationModelRegistry` in a `try/catch` block. This prevents D1 database timeouts or temporary failures from causing the entire `/models` API (and therefore the enter.pollinations.ai models dashboard) to return a 500 error, instead falling back gracefully to static models. (Fixes #12289 and #12280)
2. **Documentation Fix**: Corrected a broken link in the root `README.md` for the `pollinations.ai` React app (updated `./app/` to `./pollinations.ai/`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,7 +371,7 @@ For any questions or support, please visit our [Discord channel](https://discord
 
 Our codebase is organized into several key folders, each serving a specific purpose in the pollinations.ai ecosystem:
 
-- [`pollinations.ai/`](./app/): The main React application for the Pollinations.ai website.
+- [`pollinations.ai/`](./pollinations.ai/): The main React application for the Pollinations.ai website.
 
 - [`image.pollinations.ai/`](./image.pollinations.ai/): Backend service for image generation and caching with Cloudflare Workers and R2 storage.
 
--- a/gen.pollinations.ai/src/model-registry.ts
+++ b/gen.pollinations.ai/src/model-registry.ts
@@ -137,9 +137,17 @@ function buildRegistry(
 async function loadGenerationModelRegistry(
     dbBinding: CloudflareBindings["DB"] | undefined,
 ): Promise<GenerationModelRegistry> {
-    const communityEntries = (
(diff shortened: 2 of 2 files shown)


voodoohop:
Closing — the catch-and-serve-static fallback is the kind of silent fallback we deliberately avoid: a transient D1 blip would cache a registry without community models for the whole TTL, turning a visible error into invisibly missing models. The #12289 outage this targets was DNS-side and /models is healthy today. The README link fix is real though — feel free to resubmit just that line.

PR #12330, opened 2026-07-10
fix: update Buffer Instagram media payload

ElliotEtag:
- Updates Buffer image and carousel payloads to the current ordered `assets` input.
- Uses Instagram post metadata for multi-image carousels; Buffer now rejects `type: carousel`.
- Preserves the existing single-image and multi-image publishing behavior.

Root cause: Buffer changed both its media input shape and accepted Instagram post types. Renewing the API key restores authentication, but the old mutation payload is no longer valid.

Checks:
- Buffer API authentication and all four configured channels verified read-only
- Single-image and carousel payload assertions passed
- Python compilation passed
- Live Instagram draft mutation tested and cleaned up
- `git diff --check` passed

### DIFF
--- a/social/scripts/buffer_publish.py
+++ b/social/scripts/buffer_publish.py
@@ -133,20 +133,12 @@ def create_buffer_post_with_retry(
     return {"success": False, "error": "Max retries exceeded"}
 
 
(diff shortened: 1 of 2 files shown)


ElliotEtag:
polly please review the Buffer media payload migration.

PR #12281, opened 2026-07-07
feat: private/public visibility for community models

voodoohop:
Splits **defining** a community model (private, owner-only, free) from **publishing** it (allowlist-gated, listed, billed to callers) via a single `visibility` enum: `private | app | public`.

Ships `private` + `public`. `app` is reserved in the schema; its call-gate (owner's app users via `byopClientUserId`) and UI are a follow-up — selecting it today behaves as owner-only.

### Model
| | callable by | listed in `/models` | pricing |
|---|---|---|---|
| `private` (default) | owner only | no | none (billed to owner) |
| `app` *(follow-up)* | owner + app users | no | required |
| `public` | anyone | yes | required |

### Changes
- **Schema**: `visibility` column + migration `0038` (default `private`); threaded DB → gen registry.
- **Catalog**: only `public` endpoints are listed (`model-registry.ts` visible gate). All catalog surfaces already filter on `visible`.
- **Access gate**: private endpoints are owner-only. Caller `userId` (already in ctx before `resolveModel`) is checked in `resolveModelDefinition`; non-owner gets "invalid model" — indistinguishable from unknown, so private models aren't discoverable. Fails closed.
- **Billing**: owner reward fires only for `public`; private is billed to the owner with no markup.
- **Policy inversion**: creation is now open to everyone; the allowlist + required base pricing move to the sharing action.
- **Dashboard**: create/edit dialog is private-only (no pricing); a separate "Make public" card action opens a publish dialog (pricing + tool fees). Private/Public badge per card.

### Tests
- New: private model is owner-only callable (non-owner 400, owner 200, absent from catalog).
- Existing public-flow fixtures marked `visibility: public`; allowlist test reframed to the new create-open / publish-gated policy.
- 24 gen community tests + 13 tracking-observability pass; both packages tsc + biome clean.

### Deploy order
Migration `0038` **first**, then enter + gen. Already applied + deployed to **staging** (staging.enter / staging.gen).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12333, opened 2026-07-10
feat: private/public visibility for community models (standalone)

voodoohop:
Extracts the visibility slice onto current main, independent of the agents work.

- Adds `visibility` (`private | public`) to community endpoints; new models default private and existing models are backfilled public without changing their prices
- Any authenticated owner can register and use private models
- Private models are owner-only and appear in model catalogs only when authenticated with the owner API key
- Allowlisted owners can select Public during creation or later editing; public models may be free (`0`) or owner-priced
- Public models are globally callable/listed; priced usage retains the existing community-owner reward
- Dashboard shows the same Private/Public control on create and edit; Public is disabled with an approval explanation for non-allowlisted owners
- Polli CLI supports visibility and pricing on both create and update
- Excludes agent kinds, capability flags, tool pricing, uploaded source, prompt agents, worker deployment, and minimum-price policy

Validation: community endpoint + tracking integration tests 33/33; Gen and Enter typechecks; Drizzle migration check; Biome clean; migration dry-run preserves zero and positive legacy prices.

## Review hardening

- Endpoint probes are allowlist-gated and refuse redirects, preventing a validated base URL from redirecting to an unvalidated destination
- Updates persist visibility + the complete effective price set atomically under concurrent updates
- Community price definitions keep explicit zero rates; free private/public calls bill and reward 0 without missing-conversion-rate warnings
- Zero-priced community models bypass the balance preflight; paid public models retain normal budget and balance checks

## Known behavior

- Visibility changes propagate to gen within ≤60s (per-isolate registry cache TTL, pre-existing design)

## Rollout

1. `npm run migrate:production` (adds `visibility`, backfills existing rows public without changing prices) — staging already migrated
2. Deploy enter + gen
3. Corrective sweep for rows created by pre-deploy workers during the gap: `UPDATE community_endpoint SET prompt_text_price=0, prompt_cached_price=0, prompt_cache_write_price=0, prompt_audio_price=0, prompt_image_price=0, completion_text_price=0, completion_reasoning_price=0, completion_audio_price=0 WHERE visibility='private'` (expected to touch 0 rows)


voodoohop:
Scope decision for this PR:

- `/models` / `/v1/models` stays **caller-independent** — public, cacheable catalog listing only `public` endpoints. No auth-dependent branching in the catalog route.
- Owners see their private models via the authenticated my-models API / dashboard (already the case). Private = unlisted everywhere; discoverability is what publishing grants.
- Keep: create-always-private (server-enforced), publish only via edit with allowlist + pricing checks, drop reserved `app` state (re-adding later is TS-only, no migration).
- When removing `app`, keep the shared-visibility check in one place (billing/reward path + allowlist gate) so a third state can be reintroduced with a single-site change.

voodoohop:
Follow-up (own PR): `app` visibility for BYOP apps — models callable by the owner + users logged into the owner's app (`byopClientUserId` gate, per #12281's sketch). Deferred from this PR; re-adding is TS-only (no migration). Open questions for that PR: pricing required vs billed-to-dev, allowlist gating, edit-dialog UI.

voodoohop:
Correction to the earlier scope comment (supersedes the `/models` point): **`/models` / `/v1/models` SHOULD include the caller's own private models when the request is authenticated with a key.**

- Anonymous / no key → public catalog only (stays cacheable as-is)
- Valid key → public catalog + that key owner's private models
- Only cache the anonymous response; keyed responses are computed per caller (no shared-cache pollution)
- Private models of *other* owners must never appear — filter strictly by the resolved key's owner user id

Rest of the earlier comment stands (create-private server-enforced, publish via edit w/ allowlist + pricing, drop `app`, single-site visibility seam).

voodoohop:
UX decision (supersedes create-always-private): the **visibility selector appears on the create form too** — direct-to-public creation is allowed when preconditions pass.

- Create accepts `visibility` (default `private`), validated by the **same** server-side path as the edit/publish transition: `public` requires allowlist + pricing. No separate publish action/dialog.
- Create form: Private/Public selector defaults Private; selecting Public reveals pricing fields; disabled with "publishing requires approval" when not allowlisted.
- Edit dialog keeps the identical selector for publish-later / unpublish.
- Card keeps the Lock/Globe state chip.

voodoohop:
Addendum: **create and edit share one API contract** — identical payload (one zod schema), identical validation incl. the visibility invariants (`public` ⇒ allowlist + pricing), one form component. POST creates, PUT edits; no lifecycle-specific endpoints or create-only rules.

voodoohop:
Review hardening pushed in f9203c8f8, driven by a multi-agent adversarial review + two independent Codex (gpt-5.6-sol) reviews:

- Probes (`/test`, `/models`) now refuse redirects — base-URL validation (https + private-host blocklist) was bypassable via 3xx, and probes are open to all authenticated accounts on this branch
- Updates write visibility + the full effective price set atomically — concurrent partial updates could interleave into `public` + zeroed prices
- `communityPriceDefinition` keeps explicit zero rates — private-model calls billed 0 but logged "missing conversion rate" warnings on every request

Tests: 32/32 (community-endpoints + tracking-observability), typechecks clean on enter + gen.

Accepted as-is: visibility flips propagate within ≤60s (pre-existing per-isolate registry cache). Rollout/runbook notes added to the PR description.

PR #12366, opened 2026-07-13
chore: remove infra and scripts directories

ElliotEtag:
- Removes the `infra/` directory and its CloudFront files.
- Removes the `scripts/` directory and its Cloudflare account helper.
- Keeps the change limited to the requested directory cleanup.

Validation:
- `git diff --check`
- Confirmed both directories are absent after deletion


ElliotEtag:
polly review please

PR #12397, opened 2026-07-13
fix: validate synchronous D1 Tinybird replacement

ElliotEtag:
- Treats multipart `mode=replace` uploads as synchronous Tinybird operations.
- Removes the invalid `/v0/jobs/<datasource-id>` polling that made all five jobs fail after successful replacement in [run 29250993475](https://github.com/pollinations/pollinations/actions/runs/29250993475).
- Validates imported, quarantined, and invalid counts directly from the replacement response without logging response bodies.
- Production verification found every exported row present across all five datasources with zero quarantine rows.
- Validated with actionlint, embedded Bash syntax checking, and `git diff --check`.

### DIFF
--- a/.github/workflows/d1-tinybird-sync.yml
+++ b/.github/workflows/d1-tinybird-sync.yml
@@ -109,46 +109,13 @@ jobs:
             exit 1
           fi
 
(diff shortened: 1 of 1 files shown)


ElliotEtag:
polly please review the synchronous Tinybird response validation.

PR #12400, opened 2026-07-13
fix: validate issue automation verdicts

ElliotEtag:
- Validates Polly responses against the verdict schema requested by the workflow.
- Accepts direct, fenced, and prose-embedded verdict JSON without requiring the nonexistent action field.
- Rejects malformed verdicts before any issue comment, label, or closure is attempted.
- Leaves the backfill workflow, prompt, model, and action thresholds unchanged.

### DIFF
--- a/.github/workflows/issue-automation.yml
+++ b/.github/workflows/issue-automation.yml
@@ -94,6 +94,16 @@ jobs:
             ];
 
             // --- JSON extractor ---
+            function isVerdict(value) {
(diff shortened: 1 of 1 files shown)


ElliotEtag:
- actionlint passes with no findings.
- Exact embedded workflow harness passes direct JSON → comment.
- Fenced JSON → comment + close; prose-wrapped JSON → label; valid no-action → no writes.
- Obsolete action-only shape is rejected; no real issues were modified.

PR #12423, opened 2026-07-13
fix: harden Vast Flux deployment and canary checks

ElliotEtag:
- Runs a pre-created, remotely managed Cloudflare Tunnel from a scoped token file instead of copying an account certificate onto Vast hosts.
- Persists shell-escaped Flux settings, enables high-performance Hugging Face downloads, and ignores the secret-bearing `.env.flux` file.
- Adds a deterministic production-path canary that compares direct Vast output with `gen.pollinations.ai` output so Fireworks cannot mask a broken route.
- Updates the Flux runbook with the active RTX 5090, current listed rate, and canary-before-cutover procedure.

Checks: `bash -n` on both scripts, validation failure-path tests, secret-pattern scan, `git diff --check`.

### DIFF
--- a/image.pollinations.ai/GPU_INSTANCES.md
+++ b/image.pollinations.ai/GPU_INSTANCES.md
@@ -1,39 +1,40 @@
 # GPU Instances
 
-Last updated: 2026-07-02
+Last updated: 2026-07-13
 
 ## Capacity Summary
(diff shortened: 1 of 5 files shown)


ElliotEtag:
polly please review

PR #12427, opened 2026-07-13
fix: add provider and fallback observability

ElliotEtag:
- Adds shared, billing-neutral metadata for the actual model, provider, and fallback state that any generation modality can emit.
- Ensures every Tinybird generation event records `modelProviderUsed` and an explicit `fallbackUsed` value; primary requests use the registry provider, while response metadata can override it.
- Uses Flux → Fireworks as the first fallback integration without making the tracking contract Flux-specific.
- Preserves the strict image usage validation added in #12450; fallback observability cannot invent or default billable usage.

Root cause: generation tracking had no provider-agnostic way for a fallback path to report which provider actually served the response.

Checks: Biome; TypeScript typecheck; 60 targeted Vitest tests passed.


ElliotEtag:
polly please review

ElliotEtag:
polly please re-review the shared fallback telemetry refactor

PR #12428, opened 2026-07-13
fix: reject gpt-image-2 transparent backgrounds

ElliotEtag:
- Rejects `transparent=true` for `gpt-image-2` before calling the provider.
- Keeps transparent backgrounds available for `gptimage` and `gptimage-large`.
- Corrects the source image API schema; generated API documentation is left untouched.
- Adds schema-level and HTTP-level regression coverage, including the client-facing 400 response.

Root cause: OpenAI documents that `gpt-image-2` does not support transparent backgrounds, but the gateway advertised and forwarded the parameter, causing provider-side failures.

Reference: https://developers.openai.com/api/docs/guides/tools-image-generation#tool-options

Checks:
- `npx vitest run test/image/params.test.ts`
- `npx vitest run test/generation-vcr.test.ts -t "gpt-image-2 rejects transparent backgrounds with 400"`
- `npm run typecheck`


ElliotEtag:
polly please review

PR #12433, opened 2026-07-14
fix: restore Klein RunPod build path

ElliotEtag:
- Restores `image.pollinations.ai/klein-runpod` as the stable Klein deployment directory.
- Repairs the existing Klein container workflow by restoring its expected Docker build context.
- Keeps the Vast provisioning script and operational guide aligned with the restored path.
- Leaves the model registry provider as `vast` and does not change production routing or pricing.

Root cause: the Vast migration renamed the established directory without updating the hidden GitHub Actions workflow, causing the post-merge container build to fail because its context no longer existed.

Validation:
- `bash -n image.pollinations.ai/klein-runpod/setup-vast.sh`
- `python3 -m py_compile image.pollinations.ai/klein-runpod/handler.py`
- `git diff --check`
- Confirmed the workflow context, Dockerfile, handler, and requirements files exist


ElliotEtag:
polly please review the restored Klein workflow path and confirm the existing container build context is valid.

PR #12434, opened 2026-07-14
ci: isolate gen production deployment token

ElliotEtag:
- Routes the gen production deployment through a dedicated `CLOUDFLARE_API_TOKEN_GEN` GitHub Actions secret.
- Leaves the shared Cloudflare token and all unrelated deployment workflows unchanged.
- Requires production Cloudflare deployments to run through GitHub Actions, never local Wrangler or OAuth.
- Documents manual workflow dispatch and post-deployment version/binding verification.

Root cause: the shared CI token lacked Workers VPC permission, and replacing it with a narrowly scoped gen token could break unrelated workflows.

Validation:
- GitHub secret presence verified without reading its value.
- `actionlint .github/workflows/deploy-gen-cloudflare.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/deploy-gen-cloudflare.yml
+++ b/.github/workflows/deploy-gen-cloudflare.yml
@@ -41,7 +41,7 @@ jobs:
 
       - name: Run database migrations
         env:
-          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN_MYCELI }}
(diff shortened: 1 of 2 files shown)


ElliotEtag:
polly please review the production Cloudflare CI-only deployment rule.

ElliotEtag:
polly please re-review the dedicated gen deployment token change.

PR #12441, opened 2026-07-14
Fixes #12442: Add 3D Generation to the Model Monitor

### LINKED ISSUE #12442: Add 3D Generation to the Model Monitor
self explanatory

sharktide:


sharktide:
> ## AI Code Review
> ## Review
> **Minor issues:**
> 
> **`apps/model-monitor/src/hooks/useModelMonitor.js:36`** — Array allocation on every call `["video", "3d"]` creates a new array each time `eventTypeForDisplayType` is invoked. If this function is called frequently (e.g., during list rendering), hoist the array to a module-level constant:
> 
> ```js
> const IMAGE_EVENT_TYPES = ["video", "3d"];
> function eventTypeForDisplayType(type) {
>   return IMAGE_EVENT_TYPES.includes(type) ? "image" : type;
> }
> ```
> 
> **`apps/model-monitor/src/App.jsx:40`** — Potential missing handling for "3d" type Adding `"3d"` to `MODEL_TYPES` and mapping it to `"image"` event type may not be sufficient. Without seeing the rest of the codebase, consider whether other areas need updates:
> 
> * **Rendering logic**: Is there a switch/if on type that determines how results are displayed? 3D output (GLB/OBJ files) may need different rendering than images.
> * **API filtering**: If the backend supports a `"3d"` event type, mapping it to `"image"` here could cause 3D events to be missed or misclassified.
> * **Icons/assets**: Is there a type-to-icon mapping that needs a "3d" entry?
> 
> **Question for the author**: Is mapping `"3d"` to `"image"` event type intentional? If 3D generation produces actual 3D assets (not image previews), this mapping could cause display or data-fetching issues. The same pattern works for video (which typically has image thumbnails), but 3D may differ.
> 
> _Requested by `thomash_pollinations` via Discord_
> 
> **From:** `thomash_pollinations` • _via Discord_

yes, 3d is mapped to image, that was a choice made by @tomdacatto in the original PR

voodoohop (approved):
Reviewed the catalog and telemetry path. The 3D route deliberately emits `generate.image`, so mapping the display type to the existing image event type is correct and avoids expanding the Tinybird event schema. The two-item module-level constant is a minimal extension of the existing video rule; no YAGNI or blocking concerns.

PR #12450, opened 2026-07-14
fix: remove Vertex image billing fallbacks

ElliotEtag:
- Removes hardcoded Vertex prompt-token estimates, candidate residual inference, and default output usage.
- Bills image requests only from complete usage metadata reported by Vertex.
- Treats missing, unsupported, or inconsistent usage as one invalid upstream response—without estimation or reconstruction.
- Requires successful image and video responses to emit at least one explicit billable usage header.
- Adds real-response fixtures and a worker-level test asserting exact headers, Tinybird token counts, provider cost, user price, and 502 failure behavior.

Checks:
- `npx biome check --write <changed files>`
- `npm run typecheck`
- `npx vitest run test/image/nanobanana-e2e.test.ts test/image/vertexAIImageGenerator.test.ts test/image/trackingHeaders.test.ts`
- `npx vitest run test/generation-vcr.test.ts`


ElliotEtag:
- polly please review the exact Vertex usage validation and removal of billing fallbacks.

PR #12458, opened 2026-07-14
refactor: use URL routes in Enter dashboard

voodoohop:
- Replaces Enter dashboard hash navigation with TanStack path routes and router links.
- Persists Models, Activity, and Pollen view state in validated URL search parameters.
- Preserves sign-in and Stripe return destinations, metadata, and in-page anchors.
- Updates repository callers to destination-specific dashboard URLs.
- Validates with Enter typecheck/build, direct route probes, the pollinations.ai build, and 66 Stripe integration tests.


ElliotEtag:
- One destination-specific link appears to be missing in `packages/ui`: `AppUserMenu` labels the action **Top up account**, but uses `href={dashboardHref}`. The logged-in `/` redirect still reaches `/pollen`, but it adds an unnecessary load and depends on root redirect behavior. A minimal central fix would be `href={new URL("/pollen", dashboardHref).toString()}`.
- Three new internal links still use ordinary anchor navigation and therefore reload the app: `activity/earnings-graph.tsx` → `/keys`, `activity/usage-section.tsx` → `/keys`, and `quests/quest-overview.tsx` → `/pollen`. These could use the existing `InlineLink as={Link} to="…" external={false}` pattern.
- No broader UI-package routing change looks necessary: `NavItem` already supports `as={Link}`, and `PeriodPicker` is already controlled, so TanStack routing can remain Enter-owned.

PR #12471, opened 2026-07-14
feat: show model health in Enter

voodoohop:
- Adds 24-hour model health to the Enter dashboard.
- Shows token speed for text models and categorical health for all models.
- Keeps the performance column sortable by token speed.


voodoohop:
polly please review

PR #12473, opened 2026-07-14
feat: add community model price floor

voodoohop:
## What

- Adds one shared positive price floor: `0.000001` Pollen per 1M tokens (`1e-12` per token).
- Keeps `0` and blank dashboard prices explicitly free.
- Enforces the floor in the API and dashboard, so CLI and direct callers receive the same server validation.
- Adds API and form boundary tests for free, exact-floor, below-floor, negative, and malformed values.

## Why

Prevents misleading non-zero prices that are effectively too small to be meaningful while preserving free community models.

## Impact

No migration or repricing. Existing rows are unchanged; the floor applies when a price is created or updated.

## Checks

- `npx vitest run src/community-endpoints.test.ts`
- `npx vitest run test/community-endpoint-price-input.test.ts`
- Enter and Gen typechecks
- Biome formatting and diff checks


voodoohop:
polly please review

PR #12483, opened 2026-07-15
fix: harden community monitor cycles and probes

voodoohop:
## Summary
- Cache-bust community-model probes and preserve usage details for billing diagnostics.
- Require controlled, exact-payload comparisons before attributing anomalous token counts.
- Run one fresh Claude Opus 4.8 medium-effort process per cycle under systemd.
- Bound cycles to 45 minutes and remove the obsolete screen watchdog/nudge path.
- Keep Discord credentials in the service environment, not process arguments; remove the shared SSH key from the PR tip.

## Validation
- `node --check probe.mjs`
- `bash -n loop.sh healthcheck.sh`
- `git diff --check`
- `systemd-analyze verify` on the deployed unit
- Live EC2 verification: Opus 4.8 / medium, active service, 45-minute runtime cap, Discord MCP has no `--config` credential argument
- Fresh cache-busted minimax-m3 requests reproduce upstream 502 with `max_tokens: 10`; omitting that field succeeds, so this is request-shape/upstream compatibility rather than a token-count conclusion


voodoohop:
- Superseded by #12621.
- The complete branch history was merged into #12621, including systemd supervision, low-load cache-busted probes, billing diagnostics, secret handling, and healthcheck changes.

PR #12489, opened 2026-07-15
docs: define model rename confirmation contract

ElliotEtag:
- Documents 73 proposed canonical renames against current `main`.
- Preserves 100% of existing public names and aliases.
- Records multiplier, paid status, provider, GPU ownership, primary route, and fallback for every affected model.
- Identifies four Pollinations-operated GPU routes and the Flux Fireworks fallback.
- Requires explicit approval before rename implementation begins.

Validation: 73 complete rows, four GPU rows, no unknown values, `git diff --check`.


ElliotEtag:
- polly please verify the 73-row rename contract, alias coverage, provider/GPU classifications, and response-label policy.

PR #12500, opened 2026-07-15
fix: harden community endpoint URL validation

Marcus-Mok-GH:
## Summary

- Harden community endpoint URL validation against private/link-local upstream hosts.
- Block cloud metadata/link-local IPv4 (`169.254.0.0/16`), unspecified IPv4, private IPv6, IPv6 link-local, and canonicalized IPv4-mapped IPv6 loopback/private hosts.
- Add regression coverage for blocked host forms in the existing community endpoint test file.

## Root cause

Community endpoint registration normalized URLs and rejected obvious private hosts, but the host denylist did not cover link-local/cloud metadata addresses or several IPv6/private host encodings. A user-configured endpoint could therefore point the server-side probe/proxy path at internal metadata or private network hosts.

## Validation

- `npx --no-install biome check --write shared/community-endpoints.ts gen.pollinations.ai/src/community-endpoints.test.ts`
- `npx --no-install vitest run src/community-endpoints.test.ts --testNamePattern="normalizes OpenAI-compatible endpoint URLs"`
- `npx --no-install vitest run src/community-endpoints.test.ts` — 20/20 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+## 2026-07-16
+
+- Hardened community endpoint URL validation to reject link-local, private IPv6, IPv4-mapped loopback, and unspecified IP hosts before proxying upstream requests.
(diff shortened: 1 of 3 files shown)


Marcus-Mok-GH:
polly

PR #12495, opened 2026-07-15
feat: use canonical model names

ElliotEtag:
## Summary

- Uses `author/model` canonical names for 68 confirmed registry entries across text, image, video, audio, and embeddings.
- Keeps every replaced public name as a permanent alias and updates Gen, Enter, SDK, MCP, frontend, economics, tests, and pricing snapshots.
- Preserves provider-facing `modelId`, provider, GPU/self-hosted state, paid status, price multipliers, costs, and fallbacks.
- Leaves grouped pricing/context variants unchanged until dynamic pricing exists: Gemini Search, Sonar high-context, Veo 1080p, Wan 2.7 1080p, Pruna p-video 1080p, and Trellis quality tiers.
- Keeps `modelId` for now; removing it requires a separate Gen routing cleanup for audio, embeddings, and realtime models.

## Model catalog

This is the full catalog after this PR. `Variations` groups registry entries that can become parameters once dynamic pricing/routing is implemented. A dash means the model currently has one registry entry.

### Text

| Model name | Canonical model name | Variations |
|---|---|---|
| GPT-5.4 Nano | `openai/gpt-5.4-nano` | — |
| GPT-5 Nano | `openai/gpt-5-nano` | — |
| GPT-5.4 | `gpt-5.4` | — |
| GPT-5.4 Mini | `gpt-5.4-mini` | — |
| GPT-5.5 | `openai/gpt-5.5` | — |
| ChatGPT 5.6 | `gpt-5.6-sol` | Sol, Terra, Luna |
| Mercury 2 | `inception/mercury-2` | — |
| Qwen3 Coder 30B | `qwen/qwen3-coder-30b-a3b-instruct` | — |
| Mistral Small 3.2 | `mistral-small-3.2` | — |
| Mistral Small 4 | `mistralai/mistral-small-2603` | — |
| GPT Audio Mini | `openai/gpt-audio-mini` | — |
| GPT Audio 1.5 | `openai/gpt-audio-1.5` | — |
| Gemini 3 Flash | `gemini-3-flash` | — |
| Gemini 3.5 Flash | `google/gemini-3.5-flash` | — |
| Gemini 3.1 Flash Lite | `google/gemini-3.1-flash-lite` | — |
| Gemini 2.5 Flash Lite | `google/gemini-2.5-flash-lite` | — |
| DeepSeek V4 Flash | `deepseek/deepseek-v4-flash` | — |
| Gemma 4 26B A4B | `google/gemma-4-26b-a4b-it` | — |
[...]

PR #12525, opened 2026-07-17
refactor: remove modelId from model registry

ElliotEtag:
- Removes the `modelId` field from `ModelDefinition` and all registry entries (text, image, audio, embeddings, 3D, realtime, community). It was unused metadata for most categories and duplicated the registry key for image/video.
- Keeps load-bearing upstream IDs in Gen routing, matching how text routing already works: audio IDs live beside the ElevenLabs and Qwen provider handlers, embedding IDs live in the embedding dispatcher, and Azure deployment names live beside the realtime connector.
- Splits audio routing by provider. ElevenLabs and Qwen maps derive exact key unions, single-model music/SFX IDs stay inside their handlers, and dispatch no longer uses a mixed `Partial` lookup with casts and `undefined` handling.
- Perplexity billing adjustment logs now report the registry model name instead of the upstream ID (`calculateBillingAdjustments` gains a `model` parameter).
- VCR snapshot hashing uses `provider` only; affected Portkey snapshots were re-recorded.
- Drops the `ModelId` union and per-category `*ModelId` type exports; `ModelDefinition` is no longer generic.

Unblocks #12495: the canonical rename kept `modelId` only because this routing cleanup was pending.


voodoohop:
Reviewed locally: typecheck (gen), full grep for surviving `definition.modelId` consumers (none — remaining hits are the community-endpoint DB field), audio dispatch traced (all fallthrough TTS models covered by `AUDIO_PROVIDER_MODEL_IDS`; STT blocked by modality guard), tests green: generation-vcr/realtime/sfx/music (43), community-endpoints (20), enter pricing-data (28).

- Embeddings/realtime maps are total `Record`s → compile-time enforced for new models; audio map is `Partial` → forgetting an entry for a future TTS model is a runtime 500. Acceptable, documented in comment.
- VCR: same-provider deployment changes now need a manual `MODEL_SNAPSHOT_VERSION` bump — documented at shared/test/mocks/vcr.ts:13.

Merging.

PR #12529, opened 2026-07-18
fix: clarify private model usage

voodoohop:
- Clarifies that owners call private models through the Pollinations API with their API key.
- States that private-model calls cost no Pollen while upstream-provider charges may still apply.
- Renames “Public id” to “Pollinations model ID.”

Validation:
- `npx biome check --write` on both changed files


voodoohop:
- polly please review

PR #12539, opened 2026-07-18
fix(mcp): simplify audio generation routing

voodoohop:
- Routes `respondAudio` through chat audio output and `sayText` through `/v1/audio/speech`
- Removes local playback, `play-sound`, model/voice fallbacks, and unused voice instructions
- Lets Gen validate audio models and voices, including custom ElevenLabs voice IDs
- Keeps audio results as MCP audio content blocks

Tests:
- `npx biome check --write src/services/audioService.js`
- `env -u POLLINATIONS_API_KEY npm test` (2/2)
- Focused mocked endpoint/body/response checks

PR #12495 only overlaps some MCP files; neither PR depends on the other, so whichever lands second should sync with main.

### DIFF
--- a/packages/mcp/package-lock.json
+++ b/packages/mcp/package-lock.json
@@ -10,7 +10,6 @@
             "license": "MIT",
             "dependencies": {
                 "@modelcontextprotocol/sdk": "^1.25.1",
-                "play-sound": "^1.1.6",
                 "zod": "^3.24.3"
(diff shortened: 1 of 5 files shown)


voodoohop:
- Superseded by #12546, which includes this audio cleanup in the single comprehensive MCP simplification PR.

PR #12543, opened 2026-07-18
Fixes #12544: Replace InferencePort Sync API with Async API - Fix 524 Trellis Timeouts

### LINKED ISSUE #12544: [Bug] 3D Generation Times out on InferencePort Side with Cloudflare 524
## [Bug] 3D Generation Times out on InferencePort Side with Cloudflare 524

Cause: Current 3D generation uses the InferencePort AI Sync api, which is deprecated and can lead to timeouts.
Solution: Switch to the async api similar to the fal async api for rodin but for inferenceport as documented on https://docs.inferenceport.ai/en/latest/api/p2g-api.html#d-generation

Image

sharktide:
Fix #12544 
## [Bug] 3D Generation Times out on InferencePort Side with Cloudflare 524

Cause: Current 3D generation uses the InferencePort AI Sync api, which is deprecated and can lead to timeouts.
Solution: Switch to the async api similar to the fal async api for rodin but for inferenceport as documented on https://docs.inferenceport.ai/en/latest/api/p2g-api.html#d-generation

Image


sharktide:
@ElliotEtag please review, its urgent

ElliotEtag:
@voodoohop

- Switching away from InferencePort’s deprecated `?sync=true` API is directionally correct, but this PR still keeps Pollinations’ `/3d/{prompt}` request open while polling.
- The current limit is 120 attempts × 5 seconds = 10 minutes, while the production edge allows roughly 180 seconds. Long jobs will still terminate upstream of the Worker.
- I tested an equivalent async submit/poll flow end-to-end against InferencePort: the job was accepted, but no result arrived within 120 seconds; Pollinations returned 504 after approximately 123 seconds.
- A timed-out request can leave the provider job running, creating provider cost without delivered media or successful user billing.
- The provider-side async API therefore improves the transport, but it does not fully solve reliability while Pollinations retains a blocking response contract.
- A complete fix would require public submit/status endpoints and client polling. That is a separate product/API decision because the current requirement is one completed response within 90–120 seconds.

Recommendation: do not merge the current 10-minute polling implementation without deciding whether Pollinations should expose a genuinely asynchronous 3D API.

sharktide:
Superseded by #13071

PR #12545, opened 2026-07-18
refactor(mcp): simplify server internals

voodoohop:
- Removes 217 net lines of duplicated request, error, URL, and startup plumbing
- Uses one authenticated fetch path and one executable MCP entrypoint
- Makes static tool registration fail fast instead of silently dropping tools
- Preserves public tool names, schemas, and successful response payloads

Tests:
- `env -u POLLINATIONS_API_KEY npm test` (2/2, 22 tools)
- `npm pack --dry-run --json`
- Installed-tarball MCP handshake (22 tools)
- Biome and `git diff --check`

PRs #12539 and #12495 only overlap MCP files; there is no semantic dependency. Whichever lands second should sync with main.


voodoohop:
- Superseded by #12546, which includes this plumbing cleanup in the single comprehensive MCP simplification PR.

PR #12546, opened 2026-07-18
refactor(mcp): simplify to direct capability tools

voodoohop:
- Preserves every existing MCP capability: chat, embeddings, image, video, 3D, general audio/music/SFX, transcription, model discovery/status, balance, and usage. Adds dedicated raw TTS and usage tools.
- Replaces only `generateText` with the lower-level `chatCompletion`; no distinct generation or discovery feature is removed.
- Removes MCP-owned model/default/pricing/error policy, response reconstruction, mutable auth tools, media re-uploading, and parallel DTOs. Requests and JSON extensions pass through to Gen.
- Preserves stateless HTTP, request-scoped bearer auth, transcription URL protections, URL-body cancellation, #13672 batch rejection, and the hosted MCP registry link. Bumps `@pollinations/mcp` to 3.0.0 for the breaking API cleanup.
- Net change versus current main: 1,050 fewer lines. Local service, modern/legacy protocol, hosted Worker, packaging, formatting, cancellation, and Cloudflare dry-run checks pass.


Circuit-Overtime (commented):
Review findings:
- transcribeAudio must reject IPv4-mapped IPv6 loopback/private addresses (for example ::ffff:127.0.0.1), not only plain IPv4 and common IPv6 private ranges.
- For output=url, validate the Gen response headers but cancel the response body instead of buffering the complete image/video payload before returning its URL.

Both fixes are covered by focused MCP service tests locally.

> I have made a fix and pushed

PR #12562, opened 2026-07-18
feat: add app access for community models

voodoohop:
- Preserves **Private** as owner/account-only, so existing private models do not gain new callers.
- Adds an explicit **App users** access state for users authenticated through any app owned by the model developer; deleting the app key revokes access.
- Keeps **Public** globally listed and callable. App/Public models use configured per-token pricing, existing community rewards, and existing BYOP markup.
- Applies one access predicate to model discovery and generation, while preserving API-key model scopes and non-disclosure to unrelated callers.
- Updates the dashboard dialog, guidance, and badges. The SQLite column is already plain TEXT, so no D1 migration or sharing table is needed.

Tests:
- `npx vitest run src/community-endpoints.test.ts` (20 passed)
- `npx vitest run src/byop-markup.test.ts` (18 passed)
- `npm run typecheck` (Gen and Enter)
- `npx biome check` on changed files


voodoohop:
polly please review this focused private app model access change.

voodoohop:
polly please review the updated three-state access model, especially private-state preservation, app-key revocation, and app/Public billing.

PR #12589, opened 2026-07-19
feat: harden Z-Image deployment on Vast

ElliotEtag:
- Adds guarded Vast setup with production heartbeat disabled by default
- Pins the verified RTX 5090 stack and works around cuDNN VAE/SPAN crashes while keeping GPU execution
- Adds deterministic tunnel checks, bounded benchmarks, persistent caches, and process supervision
- Documents a shared Cloudflare Tunnel hostname across two Vast replicas

Live checks:
- Both RTX 5090 hosts pass 512x512, 1024x1024, and 768x1152 generation
- Host 1: 332/332 in 5m, 1.095 images/s, p95 3.713s
- Host 2: 69/69 in 60s, 1.090 images/s, p95 3.748s
- Shared tunnel has two healthy replicas; unauthenticated generation returns 403
- Both hosts match direct and tunneled pixels at all three sizes (RMS 0.000)
- Production traffic is reaching both Vast replicas with no observed 5xx/runtime errors

Static checks:
- `bash -n image.pollinations.ai/z-image/setup-vast.sh image.pollinations.ai/z-image/verify-vast.sh`
- `python3 -m py_compile image.pollinations.ai/z-image/server.py image.pollinations.ai/z-image/benchmark-vast.py`
- `git diff --check`

### DIFF
--- a/image.pollinations.ai/z-image/.gitignore
+++ b/image.pollinations.ai/z-image/.gitignore
@@ -5,6 +5,7 @@ venv/
 .venv/
 testImages/
 env/
+.env.zimage
 model_cache/
 gfpgan/
 *.out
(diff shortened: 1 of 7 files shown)


ElliotEtag:
polly please review the Vast deployment hardening, especially the scoped cuDNN workarounds and heartbeat safety gate.

PR #12611, opened 2026-07-20
fix: sub-batch embed requests — Pollinations caps input at 32 items

Itachi-1824:
Second full backfill (after #12609) got to 475/1852 files, 1384 vectors upserted, then died on Pollinations' 32-item input cap. Fixed that, then reviewed the whole embed path properly before pushing again rather than iterating fix-by-fix in CI:

- **32-item batch cap** — \`embed_batch\` now sub-batches at 32 instead of sending a whole file's chunks in one request.
- **Parallelized the embed loop** — was fully sequential (one file at a time), which is most of why 1852 files took so long. Now runs with concurrency=4 via a thread pool. Kept conservative rather than higher since real embeddings-endpoint concurrency hasn't been load-tested against the actual credential.
- **Fixed a silent chunk/embedding misalignment bug** — \`texts\` was filtered (dropping blank chunks) before embedding, but the result was zipped against the *unfiltered* chunk list. Any blank chunk shifted every embedding after it by one position, so a later chunk's embedding could get stored under an earlier chunk's file_path/line-range metadata. Filtering once and reusing the same list fixes it.
- **429-aware retries** — rate limits now back off using \`Retry-After\` instead of being treated like any other error.
- **No more silent partial success** — a run with any per-file failures now exits non-zero instead of reporting success. Re-running is safe (upserts are idempotent by id), so a red CI check just means "run it again," not "some files are silently missing forever."

Index currently has 1384 partial vectors from the failed run — will clear + re-run \`full\` after this merges.

### DIFF
--- a/.github/scripts/embed_code.py
+++ b/.github/scripts/embed_code.py
@@ -261,8 +261,13 @@ def collect_code_files(repo_root: Path) -> list[Path]:
     return files
 
 
-def embed_batch(texts: list[str], retries: int = 3) -> list[list[float]]:
-    """Embed a batch of texts via Pollinations qwen3-embedding-8b at 1536-dim (MRL truncation)."""
+# Pollinations' /v1/embeddings rejects requests with more than this many input items
+# ("Too big: expected array to have <=32 items") — a single large file can produce more
+# chunks than that, so embed_batch must sub-batch instead of sending everything at once.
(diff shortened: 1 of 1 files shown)


Itachi-1824:
polly review and cross check with CF vectorize and ensure everything is correct! Be as concise as possible

PR #12612, opened 2026-07-20
refactor: rename Polly to Polli

Itachi-1824:
Renames Polly to Polli across the repo — branding, docs, workflows, the trigger phrase, and the apps/polly directory.

A few things stayed as-is so nothing outside the repo breaks:

- GitHub secret names, since they're still POLLY_* in the org
- The public `polly` model id and its `polly-ai` alias
- polly.pollinations.ai

The bot now answers to `polli` instead of `polly`.


ElliotEtag (changes requested):
- `polli` is a substring of `pollinations`: both the workflow-level `contains()` gate and `body.includes()` accept ordinary mentions of `pollinations` or `gen.pollinations.ai`. Use a boundary-safe invocation check.
- `repo-triage-new-issues.yml` now requests model `polli`, but `/v1/models`, the registry, and provider config expose only `polly`. Keep the external model ID as `polly` unless the API migration is implemented atomically.
- The renamed Actions secrets are unavailable in every applicable scope: the repo retains `DISCORD_POLLY_*` and `PLN_GITHUB_AUTOPOLLY_KEY`, while the new names are absent; no organization or environment replacements are configured.
- `embed_code.py` defaults to `polli-code-embeddings`, but authenticated Cloudflare inspection shows only `polly-code-embeddings`, and the workflow provides no override.
- The repository has a `polly` label with existing usage, but no `polli` label. The triage and auto-fix workflows would therefore fail when applying the renamed label.
- The VPS rollout is not self-contained: deployment excludes `.env`, `data`, `venv`, PEM files, and extensions, then expects `/root/polli/venv` and `polli.service`; current deployment logs and live health still use Polly naming. Add an explicit migration/rollback path and resolve the current merge conflict.
- Revert the third-party app renames unless their owners migrate them: `nulls-brawl-site/polli-IDE` does not exist, and the submitted app names remain Polly.

ElliotEtag (changes requested):
- `apps/polli/src/api/polli_api.py` still returns `"model": "polli"` and names `polli` in the streaming error, while the registry, gateway config, and live `/v1/models` expose only the preserved external ID `polly`. The gateway propagates the upstream value into the public response and `x-model-used`, so this would return and track an unregistered model. Keep these external-facing values as `polly`; internal branding can remain Polli.
- `apps/polli/src/context/repo_info.txt` likewise tells the assistant that its public API model is `polli`. Lines 19 and 62 must continue to use `polly`, otherwise the bot will recommend a model that is not available.
- `.github/workflows/repo-auto-fix-polli-issues.yml` tells users to add the `polli` label, but its gate and removal logic correctly retain the existing `polly` label. `.github/docs/TRIAGE.md` repeats the nonexistent label. Update the user-facing instructions to `polly`.
- Minor cleanup: `.github/scripts/embed_code.py` still cites the removed `apps/polly` path at lines 8 and 513.

ElliotEtag (approved):

PR #12372, opened 2026-07-13
feat: relax community monitor deactivation threshold

voodoohop:
## Summary
- Rule (b) floor 85% → 80%, requires 3 consecutive degraded cycles (was 2)
- No more immediate deactivation on rule (b) — pings the owner first, waits ~2h for recovery, only deactivates if still failing
- Adds `degradedStreak` and `pendingDeactivation` to state.json

## Test plan
- [x] Deployed CYCLE.md to the EC2 monitor box (scp, no restart needed — re-read fresh each cycle)
- [ ] Observe next few cycles for correct ping → grace-period → deactivate behavior

PR #12621, opened 2026-07-21
fix: consolidate community monitor reliability and safety

voodoohop:
- Moves the community monitor to systemd-supervised fresh processes and keeps its committed prompt/runtime files synchronized from `main`.
- Adds low-load text and image probes, provider-origin image 4xx classification, bounded billing diagnostics, official-model follow-ups, and daily leaderboard handling.
- Adds a deterministic daily reliability audit: 7-day success below 70% identifies candidates, while current success of at least 80% in the freshest adequately sampled 24h/48h window prevents delayed deactivation after a fix or new fallback.
- Counts successful fallback rescues as successes and requires a fresh provider-failing probe when both recent windows contain fewer than 20 attributable requests.
- Keeps the acute 4-hour incident path separate with its 24-hour grace and limited quota-reset reactivation.

Validation:
- `npx biome check apps/operation/community-monitor/seven-day-health.mjs`
- `node --check apps/operation/community-monitor/seven-day-health.mjs`
- `git diff --check`
- Production-data dry run: 9 current candidates, 4 protected by recent health, 5 low-volume rows requiring a probe; no model state changed.

Supersedes #12372 and #12483. The live EC2 runtime is updated from this branch; automatic repo synchronization remains gated on this PR landing in `main`.


voodoohop:
- Reconciled this branch with current main; the leaderboard now keeps both generation_event_v2 and the community:leaderboard upload tag.\n- Adds a gated pre-cycle updater that refreshes only CYCLE.md, probe/loop/health scripts, and the leaderboard builder from origin/main.\n- The gate remains inactive until this updater lands on main, so the deployed unit cannot downgrade live files while the PR is open.\n- Installed and checksum-verified the updater/unit on the monitor without restarting it; the service remains active.\n- Validation: shell syntax, Node syntax, and git diff --check pass.\n\npolly please review.

PR #12336, opened 2026-07-10
feat: add editable prompt agents

voodoohop:
Depends on #13407 being deployed before Pollinations tools are enabled.

- Adds editable prompt agents with a system prompt, Pollinations base model, and public Streamable HTTP MCP servers.
- Runs every definition in one shared Enter Worker using Vercel AI SDK ToolLoopAgent and the official @ai-sdk/mcp client.
- Adds an opt-in Pollinations tools switch backed by `https://mcp.pollinations.ai`; external MCP servers never receive Pollinations credentials.
- Mints a short-lived ag_ token scoped to the caller parent key and managed agent, then uses it for base-model and built-in tool calls.
- Keeps agent definitions editable behind stable private/public My Models listings and requires managed-agent listings to remain free.
- Adds dashboard and polli agents CRUD without per-agent Workers, bundles, or persistent agent API keys.

V1 stays narrow: one agent per listing, no MCP OAuth/private headers, custom code, or versioning.

Validation: Enter and Gen typechecks, CLI build, 15 prompt-agent runtime tests, 7 run-token tests, and 6 focused managed-agent registration/delegation tests.

PR #12626, opened 2026-07-21
fix(mcp): return bearer authentication challenge

Circuit-Overtime:
- Stacks on #13111, which owns the stateless Streamable HTTP MCP Worker and focused tool surface.
- Keeps authentication bearer-first with request-scoped `sk_` keys; OAuth remains deferred.
- Adds a standard `WWW-Authenticate: Bearer` challenge for unauthorized requests.
- Prevents caching of authentication failures with `Cache-Control: no-store`.
- Covers the challenge headers in the Worker tests.

Validation: MCP Worker tests and Wrangler dry-run bundle pass.


voodoohop:
Decision from the weekly call on 2 Aug: **ship Streamable HTTP with `sk_` bearer auth first, defer OAuth.**

Rationale given: doing OAuth badly is worse than doing it properly later, and the auth work was blocking the transport from existing at all.

@Circuit-Overtime — PR #12626 already implements the full OAuth path here (protected-resource metadata, audience-bound tokens, Client ID Metadata Documents). Don't rebuild it from scratch: either descope that PR to bearer-only and keep the OAuth work on a branch, or argue that it's close enough to merge as-is and skip the bearer detour.

Also check for overlap before starting: PR #12336 ships public Streamable HTTP MCP servers for prompt agents. Two implementations of the same transport would be worth catching now.

voodoohop:
- The Bearer challenge and test were folded into #13111 in c83f178271 with co-author attribution to @Circuit-Overtime.
- #13111 now accepts and forwards any non-empty Bearer credential; key policy remains in the API gateway.
- Closing this stacked PR as superseded.

Circuit-Overtime:
@voodoohop ok cool, thank you Thom!

PR #12631, opened 2026-07-22
feat: costVariants — usage/request-conditional model pricing

voodoohop:
Usage- and request-conditional pricing: models can declare named alternate rate sheets instead of being split into one model per price tier.

## Cheapest tier by default — opt in to 1080p

Video models default to their lowest resolution so an unqualified request is the cheap one; 1080p is charged only when explicitly asked for via `?resolution=1080p`.

⚠️ **This is a breaking change for existing callers.** `veo-1080p` / `veo-3.1-fast-1080p` / `veo-1080`, `wan-pro-1080p` / `wan2.7-1080p` / `wan-pro-1080`, and `p-video-1080p` / `pruna-video-1080p` stop being separate models and become aliases of `veo` / `wan-pro` / `p-video`. They serve and bill 720p ($0.08/$0.10/$0.02 per sec) instead of 1080p ($0.10/$0.15/$0.04) unless the parameter is passed. Billing matches what is served in both cases; the change is output resolution, not price accuracy.

`drizzle/0043_canonicalize-model-permissions.sql` rewrites stored per-model permissions that referenced the removed model ids.

## Mechanism

- `costVariants` — named partial rate sheets merged over `cost`, selected once at billing time in `calculateUsageBilling`. Selectors return a **name**, never a number: selection is code, money stays data. Unknown or throwing selectors warn and bill base rates; billing never throws.
- Helpers live in `shared/registry/cost-variants.ts` (`defineCostVariants`, `longContextAbove`, `matchResolution`) — a separate module because importing the service maps into `registry.ts` creates a module-evaluation cycle that silently empties the registry in the worker bundle.
- Under a fallback, cost follows the served model and price follows the quoted listing; the variant identity follows the quote (it must reproduce `dev_price`), while `costVariantStatus` reports a selector failure from **either** side.

## Rates

- Long context, whole-request repricing above the threshold: `gpt-5.4`, `openai-large`, `midijourney-large`, `gpt-5.6-sol/terra/luna` (272K); `gemini-large`, `grok-4.5` (200K); `qwen-large` (256K); `qwen3.7-flash` (two tiers, 32K and 256K).
- Resolution variants: `veo`, `wan-pro`, `p-video` (1080p), `seedance-pro` (480p and 1080p).
- `qwen-image` bills t2i at its real $0.025 with a $0.03 `edit` variant on reference-image presence, replacing a flat $0.03 that over-charged generation.
- `resolution` is validated against each model's declared `resolutions` — unsupported values return 400 rather than falling back silently.

## Telemetry — needs a Tinybird deploy before gen ships

Four columns on **both** `generation_event` and `generation_event_v2`: `cost_variant`, `cost_variant_status`, `provider_reported_cost`, `provider_cost_delta`. No CI automation — deploy to staging and prod manually per `AGENTS.md`.

- `token_price_*` now emits the effective merged sheet resolved at billing time, so recorded rates reproduce the billed total instead of the request-time base sheet.
- OpenRouter reports its own billed amount via `x-provider-reported-cost`; the delta against our calculation is recorded and warned on above 1%. Reconciliation only — it never changes user billing.
- `x-usage-missing` lets flat provider fees (Perplexity's per-request search fee) still bill when token usage is absent, without fabricating token charges.
- `/models` publishes `pricing_variants`, and the dashboard catalog shows conditional prices — restoring the tier visibility the model collapse would otherwise remove.

## Tests

New CI step runs the billing registry suites on every PR touching Enter. Coverage includes variant boundaries per model, `quotedBy != servedBy` crossed with variant selection on both sides, and an e2e path through the real image handler asserting provider payload, variant-priced event, and deducted amount.

**Known gap:** no test covers the legacy alias path, only the explicit `resolution` parameter — which is why the downgrade above is invisible to the suite.

Closes #11683, closes #11721 (superseded designs).


ElliotEtag:
**Verdict: changes required** — reviewed at head [73ec39d](https://github.com/pollinations/pollinations/commit/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf).

The billing core is sound: variant selection at the single billing site, the multiplier applied exactly once across token lines and adjustments, one rounding at the ledger boundary, and search/request adjustments plus Perplexity handling unchanged (`perplexity-billing.ts` is byte-identical to main). What blocks the merge is what's bundled around it: the model-identity collapse breaks existing API keys, billing can't see the provider route it's pricing, variant selection is invisible to telemetry, and the PR edits generated docs. Focused billing tests 21/21 and changed gen tests 40/40 pass locally; CI is green. External rates re-verified against provider pages/APIs on 2026-07-22 (links inline).

## Blockers

**1. The registry collapse breaks existing API keys.** [auth.ts#L71-L81](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/gen.pollinations.ai/src/middleware/auth.ts#L71-L81) checks `apiKey.permissions.models.includes(model.resolved)` — stored strings against the resolved canonical id. Nothing alias-normalizes stored permissions: [api-key.ts#L273-L291](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/shared/auth/api-key.ts#L273-L291) only filters non-strings, key creation stores `allowedModels` verbatim, and no migration rewrites them. So a key scoped to `veo-1080p`, `wan-pro-1080p`, `p-video-1080p`, or `p-video-720p` starts getting 403 on the exact name it was scoped to once those become aliases. The reverse also holds: a key scoped to `veo` could never execute 1080p before and now can, at the higher rate. [visible-model-ids.ts#L45-L57](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/shared/registry/visible-model-ids.ts#L45-L57) will additionally hide those entries from key listings. Needs a permissions migration (or alias-normalization at check time) plus tests with legacy allowlists — and it argues for splitting the identity collapse out entirely (see Scope).

**2. Variant selection is invisible to billing records.** Per the PR body this is deliberate ("billing selection never throws"): a throwing or unknown selector `console.warn`s and bills base ([registry.ts#L217-L245](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/shared/registry/registry.ts#L217-L245)), returning `costVariant: undefined` — indistinguishable from "no variant applies". Track doesn't even destructure `costVariant` ([track.ts#L518-L527](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/gen.pollinations.ai/src/middleware/track.ts#L518-L527)); it has zero consumers repo-wide and no Tinybird column. The effective `token_price_*` sheet is recorded, so today's variant is inferable from rates — but after any registry change you can't reconstruct *why* a sheet was selected, and a selector regression bills base silently, undercharging long-context/1080p/edit (or overcharging seedance 480p). Needs a structured result persisted to the event: variant/rule id, selection status, failure reason.

**3. Billing can't see the provider route.** `PricingInput` is `{resolution?, hasImage?}` ([cost-variants.ts#L17-L26](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/shared/registry/cost-variants.ts#L17-L26); usage arrives separately) — no provider, endpoint, deployment, region, or OpenRouter tag/execution class. Concretely: OpenRouter lists grok-4.5 under [four execution classes](https://openrouter.ai/api/v1/models/x-ai/grok-4.5/endpoints) (standard/priority/ZDR/ZDR-priority) at a 2× rate spread, with an advertised further 2× override from 200K prompt tokens, and the event records only the *configured* registry provider while reducing the `x-fallback-target` header to a boolean ([track.ts#L553-L558](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/gen.pollinations.ai/src/middleware/track.ts#L553-L558)). Accurate OpenRouter/Azure billing is unimplementable until actual routing facts flow into billing and telemetry.

**4. `APIDOCS.md` is edited directly.** It's generated from the live OpenAPI schema after deploy by `docs-regenerate-api-reference.yml`; drop it from the diff, keep the schema/route changes.

## Scope: identity changes riding along with billing

The collapse isn't just billing plumbing:

- Former canonical ids disappear as independent `/models` rows; `p-video`'s canonical id flips (it was an alias of `p-video-720p` on main).
- `modelUsed` (the `x-model-used` header + Tinybird column) collapses `veo-1080p`→`veo`, `wan-pro-1080p`→`wan-pro`, `p-video-${res}`→`p-video`; resolution is recoverable only from the price columns. Analytics and consumers see new identities.
- seedance-pro gains genuinely new 480p/1080p *execution* (main's handler was locked to 720p) — a product change, not a billing one.
- Explicit `resolution` beats an alias' implied resolution, so `model=veo-1080p&resolution=720p` returns 720p ([handler.ts#L390-L407](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/gen.pollinations.ai/src/image/handler.ts#L390-L407)). Since the param didn't exist before, requests that don't send it behave identically — the precedence is defensible, but the PR body promises legacy names "impl[y] their resolution", so state the conflict rule explicitly.
- [legacyResolutionAliases.test.ts](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/gen.pollinations.ai/test/image/legacyResolutionAliases.test.ts) asserts only that each `-1080p`-suffixed alias' owner declares 1080p — not that legacy default/payload/billing behavior is preserved.

Recommendation: land the billing machinery against today's identities; do the collapse as its own PR with a permissions migration and compat tests.

## Rates: corrections and cost-accounting issues

One principle up front: registry `cost` is supposed to be the provider's actual cost — a deliberate subsidy is product pricing and belongs on the price side. A registry comment doesn't make an understated cost sheet correct. And today the schema can't express the two independently: user price is derived from the effective cost sheet through the model's single `priceMultiplier`, so recording a true higher cost necessarily raises the user price. Several entries below understate cost to keep price flat — a product-pricing decision currently expressed as inaccurate cost accounting.

- **grok-4.5 cached input is wrong**: registry has $0.50/M, which matches no advertised [OpenRouter execution class](https://openrouter.ai/api/v1/models/x-ai/grok-4.5/endpoints) today (standard/ZDR $0.30/M, priority/ZDR-priority $0.60/M). The fix is not a blind $0.50→$0.30 swap: capture or pin the execution class, then apply that class's rate. The metadata also advertises a 2× override from `min_prompt_tokens: 200000` — currently absorbed, i.e. understated cost above 200K; confirm inclusivity and actual billing against billed requests.
- **deepseek-pro cached input**: registry $0.14/M vs [Fireworks](https://fireworks.ai/models/fireworks/deepseek-v4-pro) published $0.145/M.
- **qwen-large (qwen3.7-plus)**: OpenRouter currently advertises a 3× override on all four rates at [`min_prompt_tokens: 256000`](https://openrouter.ai/api/v1/models/qwen/qwen3.7-plus/endpoints); the registry absorbs it, so recorded cost is understated ~3× above 256K. Confirm inclusivity and actual billing with billed requests (255,999 / 256,000 / 256,001), then record the true provider cost; if the flat user price should survive that, it needs a price-side mechanism that doesn't exist yet.
- **wan-pro 1080p**: flat $0.15/s for both t2v and i2v — carried from main and deliberate ("charges the single higher rate … never under-bills"). Replicate: [t2v](https://replicate.com/wan-video/wan-2.7-t2v) is flat $0.10/s at *any* resolution; [i2v](https://replicate.com/wan-video/wan-2.7-i2v) is $0.10/720p, $0.15/1080p. So $0.15/s for 1080p t2v remains an inaccurate provider cost and a user overcharge — pre-existing and intentional. `PricingInput.hasImage` now provides the request fact needed for an exact split; update the selector/rate sheet, public metadata, and tests here, or track that work explicitly.
- **gpt-5.6 family**: no long-context variants — deferred in the PR body pending an Azure invoice, but OpenAI now documents ">272K → 2× input / 1.5× output for the full request" on the [5.6 pages](https://developers.openai.com/api/docs/models/gpt-5.6-sol). Given the published rule, the gap is likely mispricing real long-context traffic today. Resolve it with Azure deployment/SKU evidence; until that exists, the deferral should be structured and monitored (calculated-vs-invoice alert on those deployments), not silent.
- **gpt-5.5 comment**: [OpenAI's model page](https://developers.openai.com/api/docs/models/gpt-5.5) now states >272K for 5.5 directly, so update the "presumed from meter symmetry" comment — keep the Azure-meter caveat, drop the pure-symmetry framing. Note OpenAI's rule sentence names input and output only; doubling *cached* input (as the 5.4/5.5 sheets do) is our reading of the Azure meters and worth flagging in the comment.

## Pre-existing silent-billing gaps

Not introduced here, but this PR is where billing fidelity is being built:

- Missing usage → `log.error`, event emitted with `isBilledUsage=false`, generation completes unbilled ([track.ts#L518-L523](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/gen.pollinations.ai/src/middleware/track.ts#L518-L523)). The flag is queryable but nothing monitors or reconciles it.
- A reported usage field with no registry rate warns and bills 0 for that line ([registry.ts#L173-L202](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/shared/registry/registry.ts#L173-L202)).
- Streaming usage is parsed through [openai.ts#L398-L431](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/shared/schemas/openai.ts#L398-L431), which strips nested `prompt_tokens_details.cache_write_tokens`, `.video_tokens`, and `completion_tokens_details.image_tokens`; the non-streaming path passes raw usage through and loses nothing.
- Inclusive-convention usage rows get cached/cache-write/audio/image/video capped in fixed order into disjoint buckets ([usage-headers.ts#L141-L226](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/shared/registry/usage-headers.ts#L141-L226)); raw overlapping values don't survive into billing or telemetry.
- Gemini tool-use prompt tokens: `toolUsePromptTokenCount` appears nowhere in the repo — text usage arrives pre-transformed by Portkey, and the Vertex image mapper 502s on unaccounted tokens rather than merging them. Needs a check before Gemini tool billing (#12651). Gemini code execution has [no separate execution fee](https://ai.google.dev/gemini-api/docs/code-execution#billing) — but its intermediate/tool tokens are billed as input tokens and must be preserved; that statement covers Gemini only, not other providers' code-execution products.

## Coverage

| Case | Status | Note |
|---|---|---|
| gpt-5.4 Azure | OK for text/cached; evidence gaps | strict >272K whole-request; tested at 272,000/272,001 — no just-below case or real provider e2e. The selector counts image tokens, but this model's sheet has no `promptImageTokens` rate; confirm from a sanitized Azure vision usage object whether those tokens are separated or folded into `prompt_tokens`. |
| gpt-5.5 / openai-large / midijourney-large | OK for text/cached; evidence gaps | same sheet + threshold; update the comment per Rates. The same Azure multimodal-usage question applies. |
| gpt-5.6 family | Missing; evidence/monitoring required | author-deferred; OpenAI documents the rule, but the deployed Azure SKU still needs confirmation. Until resolved, make the deferral observable rather than silently recording base cost. |
| gemini-large (3.1 Pro) | OK, gaps | strict >200K with cached + modalities summed toward threshold ([Google pricing](https://cloud.google.com/gemini-enterprise-agent-platform/generative-ai/pricing) whole-request repricing); cache *storage* correctly stays a separate untiered $4.50/M-token-hr adjustment; no just-below case. |
| qwen3.7-plus / OpenRouter | Deferred; provider cost remains understated | 3× override advertised at 256K; needs endpoint pinning + billed-boundary evidence first (Blocker 3). Documented absorption is product policy, not accurate provider-cost accounting. |
| grok-4.5 / OpenRouter | Deferred + wrong rate | execution classes unrepresentable (Blocker 3); cached $0.50 matches no advertised class — pin/capture the class, then apply its rate. |
| grok-4.3 / Azure | Deferred; evidence and safeguards missing | no deployment/meter attribution; add a calculated-vs-invoice alert rather than treating base cost as reconciled. |
| qwen-image | OK, gaps | $0.025 t2i / $0.03 edit on `hasImage`; the predicate is computed independently in routing and pricing (can't diverge today, unguarded); no real-handler billing test. |
| wan-pro 1080p | Carried provider-cost mismatch/overcharge | per Rates — `hasImage` supplies the request fact needed for the exact t2v/i2v split, with corresponding metadata and tests. |
| resolution variants | Partially E2E-tested; identity fallout | true e2e (handler → payload → event → deduction) exists for the p-video alias only; permissions/analytics fallout unaddressed (Blocker 1, Scope). |
| deepseek-pro | Stale rate | cached $0.14 → $0.145. |
| perplexity | OK, tested | request-cost extraction, static fallback, malformed handling, and 10× clamp have unit coverage plus streaming/non-streaming e2e coverage. |

## Tests

CI first: the PR job named "Check formatting and linting" *does* run the full gen suite + gen typecheck (gen/ and shared/ changed) and typechecks changed enter files — but **no workflow in the repo runs enter's vitest suite**, so `cost-variants.test.ts` and the effective-prices snapshot never execute in CI. Billing-core tests need an explicit gate.

Missing in this PR:

- Just-below-threshold cases: 271,999 and 199,999 (272,000/272,001 and 200,000/200,001 are covered).
- A variant sheet with an explicit-zero override — also relevant because public serialization drops non-positive rates ([model-info.ts#L107-L111](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/shared/registry/model-info.ts#L107-L111)), so an explicitly-free variant rate would vanish from `/models` while billing correctly.
- `priceMultiplier ≠ 1` covering tokens *and* adjustments together (the multiplier test is tokens-only; every adjustment test uses 1× models).
- Legacy allowlist permission cases (Blocker 1); legacy alias + conflicting explicit `resolution`.
- qwen t2i vs edit through the real handler; wan t2v-vs-i2v once split.

Selector-fallback tests already exist ("unknown variant name warns and bills base rates", "throwing selector warns and bills base rates") — they'd be reshaped by Blocker 2's structured result, not added from scratch. Follow-up hardening (pre-existing scope): streaming nested-usage fields, overlapping cache read/write preservation, provider error/retry/fallback events, cache-hit no-double-deduct, ledger rounding at the 1e-8 boundary.

## Telemetry

The event records requested/resolved/used model, configured provider, normalized token columns, the effective merged `token_price_*` sheet, per-rule adjustment costs/units (rule keys embed versions, e.g. `google.gemini_3.search_query.v1`), cost/price totals, `markupRate`, and a fallback boolean. It does not record: the selected variant name or selection status (Blocker 2), an overall pricing-rule version, the actual provider/endpoint/deployment/region/OpenRouter class, the fallback *target* (header value discarded), a provider request/invoice id, raw pre-normalization usage, the model's own `priceMultiplier` (derivable, not recorded), or any reconciliation state for `isBilledUsage=false` rows. The effective sheet explains *what* was billed but not *why* once registry rules change.

## Public surface

- `/models` and every native `*/models` route now publish `resolutions` + `pricing_variants` (effective merged sheets); `/v1/models` is untouched — consistent with #12548 owning that surface, but the body's "kept private here" reads as contradicting the visible `/models` exposure; clarify the wording.
- Variant *conditions* aren't published: consumers can't learn that `long_context` means >272K or `edit` means reference-image-present (resolution variants are self-describing; the others aren't). Coordinate the metadata format with #12548 rather than defining it twice.
- The enter dashboard reads only `pricing` ([model-catalog.ts#L18-L39](https://github.com/pollinations/pollinations/blob/73ec39dc7fda3bfc59dafffd6c0b994cfe874dcf/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts#L18-L39)); `pricing_variants`/`resolutions` have zero references in the frontend, so users see base price only for every variant model. Needs a follow-up.

## Required changes

1. Split the identity collapse (aliases, `/models` rows, `modelUsed`, seedance unlock) into its own PR with a permissions migration + compat tests; keep this PR billing-only.
2. Drop the `APIDOCS.md` edit.
3. Persist a structured selection result (variant/rule id, status, failure reason) — stop discarding `costVariant`.
4. Feed actual routing facts into billing + telemetry (provider, endpoint, deployment/region or OpenRouter tag/class, fallback target).
5. Rate fixes: grok-4.5 cached — pin or capture the execution class, then apply its advertised rate ($0.50 matches none today); deepseek-pro cached $0.145; wan t2v/i2v split via `hasImage`; gpt-5.6 — Azure evidence or a structured, monitored deferral (per Rates).
6. CI gate for the enter billing tests; add the missing tests above.
7. Update the gpt-5.5 "presumed" comment; flag the cached-input-doubling assumption on 5.4/5.5.

## Evidence still needed

Azure SKU/meter (and processing tier) for the deployed gpt-5.5/gpt-5.6/grok-4.3 deployments; one sanitized Azure GPT vision usage object to establish whether image tokens are separated from or folded into `prompt_tokens`; OpenRouter billed inclusivity at `min_prompt_tokens` (billed requests at 255,999/256,000/256,001) plus per-request provider tag/class capture; provider invoices for representative long/cached/tool/fallback requests; provider semantics when cache-read and cache-write counts overlap; whether `resolution` participates in cache identity on every route; Gemini `toolUsePromptTokenCount` handling before #12651.

The PR head reviewed here is unchanged at `73ec39d`, but it is now behind `main`; re-check the final diff and tests after rebase.

PR #12651, opened 2026-07-22
feat: route Google text models through OpenRouter

ElliotEtag:
- Routes all eight Google text services through exact OpenRouter Google Vertex endpoints with `allow_fallbacks: false`; upgrades `gemini` and `gemini-search-large` from Gemini 3.5 Flash to 3.6 Flash while preserving old aliases.
- Preserves function tools, multimodal input, streaming, reasoning, and Google Search. Public and legacy function-shaped `google_search` tools are translated to OpenRouter native search; Gemini 2.5 drops `logit_bias` only on search requests. Server-side code execution is intentionally removed by product decision.
- Makes `gemini-fast` the Quest Pollen option (`paidOnly: false`); the other seven remain paid-only. All retain multiplier `1`.
- Bills reported image/video inputs, OpenRouter native search at $0.014/request, and five-minute cache storage at the Flash/Pro rates. Long-context tiering and rare usage-less `max_tokens=1` responses remain explicitly assigned to the incoming pricing-system PR.
- Validated with Biome, both typechecks, all 629 Gen tests, 276 focused Enter model/pricing tests, and local E2E for every service: text, tools, native search, search billing, image/audio/video, and streaming. Full Enter: 549 pass; 63 unrelated Stripe tests require unavailable Stripe secrets.


ElliotEtag:
polli

PR #12656, opened 2026-07-22
fix: cancel timed-out Replicate predictions

ElliotEtag:
Replicate predictions could continue running—and consuming provider resources—after our local polling timeout returned a 504.

- Sends `Cancel-After: 6m` when creating a prediction, covering the 60-second synchronous wait and five-minute polling budget.
- Cancels a still-running prediction when the local polling budget is exhausted, using Replicate’s supplied cancellation URL or the standard fallback endpoint.
- Bounds best-effort cancellation to five seconds and preserves the original 504 if cancellation fails or never settles; the create-time deadline remains the cleanup safety net.
- Covers the deadline header, custom and fallback cancellation URLs, cancellation failures, and a never-settling cancellation request.

Tests:
- `npx vitest run test/image/replicateClient.test.ts`
- `npx vitest run test/image`
- `npm run typecheck`
- `npx biome check src/image/utils/replicateClient.ts test/image/replicateClient.test.ts`


ElliotEtag:
polli

ElliotEtag:
polli

PR #12683, opened 2026-07-23
docs: point generation-event references at v2

voodoohop:
Follow-up to the v2 cutover (#12640/#12641). Complements #12674, which covers the Grafana dashboards and Tinybird pipes — no file overlap with that PR.

- Fixes 4 `model-debugging` scripts that still queried the frozen `generation_event`. These returned zero rows for any window after 2026-07-22 22:28 UTC — the only functional bug here.
- Updates skill queries and prose: `abuse-detection`, `spending-analysis`, `model-debugging`, `model-management`, `monitor-services`.
- Fixes KPI tooltips and economics connector docs that named the old datasource.

Kept the archived table where it is still correct: pre-v2 historical BytePlus lookups in `bytedance.md`, and the `tinybird-deploy` tree listing (both files exist, now labelled).

Verified: `abuse-detection` already excludes anonymous users in every query, so v2 is a safe swap. Ran the updated `find-500-errors` query against live v2 — returns current rows. Left `enter.pollinations.ai/wrangler.toml` alone: it only reads `.origin` off that URL, so the datasource name there is inert.

### DIFF
--- a/.claude/skills/abuse-detection/SKILL.md
+++ b/.claude/skills/abuse-detection/SKILL.md
@@ -12,7 +12,7 @@ description: Detect and analyze abusive accounts on Pollinations. IP clustering,
 **Tinybird query pattern:**
 ```bash
 cd enter.pollinations.ai/observability
-tb --cloud sql "SELECT ... FROM generation_event ..."
+tb --cloud sql "SELECT ... FROM generation_event_v2 ..."
 ```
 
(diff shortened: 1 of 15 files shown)


voodoohop:
Folded into #12674 — same v2 cutover cleanup, easier to review as one PR.

The skills/ half was already covered by 8e7fe060a on that branch (better version — dropped stale line numbers, corrected cache-hit wording). Carried over only the 5 files it didn't touch: KPI tooltips, economics connectors, GpuTab comment.

PR #12705, opened 2026-07-23
[QUEST] Default Grok Imagine to 2K

smplstuff:
Would this work?

(It seems ai also formatted the code a bit)


ElliotEtag:
Closing because the exact Grok Imagine 2K default is already implemented on current main, so this PR is obsolete.

PR #12707, opened 2026-07-23
Move Veo 3.1 Fast to OpenRouter

ElliotEtag:
## Summary

- Routes `veo` and `veo-1080p` through OpenRouter's exact `google/veo-3.1-fast` endpoint.
- Preserves names, aliases, prices, paid-only status, 720p/1080p, 4/6/8-second durations, first/last frames, optional audio, and person generation.
- Passes `personGeneration: allow_all` to Google Vertex through OpenRouter and maps terminal policy failures to client-facing 400 responses.
- Uses no fallback, secret change, deployment, or public capability change.

## Verification

- Paid local E2E covers text/no-audio, two-frame audio, portrait alias, malformed input, billing, cache, and MP4/AAC metadata.
- Latest people-generation probe returned a valid 4-second 1280×720 H.264 video in 72.6 seconds, billed exactly 0.32 Pollen; the repeat was a byte-identical cache hit in 47 ms.
- Gen: 646 tests; focused Enter registry/pricing: 290 tests; both workers type-check; Biome clean.

## Draft gate

Latency remains the blocker: one direct 1080p frames/audio run took 129.6 seconds, above the 120-second target. Keep draft pending a later soak. The new three-minute hard ceiling is intentional; the old loop's “3-minute” comment actually allowed roughly 40 minutes.


ElliotEtag:
Closing after live reconciliation: Pollinations already serves the exact Veo 3.1 Fast checkpoint directly through Google Vertex at the same current pricing. The OpenRouter move adds no capability or cost benefit, and the tested 1080p path exceeded the normal latency limit.

PR #12729, opened 2026-07-24
feat: launch established GitHub quest

voodoohop:
- Launches `github_established` for GitHub accounts at least 730 days old with a 3 Pollen reward.
- Keys completion to the GitHub identity, preventing the same GitHub account from earning through multiple Pollinations accounts.
- Keeps `early_adopter` visible as `coming_soon`, named “Early adopter,” with a planned nine-month Pollinations account threshold and 2 Pollen reward.
- Bumps the quest catalog cache key so the state change is served immediately.

Checks: `npx vitest run test/quests.test.ts --testNamePattern="catalog returns quest definitions|nine-month Early Adopter|github established-account quest"` (5 passed)


voodoohop:
- Superseded by #13303.
- All commits from this branch are merged into the retained PR.
- Combined quest suite passes: 30/30.

PR #12749, opened 2026-07-25
feat: separate community model title from description

voodoohop:
Replaces the data-migration approach in #12692.

- Adds a nullable `title` column (one-line migration, no curated data).
- Required on create, optional on update; 42-char title, 160-char description.
- Read paths go through `communityEndpointTitle()`, which falls back to
  description → model slug. `/models` output is unchanged for every existing
  row, so nothing has to be backfilled before this ships.
- Shows the title on model cards and in `polli my-models list`.

Why not the #12692 migration: it hardcoded 86 `WHERE name = … AND description = …`
pairs — a snapshot of D1 on the day it was written. Rows added before merge would
silently get `substr(name, 1, 42)` (the raw slug), staging/local produce different
results, and `name` is not unique across owners. Backfilling existing titles is a
one-time D1 edit, done separately.

Validation:
- `gen` typecheck + `enter` typecheck clean
- `community-endpoints.test.ts` 28 passed, `tracking-observability.test.ts` 16 passed
- biome check clean

### DIFF
--- a/enter.pollinations.ai/drizzle/0038_slimy_ultimo.sql
+++ b/enter.pollinations.ai/drizzle/0038_slimy_ultimo.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `community_endpoint` ADD `title` text;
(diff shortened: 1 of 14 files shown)


voodoohop:
Folded into #12692 — same content, kept on the original branch.

PR #12807, opened 2026-07-27
fix: use defined price instead of Tinybird estimate for flat-rate models

### LINKED ISSUE #12747: Pre-flight balance check uses Tinybird avg_cost_usd instead of model-defined pricing — blocks flat-rate community models with inflated estimates
## Bug

`getEstimatedPrice()` in `gen.pollinations.ai/src/utils/model-stats.ts` estimates generation cost using **Tinybird historical averages** (`avg_cost_usd`), ignoring the model definition's actual pricing fields. For flat-rate community models, this produces wildly inflated estimates that block users with a 402 even when their balance is sufficient.

## Reproduction

User `kanga0744-art` (Indonesia, SIN edge) attempted to generate with community model `vendouple/zimage` (defined price: `completionImagePrice: 0.01` pollen/image) with an API key balance of 5.0 pollen.

**Result:** HTTP 402:
```
API key budget too low. This request costs ~33.3433 pollen, but this key has 5.0000.
```

The actual cost should be **0.01 pollen** — the 33.34 estimate is ~3334× too high.

### `getEstimatedPrice()` — `gen.pollinations.ai/src/utils/model-stats.ts`

```ts
export function getEstimatedPrice(
    stats: TinybirdModelStats,
    model: string | undefined,
): number {
    if (!model) return 0;
    const row = stats.data?.find((r) => r.model === model);
    return row?.avg_cost_usd || 0;
}
```

This function **only** consults Tinybird's `avg_cost_usd` and never references:
- The model definition's `cost.completionImageTokens` (official models)
- The community endpoint's `completionImagePrice` (community models)
- The `flat_rate` flag

### `checkBalance()` — `gen.pollinations.ai/src/utils/generation-access.ts`

```ts
const estimatedCost = getEstimatedPrice(
    await getModelStats(env.KV, log),
    model.resolved,
);
```

The `isFreeCommunityModel` check only passes if **all** `COMMUNITY_ENDPOINT_PRICE_FIELDS` are `0`. A flat-rate model priced at `0.01` is not "free", so the balance check proceeds with the Tinybird estimate — which is incorrect.

### Why the estimate is wrong

The Tinybird `avg_cost_usd` for `vendouple/zimage` appears to be contaminated or misattributed — possibly conflated with the official `zimage` model (per-token pricing at `completionImageTokens: 0.004/image`). The 33.34 figure does not correspond to any defined pricing in the registry or community endpoint definition.

## Impact

- **Paying users blocked from using community models** they have sufficient balance for
- Affects all flat-rate community models where Tinybird `avg_cost_usd` diverges from the defined price
- Revenue-blocking: users who topped up cannot spend their pollen

## Suggested Fix

`getEstimatedPrice()` (or `checkBalance()`) should prefer the **model definition's pricing** when available:

1. For **community models**: use `communityEndpoint.completionImagePrice` (or the relevant price field) as the estimate, not Tinybird averages
2. For **official models**: use `model.definition.cost.completionImageTokens` (or the relevant cost field)
3. Fall back to Tinybird `avg_cost_usd` **only** when no defined pricing exists

This ensures the pre-flight check matches the actual charge that will be applied post-generation.

## Evidence

- User report: `kanga0744-art` in Discord thread (Jul 25, 2026)
- Confirmed by `the_yerminator`: model `vendouple/zimage` is 0.01 pollen/image
- Code traced: `getEstimatedPrice` → `model-stats.ts:3`, `checkBalance` → `generation-access.ts:31`
- Related: [#12289](https://github.com/pollinations/pollinations/issues/12289) (model loading issue, separate but same user affected)

---
**Author:** `the_yerminator` (UID: `1123534156626939945`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1529418685901701211/1530509108510195793)

davealan74:
I hit this trying to call a community image model that costs 0.01 pollen a shot and getting a 402 saying the request costs ~33.34 pollen. Same thing reported in #12747 for `vendouple/zimage`. The pre-flight check in `checkBalance()` runs everything through `getEstimatedPrice()`, which only ever looks at Tinybird's `avg_cost_usd` for the model — so whenever that historical average has drifted (or picked up cost from a similarly named model), users with plenty of balance get turned away.

For flat-rate models there's nothing to estimate: the fee is fixed per generation and the registry already knows it exactly. So `getEstimatedPrice()` now takes the resolved `ModelDefinition` and, when `flatRate` is set, sums `getPriceDefinitionForModel()` instead of consulting Tinybird. Everything else is unchanged — token-priced models still fall back to the historical average, which is the case it was actually written for, since you can't know the charge before generating.

I keyed it off `flatRate` rather than special-casing community endpoints because `communityModelDefinition()` already sets that flag for request-mode image endpoints, and the flat-fee audio/3D models in the registry have the same exposure. One guard in the shared function covers all of them.

Added `gen.pollinations.ai/test/model-stats.test.ts` covering a flat-rate community model, a flat-rate registry model, and a token-priced community model, all against stats rows with deliberately inflated averages. Full gen suite passes (77 files / 721 tests) and `npm run typecheck -w gen.pollinations.ai` is clean.

Closes #12747


davealan74:
Rebased onto latest main, so this is no longer behind. @voodoohop you're the one who's been in `model-stats.ts`/`generation-access.ts` most recently — could you take a look when you get a chance?

YoannDev90:
My apologies — I mistakenly closed these PRs. This was an error on my part. Please disregard the recent closure, they should remain open.

voodoohop:
hey @davealan74. thanks so much for the PR. just reviewing. i'm checking first why the estimate logic is so off. it looks like there may be a weird bug that should be fixed. i dont know why the estimates would show such wrong values for flatRate models

davealan74:
No bug in `public_model_stats.pipe` — it's a plain `GROUP BY resolved_model_requested` on `generation_event_v2`, no join, and community model ids are namespaced `owner/name` (unique per live row via `idx_community_endpoint_owner_name`), so one model's rows can't merge into another's. Live data confirms the average is accurate at volume: `zimage` avg_cost_usd is 0.004004 vs its defined 0.004, `flux` is 0.002 vs defined 0.002. The wild swings hit low-volume community models sitting near the pipe's own `priced_success_count >= 3` floor (many `vendouple/*` models have single-digit priced counts), where a couple of atypical requests or a mid-window price change on the same endpoint name swings a tiny-sample mean by orders of magnitude — that's inherent noise in averaging, not an aggregation bug.

PR #12808, opened 2026-07-27
fix: enforce requested dimensions for self-hosted pool image results

### LINKED ISSUE #12225: Pollinations API ignores width/height parameters — returns incorrect resolution for ESP32-S3
I'm developing an application for ESP32-S3 using the Pollinations API for image generation. I've encountered a critical issue where the API completely ignores the width and height parameters passed in the URL.

What I'm sending:

https://gen.pollinations.ai/image/Elk%27s%20horns%20glow%20with%20a%20soft%20blue%20light%20at%20dusk%2C%20urban%20art%2C%20city%20life%2C%20street%20view%2C%20architectural%2C%20metropolitan%2C%20dynamic%20composition%2C%20interesting%20angle%2C%20low%20angle%20shot%2C%20dynamic%20pose?seed=34499318&negative_prompt=blurry%2C%20out%20of%20focus%2C%20ugly%2C%20disfigured%2C%20artifacts%2C%20noise%2C%20text%2C%20signature%2C%20watermark%2C%20poor%20quality%2C%20low%20resolution%2C%20missing%20limbs%2C%20extra%20limbs%2C%20mutated%2C%20distorted%2C%20bad%20anatomy%2C%20poorly%20drawn%20hands%2C%20worst%20quality&model=flux&enhance=true&nologo=true&quality=high&width=960&height=640&safe=false

Image

What I'm getting back
{
  "width": 1024,
  "height": 576,
  "model": "flux",
  "quality": "high"
}

Expected behavior:
Image generated at exactly 960×640 pixels (3:2 aspect ratio)
JSON response should reflect the requested dimensions

Actual behavior:
Server returns 1024×576 (16:9 aspect ratio)
Aspect ratio is distorted from 3:2 to 16:9
The response confirms the server is ignoring my parameters

What I've tried:
Different width/height combinations
Different models (flux, z-image, etc.)
Different quality levels (low, medium, high)
The dimensionsExplicit flag (set to true)
Nothing works — the server always returns its own chosen resolution.

This is a blocker for my project because:
My display has a fixed 3:2 aspect ratio (480×320)
Images need to fit exactly without cropping or distortion
I can't predict what resolution the API will return

Please fix the API to:
Respect the requested width and height parameters
Generate images at exactly the requested dimensions
Return those exact dimensions in the JSON response

If certain dimensions aren't supported, return a clear error

Mode | For my display | I send (requested)
-- | -- | --
LOW | 480×320 | width=512&height=384
MEDIUM | 640×480 | width=768&height=576
HIGH | 960×640 | width=1024&height=768

</div>
</body>
</html>
This is happening on gen.pollinations.ai with model flux. The issue is 100% reproducible. I've tested with multiple prompts and seeds.

Thank you for looking into this!

davealan74:
Explicit `width`/`height` only survived when a generation went through the Replicate flux fallback — that handler (`callReplicateFluxSchnellAPI`) crops its output to the requested size when `dimensionsExplicit` is set. Requests served by the self-hosted GPU pool (`callSelfHostedServer`, used by flux/zimage/sana) returned the worker's own resolution untouched, which is why the reporter's 960x640 request came back as 1024x576 with the aspect ratio distorted.

Rather than copy the crop into the pool path, I moved it up to `processImageBuffer` in `createAndReturnImageCached`, where results from every provider converge before EXIF and caching, and dropped the now-duplicate block from `replicateFluxModel.ts`. Same `transformImage` call and same `fit: "cover"` as before, just in one place, so Azure/OpenRouter handlers that snap to preset sizes get the same treatment. SVG results still skip it, and when no dimensions were requested nothing changes — the existing `convertToJpeg` path runs as before.

Extended `test/image/fluxFallback.test.ts` with two cases through `createAndReturnImageCached`: a pool response at 1024x576 with a 960x640 explicit request now comes back at 960x640, and the same response with no explicit dimensions is left at 1024x576. The Images binding is stubbed so the final buffer's dimensions can be read back. I checked the first test fails on main before the change. Full gen suite passes (720 passed, 6 skipped) along with typecheck.

Closes #12225

### DIFF
--- a/gen.pollinations.ai/src/image/createAndReturnImages.ts
+++ b/gen.pollinations.ai/src/image/createAndReturnImages.ts
@@ -50,6 +50,7 @@ import {
 } from "./utils/imageDownload.ts";
 import {
     resizeForGptImage,
+    transformImage,
     convertToJpeg as transformToJpeg,
 } from "./utils/imageTransform.ts";
 import type { TrackingData } from "./utils/trackingHeaders.ts";
@@ -902,14 +903,29 @@ const prepareMetadata = (
  * @param {Buffer} buffer - The raw image buffer
  * @param {Object} metadataObj - Metadata to embed in the image
  * @param {Object} maturity - Additional maturity information
+ * @param {Object} safeParams - Parameters the image was requested with
  * @returns {Promise<Buffer>} - The processed image buffer
  */
 const processImageBuffer = async (
     buffer: Buffer,
     metadataObj: object,
     maturity: object,
+    safeParams: ImageParams,
 ): Promise<Buffer> => {
-    const processedBuffer = await convertToJpeg(buffer);
+    // Backends only approximate the requested size: the GPU pool returns
+    // whatever resolution the worker produced, Replicate snaps to a preset
+    // aspect ratio, Azure to a fixed size. When the caller asked for exact
+    // dimensions, crop to them here — once, for every backend — instead of
+    // per model handler (issue #12225).
+    const processedBuffer = safeParams.dimensionsExplicit
+        ? await transformImage(buffer, {
+              width: safeParams.width,
+              height: safeParams.height,
+              fit: "cover",
+              format: "image/jpeg",
+              quality: 90,
+          })
+        : await convertToJpeg(buffer);
     return await writeExifMetadata(processedBuffer, metadataObj, maturity);
 };
 
@@ -956,6 +972,7 @@ export async function createAndReturnImageCached(
                       result.buffer,
                       metadataObj,
                       maturity,
+                      safeParams,
                   );
 
         return {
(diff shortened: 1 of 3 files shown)


davealan74:
@ElliotEtag since you've been working in `createAndReturnImages.ts` and the flux model, could you take a look at this one when you have a chance? It now conflicts with the fireworks→replicate rename, happy to rebase it over that if the approach looks right to you.

voodoohop:
Thanks for this, and sorry it sat so long — the diagnosis was right and the patch lands in a sensible place.

We're going to leave resizing to the client though:

- Scaling to exact pixels is cheap client-side, and doing it here costs a Cloudflare Images transform on every sized request.
- `width`/`height` are unbounded at the gateway, so `8000x8000` would ask CF to upscale a ~1MP render to 64MP.

Closing for that reason rather than anything wrong with the code. Genuinely appreciate you picking it up.

PR #12809, opened 2026-07-27
fix: round a fractional key expiry to whole seconds

### LINKED ISSUE #10743: BYOP form: null/empty values cause JSON validation error instead of using defaults
## Bug Description

The BYOP authorization form at `enter.pollinations.ai` has strict validation that prevents using `null` or empty values for optional parameters like `budget`.

## Current Behavior

| Scenario | Result |
|----------|--------|
| Set `budget=null` | Not accepted, form forces default value |
| Delete number from field (empty) | **"JSON validation error"** on clicking "Authorize" |
| Leave default numbers | Works, but forces explicit values |

## Expected Behavior

Optional parameters should accept:
- Explicit `null` values
- Empty/blank fields (interpreted as "use default/unlimited")

## Steps to Reproduce

1. Go to BYOP authorization form
2. Remove the number from `budget` field (make it empty)
3. Click "Autorizar"
4. Observe JSON validation error

## Additional Context

This affects the UX when app creators want to let Pollinations use **default/unlimited** values rather than hardcoded limits. Currently forced to specify explicit numbers.

/cc @voodoohop 

---
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1500964478936809575/1501951001597575269)

davealan74:
The expiry field takes days and allows fractions (`step="any"`), but key creation takes whole seconds — `expiresIn: z.number().int().positive()`. A fraction landing mid-second failed the `.int()` check, so `0.123456` days (10,666.4s) 400'd with `"JSON body validation failed"` instead of creating a key.

**The fix is `Math.round`.** That's the whole thing.

- `expiryDaysToExpiresIn` moves to `shared/auth/authorize-config.ts` so it can be unit-tested — importing it from `frontend/src/lib` drags in `apiClient`, which needs `window`
- Rounds days to whole seconds
- An empty field stays the only thing meaning "no expiry"; `0`, negatives and over-cap values stay invalid and the server rejects them, which the dialogs already surface via `setError`

Verified: every fractional value the field accepts (`0.5`, `0.25`, `0.1`, `1.5`, `30.5`) converts identically to before — only mid-second fractions change, from a 400 to a working key.

Addresses #10743


davealan74:
Flagging for a reviewer with dashboard access: I couldn't verify this in the live enter.pollinations.ai UI (no account there), so the fix is only confirmed at the unit and API level. Worth a sanity-check that a 0 or blank expiry now mints a never-expiring key from both the consent screen and the dashboard's create-key dialog.

voodoohop:
Pushed a follow-up (0e00083). Your read on the edit dialog is right and the 400 is real — but `0` also reaches this helper from the consent screen's `?expiry=` query param, which a third-party app controls. `expiry ?? DEFAULT_CONSENT_EXPIRY_DAYS` let `0` through, so the screen would display "0 days" while the key it minted had no expiry at all.

The follow-up adds `sanitizeConsentExpiryDays`: a caller-supplied expiry is honoured only when it's positive, finite and within the server's 365-day ceiling, otherwise it falls back to the 7-day default rather than to "never". `expiryDaysToExpiresIn` now also clamps a positive value to at least one second, so a fractional day can't round down into "no expiry" either. All of your existing assertions still pass.

Also switching `Closes #10743` to `Addresses` — your own investigation showed the budget half was already fixed, so this shouldn't auto-close it.

davealan74:
Thanks for catching the consent screen path — I'd only tested the create-dialog case and missed that `?expiry=0` from a third-party app could slip through the `??` fallback and mint a permanent key while showing "0 days". `sanitizeConsentExpiryDays` is the right fix: falls back to the 7-day default rather than to "never", so a hostile caller can't abuse it. The `Math.max(1, …)` tightening in `expiryDaysToExpiresIn` is a good clean-up too.

Good call on `Addresses` instead of `Closes` — the budget half was already working and shouldn't be auto-closed by this.

voodoohop:
@davealan74 do you happen to be in our discord too?

voodoohop:
Thanks for your contributions — we've sent a bonus your way:

- **25 Paid Pollen** and **25 Quest Pollen**, as two separate rewards
- Claim them under *Bonus rewards* at https://enter.pollinations.ai/quests
- Paid Pollen is the balance that paid-only models draw from

voodoohop:
Simplified this and brought it up to date with main — it's green now.

**The rounding was the whole bug.** A fractional expiry landing mid-second failed the server's `.int()` check. `Math.round` fixes it. Everything else came out:

- **`0` and negatives no longer become "no expiry."** They stay invalid, the server rejects them as it always did, and the dialog shows that error. Turning an unusable entry into a key that never expires was the wrong direction to fail in — an empty field is the only way to ask for no expiry.
- **`sanitizeConsentExpiryDays` is gone.** It only existed to compensate for the above. Now that the helper just rounds, a third-party `?expiry=0` can't mint a permanent key — it errors, and `createKeyWithPermissions` rethrows that message into the consent screen's `setError`, so it's visible rather than silent.

Net: **+54/−9** instead of +111/−10, and every fractional value the field accepts (`0.5`, `0.25`, `1.5`, `30.5`) converts exactly as before.

Two other things — are you in our Discord? And thanks for the contributions, we've sent a bonus your way:

- **25 Paid Pollen** and **25 Quest Pollen**, as two separate rewards
- Claim them under *Bonus rewards* at https://enter.pollinations.ai/quests
- Paid Pollen is the balance that paid-only models draw from

Issue #12832, opened 2026-07-28
feat: short-lived agent run tokens for user-pays hosted agents


voodoohop:
## Problem

Hosted agents receive the caller's raw `sk_` and reuse it for every internal call. [#12702](https://github.com/pollinations/pollinations/pull/12702) (polli-agent) parks the key in a context variable and re-attaches it to brain/image/video/audio/search calls for up to 100 iterations. Billing lands on the right user, but the agent process holds a reusable, non-expiring, full-scope credential — exposed to any logging mistake, malicious tool, or dependency compromise.

## Approach

A **run token**: short-lived signed credential, minted per agent invocation, resolving to the same identity the parent key produces.

- Format: `ag_<base64url(payload)>.<hmac-sha256>`, claims `{sub, pk, run, agent, aud, models?, exp}`
- **No row anywhere.** Not in `apikey`, not a new table — so the user's key list stays clean
- Billing unchanged: `apiKey.id` resolves to the parent key, so `generation-access.ts:42` (preflight 402) and `track-helpers.ts:181` (per-key deduction) act on the parent's `pollen_balance`
- Cap = the parent key's existing pollen budget. Users who want agents bounded put a budget on the key they hand out — already shipped, no new metering
- Revocation free: the parent key row is read on every request, so deleting/expiring it kills all derived tokens
- `aud: "gen"` — run tokens are never accepted by enter

Industry convention (AWS STS, GitHub installation tokens, GCP short-lived SA creds) is exactly this: stateless signed credential, never listable. No production system encodes a running spend cap in the credential itself — that always lives in a separate meter. Here the meter already exists.

## Follow-ups (independent, not blockers)

- [#11637](https://github.com/pollinations/pollinations/pull/11637) — bound child keys by parent
- `atomicDeductApiKeyBalance` (`shared/billing/deduction.ts:85-93`) lacks `AND pollen_balance >= amount`, so budgets are a soft ceiling under concurrency
- Emit `runId`/`agentId` into the generation event for per-run cost in Tinybird
- Decide pass-through vs bundled economics vs [#11373](https://github.com/pollinations/pollinations/issues/11373) so agent runs aren't billed twice

voodoohop:
Shipped in #12844 (`8f9ca94b1`), agent side in #12842 (`5e3dc7e81`). Closing.

**As designed:** no row anywhere, billing resolves to the parent key so preflight/deduction/attribution are untouched, cap is the parent key's existing pollen budget, revocation is free.

[...]

PR #12833, opened 2026-07-28
feat: add agent run tokens

voodoohop:
Addresses #12832. End-to-end: a hosted agent can spend the caller's pollen without ever holding their key.

- Adds `shared/auth/run-token.ts` — HMAC-signed `ag_<payload>.<sig>`, claims `{sub, pk, run, agent, aud, models?, exp}`, fixed 30 min TTL bounded again at verify time
- Adds a branch in `authenticateApiKeyRequest` resolving a run token to the parent key's identity, so the preflight 402 and per-key deduction act on the parent's `pollen_balance` — no billing changes
- `gen` mints a token per community-endpoint call and forwards it as `X-Pollinations-Key`; `Authorization` keeps carrying the endpoint's own access credential
- A request already running under a run token forwards it unchanged, so chained agents cannot push the deadline out
- Persists nothing: no `apikey` row, no new table, key list stays clean
- Attenuation only — models intersected with the parent's allowlist (omitted by default, so the token inherits caller scope), account permissions dropped
- Parent row read per request, so deleting or expiring it invalidates every derived token
- 17 tests, incl. one proving a run token's spend debits both the parent key's budget and the parent user's wallet; `byop-markup`, `community-endpoints`, `portkey-fallback-headers` and `model-permissions` re-run as regressions

Requires `RUN_TOKEN_SECRET` on gen. Until it is bound, no run token is minted and none is accepted — fails closed. Enter needs it too for the `media → /account/key` upload path.

Agent side is #12842. Per-run cost attribution and creator earnings are deliberately out of scope.


voodoohop:
Superseded by #12844.

Nothing from this branch shipped. The hand-rolled HMAC module lost to jose, and the two-header transport was dropped once we settled on delivering the run token as the endpoint's bearer — a delegating community model stays a plain OpenAI-compatible server that way, with no Pollinations-specific header to implement.

The one thing worth keeping is ported: the test proving a run token moves both the parent key's budget and the parent user's wallet through `handleBalanceDeduction`. #12844's suite asserted identity resolution but never ran a deduction, so the central claim — billing never learns run tokens exist — was untested.

PR #12835, opened 2026-07-28
feat: add scoped agent run tokens

voodoohop:
Addresses #12832.

- Adds `POST /v1/agent/run-token` for explicit, user-authorized 30-minute `ag_` bearer credentials.
- Resolves tokens through shared Gen/Enter auth to the live parent key, preserving billing, BYOP attribution, rate limiting, revocation, and budgets.
- Lets callers inherit or narrow parent model access; run tokens cannot mint nested tokens or call community models.
- Uses the existing auth secret with domain separation; adds no table, migration, or secret.
- Covers scope enforcement, expiry bounds, tampering, revocation, model visibility, and Enter `/account/key` media verification.

This PR deliberately does not auto-delegate credentials to community endpoints. Users mint a token and hand it directly to an agent they choose.


voodoohop:
Superseded by #12844, which keeps this PR's core intact: the JOSE token module, the shared gen/enter verification, the mint endpoint, and the no-community-models rule. #12844 adds registering an agent as a community model — a `delegates_generation` flag on the endpoint row, off by default, with the run token delivered as the endpoint's bearer.

Later revisions from here are carried over: `isAgentRunToken` unexported, and `agent_id` documented as a caller-provided audit label rather than a verified identity.

PR #12843, opened 2026-07-29
test: verify delegated Polli run tokens

voodoohop:
Stacked on #12702; pairs with #12835.

- Confirms an opaque `ag_` bearer reaches Polli's per-request credential context.
- Covers the shared credential path used by the brain, registry, generation tools, and media uploads.
- Documents that the user's original API key is never required by Polli.
- Makes no runtime changes because #12702 already forwards arbitrary bearer credentials correctly.

### DIFF
--- a/apps/polli-agent/README.md
+++ b/apps/polli-agent/README.md
@@ -27,7 +27,7 @@ docker build -t polli-agent .
 docker run -p 8000:8000 --env-file .env polli-agent
 ```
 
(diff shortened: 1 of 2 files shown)


voodoohop:
Closing — both changes now live on #12842's branch (`cdc8247fc`, `6991f74aa`): the propagation test asserting the run token reaches the brain and tools, and the tightening to `ag_`-only so a long-lived `sk_`/`pk_` can never be spent by the agent.

Issue #11127, opened 2026-05-18
Tinybird staging workspace — deferred follow-ups


ElliotEtag:
## Overview
Follow-ups from the session that split off a separate `pollinations_enter_staging` Tinybird workspace. Split is live (staging worker writes there, prod workspace is clean of staging rows, pipe filters removed). These are the items we explicitly deferred.

## Tasks
- [ ] PR-time CI check: `tb deploy --check` against the staging workspace
- [ ] Auto-deploy datafiles to prod workspace on merge to `production` branch
- [ ] Workspace drift check: flag when staging and prod schemas / pipe SQL diverge (read-only prod token, non-blocking warning)
- [ ] Create a `tinybird_sync` token in staging workspace, swap `TINYBIRD_SYNC_TOKEN` in `secrets/staging.vars.json` (today still points at prod)
- [ ] Adopt 3-tier dev workflow: `tb local start` for local dev + `tb test run` fixtures for the busiest pipes
- [ ] Review the staging tier-refill cron (`*/15 * * * *`) — is it useful, or drop it from `[env.staging.triggers]`?
- [ ] Update `.claude/skills/tinybird-deploy/SKILL.md` to document the dual-workspace setup and the "always deploy to both" rule
- [ ] Optional: strip the now-redundant `environment:` field from worker event payloads (column kept in schema, just stops being set)

## Related
- PR with the staging-workspace split: TBD (from `feat/observability-staging-workspace`)

ElliotEtag:
Follow-up findings from staging end-to-end test:

- D1 sync is broken on staging because `TINYBIRD_SYNC_TOKEN` in `secrets/staging.vars.json` was not swapped (still points at prod workspace). Staging worker's `/api/admin/trigger-d1-sync` writes to prod.
- `app_directory` is also empty on staging — the `tinybird-sync-apps.yml` GitHub Action uses a single repo-level `secrets.TINYBIRD_SYNC_TOKEN` pointing at prod. Needs a staging-side variant (new repo secret + parallel workflow, or matrix).
- Net effect on staging: `d1_user`, `d1_apikey`, `d1_session`, `d1_account`, `app_directory` all empty. Pipes that join against them return empty rows. Worker isolation is intact (no pollution back to prod), just no reference data on staging.

Verified working into staging: `stripe_event` (12 rows, live), `tier_event` (9 rows, cron-triggered).

ElliotEtag:
Closing — staging follow-ups no longer tracked as a standalone epic.

PR #12888, opened 2026-07-30
fix: stop counting fallback attempt rows as requests

voodoohop:
Prepares every consumer of `generation_event_v2` for the per-attempt rows that #12819 emits. **Deploy this before #12819 reaches production.**

- Adds `is_attempt_row Bool DEFAULT false` so each consumer picks its own reading: exclude attempts to count requests, or read exactly them to judge whether a model's own upstream works
- Filters the 11 pipes/materializations that count rows; four consumers act on that number unattended — the community monitor deactivates endpoints, Grafana pages, abuse-detection drives ban recommendations, the leaderboard has a request floor
- Fixes `activity_earnings_model_rewards_mv` labelling a rescuer's payout with the model the *caller* asked for, i.e. another owner's listing

Provably a no-op until the worker ships: nothing writes the column, and every existing row is a settlement row, so `DEFAULT false` keeps all history counting exactly as it does today.

Not deployed — Tinybird has no CI auto-deploy (#11127), so this needs `deployment create --check` on staging first, then both workspaces by hand. Grafana rules are a separate deploy path.

Addresses the telemetry defect found reviewing #12819.

### DIFF
--- a/.claude/skills/abuse-detection/SKILL.md
+++ b/.claude/skills/abuse-detection/SKILL.md
@@ -76,6 +76,7 @@ FROM (
         countIf(g.response_status >= 400) * 100.0 / count() as err_pct
     FROM generation_event_v2 g
     WHERE g.start_time >= now() - INTERVAL 7 DAY
+        AND g.is_attempt_row = 0
         AND g.user_id NOT IN ('undefined', '')
     GROUP BY g.user_id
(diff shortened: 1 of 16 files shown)


voodoohop:
Folded into #12819 — the consumer filters and the worker change belong in one PR.

PR #12906, opened 2026-07-30
docs: clarify static browser PKCE clients

q8paradise:
## Summary

- Clarifies that PKCE-enabled static/public browser clients can exchange the authorization code directly, instead of implying every integration needs a server.
- Adds a browser `fetch` example that validates `state`, requires the saved verifier, removes callback parameters, discovers the token endpoint, and reuses the exact registered redirect URI.
- Documents memory-first/session-only token handling and warns against `localStorage`, URLs, analytics, and logs.
- Links both server-backed and browser-only in-repo references.

## Verification

- Confirmed discovery advertises S256 and token authentication method `none`; the live token endpoint's CORS preflight allows browser POSTs.
- Checked `enter.pollinations.ai/src/routes/oauth.ts`, `src/routes/well-known.ts`, `src/index.ts`, `apps/oauth-client-demo`, and `apps/oauth-test`.
- Checked all open PR file overlaps; none covers this static-client guidance.
- Ran `git diff --check`, Node syntax validation for the new snippet, Markdown fence/link checks, directory checks, and an added-line secret scan. No packages, builds, or test suites were run.

## Security

The example fails closed on missing/mismatched state or verifier, removes transient OAuth data, and avoids persistent/callback-URL token exposure.


voodoohop:
Thanks for this — the point that static browser clients can exchange the code themselves was missing from the docs, and the `apps/oauth-test/` link is a good catch.

I pushed a follow-up that trims the example. The curl right above already shows every parameter of the same request, so the browser version only needed to show that it's the same call from `fetch`. I also moved it after the expiry/scopes prose rather than between them, and dropped the discovery fetch since the curl hardcodes that URL three lines up.

Net effect is the same guidance in about a third of the lines. Merging now.

PR #12912, opened 2026-07-31
refactor: share community endpoint error parsing

voodoohop:
- Moves shared OpenAI-style error envelope parsing into the community endpoint utilities.
- Keeps Gen and Enter status prefixes and remapping behavior unchanged.
- Removes duplicate parser implementations so accepted upstream shapes stay in sync.
- Tests: 35 Gen community endpoint tests, 8 Enter endpoint tests, both typechecks, and Biome.


voodoohop:
Superseded by #12946, which keeps the current shared error parser and combines it with the related community endpoint test cleanup.

PR #12914, opened 2026-07-31
refactor: remove dead Portkey fallback config

voodoohop:
- Removes the unused Portkey `strategy/targets` header builder left after Airforce routing was reverted in #11870.
- Keeps `fallback_used` observability and the new in-worker per-model fallback path unchanged.
- Drops the obsolete Airforce-specific test while retaining coverage for active single-provider Portkey headers.
- Tests: 47 focused tests, Gen typecheck, and Biome.


voodoohop:
Superseded by #12921

PR #12915, opened 2026-07-31
refactor: share credential query redaction

voodoohop:
- Moves credential query-string redaction into the existing shared observability utility.
- Replaces duplicate logger and realtime implementations with the same canonical credential-name set.
- Expands both paths to cover aliases such as `apikey` and `bearerToken` without changing non-credential query parameters.
- Tests: 32 realtime/error-observability tests, Gen typecheck, and Biome.


voodoohop:
Superseded by #12955, which now contains this branch's full history and ports its credential-alias assertions onto the shared realtime session harness. The combined draft is updated from current main and revalidated.

PR #12919, opened 2026-07-31
refactor: remove unused KPI GitHub App auth

voodoohop:
- Removes dormant GitHub App JWT signing and installation-token exchange code from the KPI Worker.
- Keeps the deployed `GITHUB_TOKEN` path and unauthenticated local behavior, with one shared header builder for both GitHub endpoints.
- Verified read-only against production binding names: none of the three GitHub App credentials are configured; no secret state was changed.
- Checks: production frontend build, Wrangler Worker dry-run, and Biome.


voodoohop:
Superseded by #12936

PR #12925, opened 2026-07-31
refactor: trust normalized image parameters

voodoohop:
- Removes stale `string | string[] | undefined` handling after `ImageParamsSchema` has normalized every image input to `string[]`.
- Deletes impossible zero-image guards inside edit-only branches across Azure, Vertex, Bedrock, and self-hosted adapters.
- Keeps generation/edit routing and provider request shapes unchanged.
- Corrects GPT Image error telemetry so an empty normalized image list is not reported as an image input.

Tests:
- `npx vitest run test/image/params.test.ts test/image/gptImageRouting.test.ts test/image/novaCanvasModel.test.ts test/image/veoVideoModel.test.ts`
- `npm run typecheck`
- `npx biome check --write` on all changed files

### DIFF
--- a/gen.pollinations.ai/src/image/createAndReturnImages.ts
+++ b/gen.pollinations.ai/src/image/createAndReturnImages.ts
@@ -420,7 +420,7 @@ const callGPTImageWithEndpoint = async (
         );
     }
 
-    const isEditMode = safeParams.image && safeParams.image.length > 0;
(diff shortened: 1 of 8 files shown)


voodoohop:
Superseded by #12967, which re-audits the same normalized-image assumptions, keeps the safe adapter subset, and preserves the tested direct-call tolerance.

PR #12926, opened 2026-07-31
test: share community fallback fixtures

voodoohop:
- Replaces five copies of the same two-owner, two-endpoint fallback setup with one focused fixture.
- Keeps each scenario's routing, billing, failure, moderation, and permission assertions unchanged.
- Centralizes the shared encrypted credentials, price hierarchy, model IDs, and endpoint URLs.
- Removes 125 lines from the recent community fallback integration coverage.

Tests:
- `npx vitest run src/community-endpoints.test.ts`
- `npm run typecheck`
- `npx biome check --write src/community-endpoints.test.ts`


voodoohop:
Superseded by #12946, which includes this fixture deduplication with the related community endpoint runtime cleanup.

PR #12929, opened 2026-07-31
test: share realtime session setup

voodoohop:
- Shares paid realtime WebSocket setup, shutdown, and telemetry parsing across billing tests.
- Reuses the identical cached-modality usage event for full and mini model assertions.
- Keeps model-specific pricing, forwarding, fallback, referrer, and telemetry expectations explicit.
- Removes 198 lines without changing production behavior.

Tests:
- `npx vitest run test/realtime.test.ts`
- `npm run typecheck`
- `npx biome check --write gen.pollinations.ai/test/realtime.test.ts`

### DIFF
--- a/gen.pollinations.ai/test/realtime.test.ts
+++ b/gen.pollinations.ai/test/realtime.test.ts
@@ -137,6 +137,82 @@ async function waitForTinybirdRequests(
     }
 }
 
+async function openPaidRealtimeSession({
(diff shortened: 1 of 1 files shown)


voodoohop:
Superseded by #12955, which combines this realtime session-fixture cleanup with the matching registry deduplication.

PR #12931, opened 2026-07-31
refactor: share project classification

voodoohop:
- Reuses the live project-manager classifier for project backfills instead of maintaining a second API client and response parser.
- Renames the module to an importable Python name and removes the dynamic-import scaffold.
- Preserves backfill failure semantics: an unclassified item remains empty and migrations skip it.
- Updates the workflow and existing filename references; removes 32 lines overall.

Tests:
- Python AST parse for both scripts
- Mocked classifier request, response normalization, explicit backfill context, import, and failure-contract checks
- `git diff --check`


voodoohop:
Superseded by #12968, which now includes this shared project-classification cleanup together with the related repository automation simplifications. The consolidated draft was updated from current main and revalidated.

PR #12932, opened 2026-07-31
refactor: share API key metadata parsing

voodoohop:
- Replaces three JSON metadata parsers and two redirect-list readers with one shared auth utility.
- Updates every auth, billing, and Enter consumer and removes the route-local compatibility module.
- Preserves invalid/null/non-object rejection and non-empty string redirect filtering.
- Removes 27 lines and adds direct coverage for both shared behaviors.

Tests:
- `npx vitest run test/api-key-metadata.test.ts test/integration/account-keys.test.ts`
- `npx vitest run src/byop-markup.test.ts`
- Enter and Gen `npm run typecheck`
- Biome on all changed TypeScript files


voodoohop:
Superseded by #12924, which combines this metadata normalization with the overlapping API-key auth loading cleanup.

PR #12934, opened 2026-07-31
refactor: use Buffer for image base64 decoding

voodoohop:
- Replaces the handwritten bit-shifting decoder used across image/model paths with Node's existing `Buffer` decoder.
- Preserves data URI stripping, whitespace handling, unpadded and URL-safe input, plus the strict invalid-character 502 error.
- Removes 25 lines and retains the same helper interface for all callers.
- Checks: Biome, Gen typecheck, focused image tests (17), direct and differential decoder checks.


voodoohop:
Superseded by #12947, which now combines this Buffer-based decoder simplification with the related obsolete Vertex image integration removal. The consolidated branch was updated from current main and revalidated.

PR #12938, opened 2026-07-31
refactor: simplify model info helpers

voodoohop:
- Collapses an unreachable model-description fallback chain to its actual source field.
- Removes unused modality helpers, output handling, exports, and the empty per-model logo override layer.
- Reuses canonical input-modality extraction for labels instead of duplicating the same checks.
- Reduces the helper surface by 60 net lines; Biome and the full Enter frontend production build pass.


voodoohop:
Superseded by #12930.

PR #12942, opened 2026-07-31
refactor: remove obsolete registry billing API

voodoohop:
- Removes the unused explicit cost- and price-definition calculation exports.
- Keeps the live model-definition path backed by `calculateUsageBilling` as the single billing implementation.
- Eliminates 29 lines of parallel billing surface with zero callers.
- Biome passes and the focused shared pricing suite passes all 28 tests.

### DIFF
--- a/shared/registry/registry.ts
+++ b/shared/registry/registry.ts
@@ -476,17 +476,6 @@ export function calculateCostForModelDefinition(
 }
 
 /**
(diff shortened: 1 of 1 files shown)


voodoohop:
Superseded by #12952, which now includes this registry API removal together with the related unused Gen API cleanup. The combined draft has been rebased through a normal merge from current main and revalidated.

PR #12941, opened 2026-07-31
refactor: simplify Chat UI and storage

voodoohop:
- Removes eight unreferenced Chat components and their dedicated stylesheets; none are imported by the active app.
- Removes the unused React starter asset.
- Simplifies local storage to direct JSON serialization while preserving the same stored values and fallback behavior.
- Reduces the Chat app by 2,059 net lines without removing a reachable UI path.

Validation:
- npm run build --prefix apps/chat
- git diff --check origin/main...HEAD
- Exact-reference search confirms the deleted components have no remaining consumers.
- npm run lint --prefix apps/chat still reports 4 pre-existing errors and 9 hook warnings outside the changed code.

PR #12943, opened 2026-07-31
refactor: simplify Chat storage serialization

voodoohop:
- Replaces a nested chat/message mapping whose every branch returned the original value with direct JSON serialization.
- Removes the unused accent-color storage key and accessors.
- Preserves stored chat JSON and quota-error behavior while reducing the module by 32 net lines.
- Biome, a differential serialization check, and the Vite production build pass.


voodoohop:
Superseded by #12941.

PR #12947, opened 2026-07-31
refactor: simplify legacy image utilities

voodoohop:
- Removes the unreachable direct Vertex AI image client, generator, and tests left after NanoBanana moved to OpenRouter.
- Replaces the handwritten image base64 decoder with Buffer while preserving data URI, whitespace, unpadded, URL-safe, and strict invalid-character behavior.
- Removes 993 net lines while retaining active OpenRouter routing, usage validation, billing, and content handling.
- Combined validation: Biome, no-reference search, and git diff checks. The component branches previously passed Gen typecheck, 47 focused image tests, and direct/differential decoder checks.


voodoohop:
Superseded by #12908

PR #12951, opened 2026-07-31
refactor: remove unused Enter type facade

voodoohop:
- Removes the orphan activity barrel; active routes already import its implementation modules directly.
- Removes four internal API/auth route type aliases with zero source or test references.
- Leaves live client, user, and backend route types unchanged.
- Verified with Biome, Enter's full typecheck, and the production frontend build.


voodoohop:
Superseded by #12930.

PR #12954, opened 2026-07-31
refactor: simplify micro email facade

voodoohop:
- Replaces the object of one-to-one `MicroService` forwarding methods with a direct subclass.
- Keeps the documented wrapper API, including the `verifyConnection()` compatibility name, through one small adapter.
- Removes 57 net lines and the stale SMTP example embedded in the wrapper source.
- Verified with Biome, 10 Vitest tests, TypeScript, the production Next build, and a six-method API smoke check.


voodoohop:
Superseded by #12958

PR #12961, opened 2026-07-31
refactor: simplify media app state

voodoohop:
- Derives authentication from the API key in both media apps instead of synchronizing duplicate state.
- Makes the packaging app generate only from its selected `File`, removing the unreachable data-URI source fallback and duplicate size check.
- Uses the media service’s required canonical `url` response field instead of unsupported compatibility aliases.
- Shares loading cleanup through `finally` and fixes the packaging app’s invalid React type import.
- Removes 34 net lines. Biome, both ESLint checks, packaging typecheck, and both production builds pass.

### DIFF
--- a/apps/product-packaging-designer/src/App.tsx
+++ b/apps/product-packaging-designer/src/App.tsx
@@ -12,7 +12,7 @@ import {
     Sun,
     Upload,
 } from "lucide-react";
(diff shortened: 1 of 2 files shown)


voodoohop:
Superseded by #12950, which combines this consumer-app cleanup with the corresponding media upload plumbing simplification.

PR #12965, opened 2026-07-31
refactor: remove stale AssemblyAI fallback shape

voodoohop:
- Replaces singleton fallback arrays with one upstream model string per registry model.
- Creates AssemblyAI's required one-element `speech_models` array only in the request body.
- Removes reverse fallback billing logic; singleton requests always bill the requested registry model.

Validation:
- Verified current AssemblyAI model-selection semantics in the official docs.
- `npx biome check --write src/routes/assemblyai-transcription.ts`
- `npx vitest run test/assemblyai-transcription.test.ts` (5 tests passed)


voodoohop:
Consolidated into #12907, which now includes this branch's full commit history and combines the ElevenLabs and AssemblyAI audio-route simplifications. Closing this draft in favor of the consolidated survivor.

PR #12948, opened 2026-07-31
refactor: simplify Cohere stream sanitization

voodoohop:
- Replaces Cohere's hand-written SSE buffering and boundary parser with the existing `eventsource-parser` dependency.
- Consolidates duplicated leading/trailing control-token stripping and request rejection plumbing.
- Preserves split-token handling, choice metadata, malformed JSON pass-through, and pending-content flushes.
- Verified with Gen TypeScript and all 9 focused Cohere request/stream tests.

PR #12969, opened 2026-07-31
refactor: reuse SSE parser for Cohere streams

voodoohop:
- Reuses the existing `eventsource-parser` dependency for Cohere stream framing
- Removes the handwritten UTF-8 buffer, boundary scanner, and flush parser
- Keeps Cohere-specific content sanitization and buffered token handling unchanged
- Preserves standard SSE data, event, and id fields when re-serializing

Tests:
- `npm run typecheck`
- `npx vitest run test/text/transforms/sanitizeCohereResponse.test.ts` (9 passed)


voodoohop:
Closing as a duplicate of #12948, which already contains the Cohere SSE parser simplification and broader Cohere cleanup.

PR #12972, opened 2026-07-31
refactor: share model debugging queries

voodoohop:
- Adds one private helper for Tinybird token loading, validation, and SQL requests
- Removes the same transport setup from four model-debugging commands
- Keeps each command's arguments, SQL, and output unchanged
- Cuts 24 net lines across the debugging scripts

Check:
- `bash -n .claude/skills/model-debugging/scripts/*.sh`


voodoohop:
Superseded by #12968, which combines this model-debugging query deduplication with the related repository-maintenance tooling cleanup.

PR #12976, opened 2026-07-31
refactor: simplify model monitor polling

voodoohop:
- Combines the initial catalog/health refresh and recurring poll into one effect
- Lets the effect own its interval directly instead of coordinating through a mutable ref
- Preserves immediate refreshes, aggregation-window changes, polling cadence, and unmount cleanup
- Removes 20 net lines from the hook

Checks:
- `npx biome check apps/model-monitor/src/hooks/useModelMonitor.js`
- `npm run build --prefix apps/model-monitor`
- `git diff --check`


voodoohop:
Superseded by #12936, which combines this polling cleanup with the related operations-dashboard dead-code removal.

PR #12980, opened 2026-07-31
fix: preserve OpenRouter provider routing

ElliotEtag:
- Routes the five pinned Gemini models through Portkey's native OpenRouter adapter.
- Fixes the generic OpenAI custom-host adapter dropping OpenRouter's body-level `provider` constraint.
- Preserves every existing model ID, Google Vertex region pin, and `allow_fallbacks: false` setting.
- Updates the routing regression test to require the native adapter and reject the old custom host.
- Validated with the 33-test Gemini routing suite, TypeScript checking, and Biome.


voodoohop:
- Closing for now while we reassess the provider-routing behavior.
- Thanks for keeping the change focused.

PR #13006, opened 2026-08-01
fix: route Flux fallback through registry

ElliotEtag:
## What changed

- Moves the existing Vast → Replicate Flux fallback into the generic registry fallback system.
- Keeps the public `flux` contract unchanged: Vast primary, Replicate backup, free access, multiplier 1, and $0.002/image.
- Records failed Vast attempts separately and attributes successful fallback cost/provider data to Replicate ($0.003/image).
- Keeps the internal route out of catalogs, filenames, and EXIF while preserving scoped-key access to public `flux`.

## Why

The old catch-all adapter fallback hid which provider served a request and could mask non-operational errors. The generic path retries only eligible upstream failures and produces accurate attempt-level telemetry.

## Confirmed model contract

- Canonical name: `flux`
- Aliases: none
- Price multiplier: `1`
- Paid-only: no
- Pollinations-operated GPU: yes
- Registry provider: Vast
- Primary route: Pollinations Vast GPU pool → `black-forest-labs/FLUX.1-schnell`
- Fallback: Replicate → `black-forest-labs/flux-schnell`; no further fallback

## Validation

- Full gen suite: 805/805 tests passed before the latest main merge.
- Post-merge focused suite: 89/89 tests, typecheck, and Biome passed.
- Enter aliases: 286/286; effective-price snapshot passed.
- Live local E2E: three fresh 512×512 images passed in 1.84–2.31s; EXIF/catalog/cache retained public `flux` identity; billing logs showed $0.003 cost and $0.002 user price.
- Latest main refresh: focused fallback/tracking 56/56, Enter registry/pricing 328/328, broader image/billing 250/250, gen typecheck, and Biome passed.
- No secret changes.


ElliotEtag:
@voodoohop Could you confirm the intended scoped-key behavior before this merges?

- The core change follows the generic registry fallback design: public `flux` uses hidden `flux-replicate`, caller pricing stays on `flux`, and provider cost/attempt telemetry follows the route that ran.
- This branch currently lets a hidden static fallback inherit permission from the requested public model. A key scoped only to `flux` can therefore use `flux-replicate` internally.
- PR #12819 explicitly says scoped keys drop targets not present in their allowlist, while PR #12986 deliberately permits hidden registry targets to be linked. The interaction between those two rules is not explicit.
- My reading is that we should remove the inheritance exception, leaving restricted `flux` keys on Vast only unless `flux-replicate` is explicitly allowed. Please confirm whether that matches your intended policy, or whether hidden provider implementations should inherit the primary model permission.

PR #13014, opened 2026-08-02
Add confirmation dialog for activate/deactivate community models

MarcosFRG:
Currently, toggling a community model's active state fires immediately on click — no confirmation. This adds a confirmation dialog (matching the existing delete confirmation pattern) so users don't accidentally deactivate or reactivate a model.

**Changes:**
- New `community-endpoint-toggle-confirmation.tsx` — dialog with conditional title/intent based on `endpoint.disabled`
- `community-endpoints.tsx` — `onToggle` now opens the dialog instead of calling `handleToggle` directly

`handleToggle` and the card component are unchanged.


Circuit-Overtime:
- Reviewed against current `main`; the stale fork merges cleanly.
- Activate/deactivate now requires confirmation while preserving the existing mutation and loading behavior.
- No unresolved review threads.
- Validation: Biome and `git diff --check` pass; existing GitHub checks are green.

PR #13020, opened 2026-08-02
fix: surface catalog fetch failure instead of rendering empty models table

tomdacatto:
## Summary
- `fetchModelCatalog` now validates the parsed `/models` response and throws when it is empty, not an array, or every entry is missing a `name`/`id` — instead of silently resolving with unusable data.
- That throw routes through the existing `catalogError` handling in `models.tsx`, so the dashboard shows **"Could not load models."** instead of a blank table with no indication anything failed.
- Adds `parseModelCatalogResponse` unit tests covering valid, non-array, empty, and all-unidentifiable-entries responses.

## Why
Across #12289 and #12870, the recurring symptom was a **silently empty models table** — page skeleton loads, console is clean, no error banner — traced each time to a different edge/DNS failure (Cuba DNS, SIN edge timeout, Italy edge returning malformed 200s). The #12870 investigation flagged the actual gap: `getModelPricesFromCatalog`'s `.filter(Boolean)` swallows any unusable response with no signal to the user. This does not (and cannot) fix the underlying regional edge/DNS issue — that is infra-side — but it closes the frontend gap so any future recurrence is visible instead of silent.

Addresses #12289, #12870

## Test plan
- [x] `npx vitest run test/model-catalog.test.ts` — 4/4 passing
- [x] `npx tsc --noEmit` on the frontend project — clean
- [ ] Manual check on a deploy preview: confirm the models page still renders normally against the live catalog

### DIFF
--- a/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
+++ b/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
@@ -85,6 +85,21 @@ const formatEstimatedTtsPricePerSecond = (pricePerChar: number): string => {
         : pricePerSecond.toFixed(4);
 };
 
+// A 200 response with an empty array, a non-array body, or entries that all
+// lack an identifiable name/id is indistinguishable from "no models" to the
+// caller — it must be treated as a fetch failure, not a valid empty catalog,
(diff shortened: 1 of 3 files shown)


tomdacatto:
!polli review

Issue #13000, opened 2026-08-01
Quest Pollen Claim Issue


Marcus-Mok-GH:
This is probably related to the quest i helped fix, it says: “ 1 new quest completed 5 pollen ready to claim!”, and the quest to claim the pollen is just not there. Could you guys make it so we could claim the pollen with one “Claim all” button if possible? Or make quests not disappear if the quest has been completed and requires user to claim the credits? Thanks!

TRDEFI:
## Root-cause analysis (confirmed in code) ? this is the same bug I reported on PR #12714

I hit this exact issue: *"1 new quest completed, 5 pollen ready to claim!"* banner shows, but there is **no quest row and no Claim button** anywhere. After digging through the source, here's what's happening:

### Bug 1: Rewards with questId: null never render a claim button

The banner counts **all** unclaimed rewards (quest-overview.tsx L850-864), but the quest rows only map rewards that match a catalog quest:

- 
ewardByKey skips rewards without a questId: if (reward.questId) map.set(reward.questId, reward) (L719-725)
- Quest rows only get 
ewardId when 
ewardByKey.get(quest.id) matches (L740-741)
- The Claim button renders only when card.status === "claimable" i.e. a matched 
ewardId exists (L443-444, L505-515)

So any reward created **manually by an admin** (or with a questId that doesn't match a catalog quest) produces the "ready to claim!" banner but an **invisible claim path**. In my case: reward id af69b09-cdb3-479f-9a6f-077dc816eace ("Status notice contribution thank-you", 5 pollen) has questId: null.

### Bug 2: Backend blocks API-key claim, so the only path is the broken UI

- POST /api/quests/rewards/:id/claim requires a **dashboard session** ? API keys get 403 (quests.ts L237-290)
- claimReward itself (shared/billing/rewards.ts L116) works fine with 
ewardId + userId and ignores questId ? the orphaned reward is claimable if you can authenticate

### Suggested fixes (any would unblock everyone):

1. **Frontend**: render a claim path for unclaimed rewards with questId: null ? either a "Claim all" button in the banner (claiming every unclaimed reward by id), or an "Unassigned rewards" section with per-reward Claim buttons. I can open a PR for this.
2. **Backend**: none needed ? claimReward already handles orphans.

### Separate but related: issue bounties only pay when the issue is assigned to you

[...]

PR #12714, opened 2026-07-23
feat: add admin-controlled dashboard status notice

TRDEFI:
Fixes #12585

Add admin-controlled dashboard status notice feature. Administrators can publish, update, or clear a dashboard-wide status notice.

### Backend API
- `GET /api/status-notice` - Get current notice (public)
- `PUT /api/status-notice` - Set/update notice (admin only, requires PLN_ENTER_TOKEN)
- `DELETE /api/status-notice` - Clear notice (admin only)

### Frontend
- `StatusNoticeBanner` component displayed at top of dashboard
- Users can dismiss notices temporarily (returns after refresh)
- localStorage-based dismiss persistence
- Warning-styled banner with optional link support

### Tests
- Unit tests for status notice routes

## Acceptance Criteria
- ✅ Only authorized administrators can publish, update, or clear the notice
- ✅ An active notice supports a concise message and an optional safe link
- ✅ The notice is visible across dashboard pages and absent when inactive
- ✅ Users can dismiss it temporarily, but it returns after refresh while still active
- ✅ The notice is accessible, responsive, and rendered safely
- ✅ Focused tests and Biome pass on modified files

References #10482

PR #13027, opened 2026-08-02
fix(quests): add Claim all button so orphaned rewards (questId: null) can be claimed

TRDEFI:
Fixes the orphaned-reward claim problem reported in #13000 and on #12714.

## Problem

Rewards with questId: null (e.g. admin-created thank-you rewards) never render a Claim button: the quest rows only map rewards that match a catalog quest, while the "N pollen ready to claim!" banner counts **all** unclaimed rewards ? including orphans. Result: the banner shows pollen that can never be claimed from the UI.

## Fix

Adds a **Claim all** button to the claimable banner that claims every unclaimed reward by id in one pass (backend claimReward already handles questId: null). This unblocks orphaned/manual rewards and is also a general UX improvement for anyone with multiple pending rewards.

Verified: iome check clean, 	sc --noEmit clean.


TRDEFI:
Hi @voodoohop! This PR was cleaned up: removed the accidentally included carouselcraft-ai files and apps/apps.json change, rebased onto the latest main. It now contains only the quest fix (Claim all button for orphaned rewards with questId: null) as a single commit. Mergeable and ready to review. Thanks!

voodoohop:
Thanks for digging into this one — your diagnosis was right, and the root cause got fixed upstream in #13470 while this was open.

Rather than a bulk-claim button, `quest-overview.tsx` now derives cards straight from the reward ledger and renders them in a "Bonus rewards" lane, so every unclaimed reward gets its own Claim button. The filter is `questId == null || !catalogIds.has(questId)` — that also covers rewards whose quest card disappeared from the catalog (e.g. a POLLEN-QUEST issue losing its merged-PR reference), not just manual grants.

Already live in production, so closing this as superseded. Good catch on the banner/card mismatch.

PR #13067, opened 2026-08-04
fix: return 400 for Trellis validation errors

ElliotEtag:
## Summary

- Maps InferencePort HTTP 400 and 422 validation failures to public HTTP 400 for Trellis.
- Keeps authentication, routing, rate-limit, and upstream server failures on their existing paths.
- Adds adapter and worker-level coverage proving invalid image dimensions no longer surface as 502.

## Root cause

The InferencePort client mapped every upstream status except 402 and 429 to 502. InferencePort documents 422 as a request validation error and may return 400 for invalid image properties, so client mistakes polluted official model health.

## Validation

- `npx biome check --write` on changed files
- `npx vitest run test/model3d` (32 tests)
- `npm run typecheck`


g33ky00:
Thanks for the focused fix, @ElliotEtag.

- The classification change is clear and low-risk: 400/422 now correctly surface as client validation errors instead of 502, which should stop false negatives on Trellis health.
- The added tests cover both the adapter-level mapping and the worker-level behavior, which is the right surface here.
- Existing auth/rate-limit/upstream-failure paths appear preserved.

If the maintainers agree with the direction, it would be helpful to mark this PR as ready for review so CI/review can proceed.

PR #13071, opened 2026-08-04
fix: migrate Trellis to InferencePort async jobs

ElliotEtag:
- Submits Trellis 2 generations through InferencePort's async API and polls the job endpoint instead of using the synchronous transport.
- Keeps the Pollinations API synchronous while allowing up to five minutes for the upstream job.
- Returns 502 for failed or malformed jobs and 504 when the total deadline expires.
- Covers submission, polling, terminal states, timeout, status mapping, fallback, and resolution handling.

Model contract: canonical `trellis-2`; aliases low/medium/high; multiplier 1; not paid-only; no Pollinations GPU; InferencePort `trellis2`; no fallback.

Validation:
- 35 model-3D tests pass
- Typecheck passes
- Local authenticated E2E returned HTTP 200 with a valid 4.2 MB GLB after 165s and billed $0.24

Depends on #13067. No deployment included.


sharktide:
Fix #12544

PR #13073, opened 2026-08-04
fix: standardize model display titles

ElliotEtag:
- Standardizes 52 first-party model display titles across text, image, audio, and embeddings.
- Uses the approved official product spelling and exact variant labels, including Preview, Pro, Instruct, and version qualifiers where applicable.
- Changes display metadata only: canonical IDs, aliases, pricing, providers, routes, capabilities, and descriptions are unchanged.
- Keeps this work separate from the canonical model-ID rename; `APIDOCS.md` is untouched.

Checks:
- `npx biome check --write` on all changed registry files
- Gen: 80 files passed; 841 tests passed; 6 skipped
- Gen typecheck
- SDK model tests and typecheck
- Structural diff guard confirming every changed line is a `title` field


ElliotEtag:
- Superseded by #13074, which keeps only exact, human-readable display-title corrections.

PR #13078, opened 2026-08-04
fix: let publishable keys edit redirect URIs and earnings settings

### LINKED ISSUE #10983: Cannot edit details of publishables
**Reported by:** `morrisdweck`

**Issue:** Unable to edit details of publishables.

**Details needed:**
- What specific details are you trying to edit? (API key settings, profile info, etc.)
- What error message (if any) appears?
- Steps to reproduce
- Browser/environment info

---
*This issue was filed via Discord request. Additional information needed from reporter.*

---
**Author:** `morrisdweck` (UID: `930274006869884928`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/889573359111774329/1504906045062119434)

davealan74:
- `EditApiKeyDialog` gated both the redirect-URI/earnings section and the metadata POST on `isAppKey`, which requires a key to *already* have a redirect URI or earnings enabled. A plain publishable key could therefore never get its first one: the section was hidden, and edits were silently dropped on save.
- Both gates now use `isPublishableKey`, so the save gate can no longer be narrower than what the user is shown.
- Moved the metadata POST ahead of `onUpdate()`. `onUpdate` ends in `router.invalidate()`, so the list refetched before the metadata write landed and the dialog closed onto a stale card.
- Moved the shared key predicates into `key-type.ts`; their logic is unchanged.

Notes:

- Visible UX change: every publishable key now shows the redirect-URI/earnings section, and saving a first redirect URI moves the card from the "API keys" group to "App keys".
- The POST-before-`onUpdate` ordering has no automated guard — `enter.pollinations.ai` runs the Workers vitest pool with no DOM environment, so `handleSave` can't be rendered. It carries an inline comment; the gating logic it depends on is unit-tested.

Test: `cd enter.pollinations.ai && npx vitest run test/edit-api-key-dialog.test.ts` — 6 passed.

Fixes #10983


davealan74:
@ElliotEtag you have the most recent commits on `edit-api-key-dialog.tsx`, so flagging the two behaviour changes: the redirect-URI/earnings section is now gated on `isPublishableKey` rather than `isAppKey`, since the old gate meant a publishable key could never receive its first redirect URI, and the metadata POST now runs before `onUpdate()` so the `router.invalidate()` inside it does not refetch the list ahead of the write.

PR #13095, opened 2026-08-05
refactor: preserve text request and response pass-through

voodoohop:
- Passes validated chat and simple-text options through without rebuilding an allowlisted request.
- Preserves provider/message extensions, every upstream choice, and response extensions while retaining required provider compatibility repairs.
- Strips only private cost/search_context_size usage fields, uses epoch seconds for missing created, and validates the actual upstream SSE content type.
- Deletes the duplicate request parser, option normalizer, parameter validator, and lossy message sanitizer: 140 fewer production-source lines.

Checks:
- Gen TypeScript typecheck
- test/text: 265/265
- generation-vcr.test.ts: 27/27
- Biome and git diff --check


voodoohop:
- Superseded by #13766, reconstructed from current main to keep the text cleanup behavior-preserving and exclude encrypted-secret history.\n- #13766 retains the typed request projections, provider extensions, immutable all-choice handling, epoch-second created fallback, and private-usage stripping.\n- Existing request defaults and downstream response validation remain unchanged.

PR #13112, opened 2026-08-05
fix: set Myceli proxy timeout to 300 seconds

ElliotEtag:
- Adds a manual production-only workflow for the `myceli.ai` zone.
- Reads and reports the current `proxy_read_timeout` before changing it.
- Sets the value to 300 seconds and verifies it through the Cloudflare API.
- Fails without changing anything if the zone is missing, the setting is not editable, or the existing credential lacks permission.

Validation: Prettier, `git diff --check`, and actionlint.


ElliotEtag:
Closing: production only accepts promotion PRs from main. Reopening this change against main.

Issue #13052, opened 2026-08-03
feat: launch the four built-but-hidden quests


voodoohop:
Four quests are implemented but hidden and non-grantable:

- `github_established`
- `github_stars`
- `early_adopter`
- `app_paid_request`

Each carries the comment "Built but not launched — hidden from the UI, not grantable" in `enter.pollinations.ai/src/services/quests/groups/`.

Turning them on is the cheapest way to put Pollen back in active users' hands while staying cost-cautious. Raised in the 2 Aug call: users report having nothing to spend, and API usage is down because of it.

Next round after the previous enablement PRs — #12088 (`app_listed`) and #12393 (`use_app`, `app_active`).

Related trackers, none of which names these four: #11677, #11374, #11899.

PR #13132, opened 2026-08-06
fix(quests): enable four built-but-hidden quests (#13052)

g33ky00:
- early_adopter: available
- github_established: available
- github_stars: available
- app_paid_request: available


YoannDev90:
No

PR #13136, opened 2026-08-06
fix: recheck quests when the page becomes visible

### LINKED ISSUE #12474: Quests not automatically giving rewards and sometimes taking multiple times to register
Sometimes, when you do quests, it doesnt register, and you need to do it multiple times for it to register the quest.

davealan74:
## Summary

Fixes #12474

- Rechecks quests when the page becomes visible.
- Drops overlapping checks and retries one throttled check using `Retry-After`.
- Refreshes rewards after a successful check.
- Keeps the flow frontend-only: no polling or backend changes.

## Validation

- Enter typecheck
- Production frontend build
- Biome on the changed file


davealan74:
@voodoohop this implements the four points you specified: recheck on visibilitychange, in-flight dedup, single Retry-After retry on 429, and a rewards refresh after a successful check. Frontend-only, one file — no server changes.

voodoohop:
- Thanks @davealan74 — the investigation and focused implementation were useful.
- Quest status changes infrequently enough that a manual reload is sufficient for now, so we're keeping the frontend simpler and closing this.

PR #12921, opened 2026-07-31
refactor: remove obsolete Gen and registry plumbing

voodoohop:
- Reconstructs the useful part of the stale branch on current main; keeps schema types and logging that now have real callers.\n- Removes the unused Portkey multi-target request path and success-header reconstruction; the Worker fallback loop remains the sole retry owner.\n- Rejects empty generic and Azure response streams as retryable 502s instead of fabricating successful SSE output.\n- Removes unused internal exports and duplicate registry billing entry points: 132 fewer production lines, 264 fewer lines overall.\n\nValidation:\n- Biome and Gen typecheck\n- 109 focused fallback, text-client, tracking, and observability tests\n- Full Gen run: 104 files / 1,248 tests passed; the remaining community file passed 80/80 after supplying its Enter-only Stripe test dependency

Issue #12212, opened 2026-07-03
Perplexity streaming returns empty stream (no chunks, no usage, nothing billed)


voodoohop:
- `stream: true` on any perplexity alias returns only `data: [DONE]` (14 bytes) — no content chunks, no usage
- Reproduced on production `gen.pollinations.ai` AND staging — pre-existing, unrelated to #12210
- Worker logs: `No usage object found in event stream` → `Failed to extract model usage` → tracked as `is_billed_usage: false` (request is free)
- Non-stream Perplexity works fine
- Likely in the Portkey gateway Perplexity stream transform; note Perplexity puts `usage.cost` only in the final SSE chunk

Repro:
```bash
curl https://gen.pollinations.ai/v1/chat/completions -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" -d '{"model":"perplexity-fast","stream":true,"messages":[{"role":"user","content":"hi"}]}'
```

Related: #11703 (usage.cost passthrough)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gggff123:
Closing this Hope it wasnt a tracking issue

Issue #11675, opened 2026-06-08
fix: fail closed when billable text usage is missing


voodoohop:
- Missing usage should fail closed for all billable text generation, not only community endpoints.
- Current streaming tracking can return successful responses with `modelUsage: null` when no usage object is found, which means no billing.
- Desired behavior: if a billable response cannot produce usage, return or raise an error instead of serving unbilled success.
- Scope: apply to community and non-community text paths, streaming and non-streaming where usage is required for billing.
- Add focused tests for missing usage on streamed and non-streamed text responses.
- Keep separate from the community models PR to avoid expanding rollout scope.

PR #13146, opened 2026-08-06
fix: mark text rows that completed with no billable usage

davealan74:
- A 200 whose usage can't be extracted and whose model has no determinable fee fell through to a plain unbilled row — indistinguishable in Tinybird from a legitimately free one.
- Adds `errorTracking` to `ResponseTrackingData`; the `!modelUsage` fallthrough sets `errorResponseCode: "usage_missing"` for `generate.text`, and settlement `emitRow` prefers it over the status-derived data.
- `adjustmentOnlyBilling` keeps first refusal, so models with a knowable flat fee are unaffected. #12212's `perplexity-fast` repro is already covered by that mitigation (it bills the static search fee) — this closes the residual hole where no charge is determinable at all. Nothing on the Portkey side changes.
- Telemetry only: the client still gets its 200. For streams the bytes are on the wire before tracking runs; pre-send empty-stream → 502 is #12921. New row shape for dashboards: `responseStatus` 200 with a populated `error_response_code`.
- Extends `test/tracking-observability.test.ts` (33 pass): non-streaming, streaming, streaming via community endpoint, plus a flat-fee regression guard.
- #13086 rewrites these lines (`notBilled` → `tracked`); needs a rebase depending on merge order.

Addresses #11675


davealan74:
@voodoohop the `!modelUsage` fallthrough now stamps `errorResponseCode: usage_missing` on `generate.text` rows, so a 200 that billed nothing and could bill nothing is queryable instead of looking like an ordinary free row.

Cases covered in tests: non-streaming, streaming, streaming via a community endpoint, plus a `perplexity-fast` guard confirming the flat-fee path still bills exactly as before.

PR #13154, opened 2026-08-06
feat: scan community requests for secrets locally

### LINKED ISSUE #13150: Enable safe=privacy,secrets by default for community models + add in-memory reversible redaction vault
## Summary

Split out from #12987. That issue scoped a full benchmark of local PII libraries vs AWS Bedrock — premature. The detection engine (Bedrock Guardrails) works and is accurate; the actual leak is that `safe` defaults to **off**, so community-model calls with `safe` unset send raw prompts (API keys, emails, phones) straight to the provider. Fix the enforcement gap and the reversibility gap first; benchmark only if shipped-enforcement shows real cost/latency pain or false-positive complaints.

### 1. Default-on enforcement
Auto-enable `safe=privacy,secrets` for community models, with an opt-out. This is the "separate, tracked change" already referenced in #12987's context.

### 2. In-memory reversible redaction vault
Bedrock returns `redactedIds` as a list of **types** (`EMAIL`, `PHONE`, …), not matched values — there is no restore mechanism today; `safety.ts` just splices `response.outputs[].text` back in. Add a per-request in-memory Map (match value ↔ placeholder) so redacted entities are restored in the final response. No persistent store, no cross-request state.

## Acceptance criteria

- [ ] Community-model requests auto-apply `safe=privacy,secrets` when `safe` is unset; explicit opt-out respected.
- [ ] Reversible vault restores original PII values in responses; redacted text never reaches the end user.
- [ ] **Correctness trap — vault MUST respect the `selectGuardrailInputs` sliding window.** Only the last ~50k chars / 25 parts are scanned by Bedrock; earlier turns pass through unscanned. The restore Map must be built and applied against the *scanned window*, not the full `texts[]`, or unredacted context gets corrupted. Mirror the existing `offset`/`originalIndex` splicing logic in `safety.ts`.
- [ ] No cross-request state; vault is request-scoped and GC'd.
- [ ] Latency/cost impact measured on merge (lightweight — enough to inform a later benchmark go/no-go, not a full study).

## Out of scope (deferred to a conditional benchmark issue)
- Swapping Bedrock for a local/self-hosted lib (Presidio, pii-vault, redact-core). Only revisit if this ships and shows pain. See #12987 for the candidate analysis.

## Context files
- `gen.pollinations.ai/src/middleware/safety.ts` — `applyGuardrail`, `redactedIds` handling, `response.outputs` splicing
- `gen.pollinations.ai/src/utils/safety-features.ts` — `FEATURE_TRIGGERS.secrets` / `FEATURE_TRIGGERS.privacy`
- `selectGuardrailInputs` — sliding-window input selection (the correctness constraint above)

/cc @the_yerminator — implementing

---
**Author:** `Yerminator (@the_yerminator)` (UID: `1123534156626939945`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1534941140514635987/1534942258787258521)

YoannDev90:
Fixes #13150

- Enable privacy+secrets safety by default for community models (explicit opt-out via `safe=`)
- Scan request text in-worker with vendored redact wasm (no Bedrock roundtrip, no unscanned window)
- Block requests containing secrets (AWS keys, JWTs, credit cards, API tokens) with checksum validation
- Keep Bedrock guardrail for privacy filtering on texts that pass the local scan
- Add redact-secrets + safety tests (14/26), exclude vendored wasm glue from biome

Verified:
- typecheck green
- 47/47 tests (redact-secrets, safety, model-permissions)
- build:staging emits wasm asset + bare import
- E2E in workerd: detects AWS_ACCESS_KEY, JWT_TOKEN, CREDIT_CARD


YoannDev90:
Should I convert this into a draft ?

YoannDev90:
Reopening to re-trigger CI on the fixed head

sharktide:
what if the user really needs to send an api key to a model? besides that's the user's fault

sharktide:
if we catch false positives or placeholders it could break coding models

YoannDev90:
This PR comes with restoration feature, so your api key appear as a placeholder for the model, then it restores the api key, the model never seen the key but it should work. Anyway you can opt in by setting safe=false in the request.

YoannDev90:
Any update ?

voodoohop:
- Closing: default-on scanning adds a Bedrock call to every community request, and we cannot ship a prebuilt wasm without its build source
- Opt-in \`safe=secrets\` already covers users who want it
- Thanks for the work

PR #13076, opened 2026-08-04
refactor: standardize canonical model names

ElliotEtag:
- Promotes 158 public model IDs, preserves aliases, and replaces catalog `brand` with `publisher`. `brand_url` stays unchanged. Removes the upcoming-rename Enter banner; API docs retain the migration explanation.
- Keeps 49 provider-qualified hidden fallback IDs, existing routing, pricing, access, upstreams, and tracking semantics. Primary execution-ID work remains deferred to #14543.
- Preserves canonical permission writes and migration `0062`; normalizes permission reads. Production remains gated until live compatibility verification, with final cleanup after both workers deploy.
- Fixes app alias lookups, community-ID collisions, Monitor/Enter consumers, and provider-month Economics identity reconciliation. Removes `MODEL_SLUGS.md`; updates model-management guidelines. No Tinybird or Observability deployment/change.
- Compatibility #14545 and promotion #14546 are live. Enter/Gen versions and production bindings verified; no migrations were pending. All 158 future IDs appear in production's alias catalog. Nine live checks pass: old/new/community requests, denied models, catalogs, readback, unchanged storage, expiry and budget.

Staging proof: all 159 SQL statements tested on 150,000 synthetic rows. Migration 11.7s; late-writer cleanup 8.7s; repeat wrote zero rows. All expected permissions matched; test table removed; real keys untouched.

Validation: prior full Gen 1,841 passed/6 skipped; Enter 1,113 passed; 73 compatibility tests. Frontend build/Biome pass. Main merge: 15 permission tests and Gen typecheck pass. Live verification complete; gate-variable approval and Thomas sign-off remain before promotion.

PR #13197, opened 2026-08-08
fix: make model allowlists alias-aware

ElliotEtag:
- Adds the approved 128 future canonical IDs as compatibility aliases; no canonical names change in this PR.
- Important: deploying this PR publishes those IDs as public API surface through model alias listings and model-derived documentation. The names are final at this deployment, not at #13076.
- Canonicalizes known IDs when API-key allowlists are created, updated, and read back, while preserving community and unknown IDs.
- Makes Gen authorization, fallback selection, realtime validation, and catalog filtering compare resolved model identities.

Validation:
- Promotion set matches #13076's migration map exactly: 128/128
- No collisions across 223 live public/community model IDs
- Full Gen: 83 files passed; 871 tests passed, 6 skipped
- Full Enter: 737 passed; 4 pre-existing API-key fixture failures reproduced unchanged on the parent commit
- PR-focused Enter: 334 passed
- Enter and Gen typechecks, Biome, and diff checks passed

Deployment gate: merge and verify this compatibility layer in production before rebasing or merging #13076. The rebased rename PR must add a mirror test against `CANONICAL_MODEL_PROMOTIONS`.


UCIRANX:
👌

PR #13125, opened 2026-08-05
feat: add per-user rate limit for community models

YoannDev90:
## What
One user can exhaust an upstream quota for everyone on a community model (502 / upstream 429). Adds a per-caller rate limit on community-model calls, derived from the endpoint's declared RPM.

## Changes
- **DB/API/dashboard**: `rateLimitRpm` column + migration, surfaced through the community-endpoints route and dashboard form
- **Limit derivation**: per-user budget = 25% of declared RPM (default 120 RPM → 30 req/min/caller), in `communityEndpointPerUserRateLimitRpm`
- **Enforcement**: new `CommunityModelRateLimiter` Durable Object (fixed 60s window, one DO per caller+endpoint) + `communityModelRateLimit` middleware
- **Wiring**: middleware inserted after `generationAccess` in all 6 proxy chains (chat, simple text, image, edits)
- **Caller identity**: user id → API key id → IP
- **Tests**: helper, DO, middleware (pass-through, within budget, 429 with Retry-After, independent caller budgets)

Fixes #13030

PR #13257, opened 2026-08-10
feat: add per-user RPM for community models

### LINKED ISSUE #13030: Community models: shared upstream rate limit lets one user exhaust quota for everyone (502 / upstream 429)
## What happened

A community model returned a 502 to end users because its **upstream provider** hit a rate limit. The 429 originates at the community model's own upstream, not at Pollinations:

```json
{
  "upstreamStatus": 429,
  "upstreamHost": "portkey.myceli.ai",
  "upstreamBody": "{\"error\":{\"message\":\"openai error: Rate limit exceeded\",\"type\":\"rate_limit_error\",\"param\":null,\"code\":\"rate_limit_exceeded\"},\"provider\":\"openai\"}",
  "status": 502
}
```

Reported in Discord by `ltlani` — "someone is ruining it for everyone."

## Root cause

Community models are user-hosted OpenAI-compatible endpoints (registered via `/account/my-models`) routed through the gateway to a single shared upstream credential per model. The gateway correctly remaps the upstream 429 → 502 via [`remapUpstreamStatus`](https://github.com/pollinations/pollinations/blob/main/shared/error.ts) (`shared/error.ts`), and 429 is in [`FALLBACK_ON_STATUS_CODES`](https://github.com/pollinations/pollinations/blob/main/gen.pollinations.ai/src/fallback.ts) so it fails over to declared fallbacks — but:

- If the model has **no fallbacks** declared (or all fallbacks share the same upstream), the 502 surfaces directly to every caller.
- There is **no per-user isolation** on the gateway side for community models. All Pollinations users calling a given community model share that model's single upstream OpenAI/account quota. One heavy user can exhaust it, and every other user gets 502s until the upstream window resets.

This is distinct from Pollinations' own rate limits (`EDGE_RATE_LIMITER` ~10 req/s per IP in `gen.pollinations.ai/src/middleware/rate-limit-edge.ts`, and the pollen 1/IP/hour limiter), which return a plain `429` and are not the source of this error.

## Impact

- A single abusive/heavy caller can take down a community model for **all** users until the upstream rate window resets.
- The model owner's upstream bill/quota is consumed by other users with no per-user cap.
- Community model owners have no tooling to cap or isolate per-Pollinations-user throughput on their model.

## Suggested directions

- Per-user (or per-API-key) rate limiting on community model calls on the gateway side, independent of the edge IP limiter, so one user can't monopolize a shared upstream.
- Surface upstream 429s to the model owner (dashboard/alerting) so they can detect abusive callers and configure fallbacks.
- Encourage/require fallback targets for public community models so a single upstream 429 doesn't become a hard 502.
- Consider whether community models with no fallbacks and a shared upstream should be eligible for public listing, or require a per-user cap.

## Context

- Fallback logic: `gen.pollinations.ai/src/fallback.ts` (`FALLBACK_ON_STATUS_CODES`, `isRetryableFallbackError`)
- Status remap: `shared/error.ts` (`remapUpstreamStatus` — 429 → 502)
- Community endpoint definitions: `shared/community-endpoints.ts`
- Edge rate limiter: `gen.pollinations.ai/src/middleware/rate-limit-edge.ts`

---
**Author:** `Ouadie (@ltlani)` (UID: `1468504208423780456`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/889573359111774329/1533607577408245860)

voodoohop:
## What

- Lets each community model owner set the exact RPM allowed per Pollinations user, including decimals such as `0.5` (one request every two minutes).
- Leaves the model uncapped when blank; genuine upstream capacity failures remain model failures.
- Enforces the selected model’s own token bucket immediately before each text or image attempt, including fallback targets.
- Returns `429` + `Retry-After` when the caller reaches that model’s limit. This local limit does not cascade to another fallback or reduce upstream-health scores; genuine upstream failures still do.
- Exposes `per_user_rpm` in the rich model metadata and OpenAI-compatible `/v1/models`, and shows it in My Models and the model catalog.

## Simplicity

One nullable `REAL` field, one small Durable Object, and one pre-attempt enforcement hook. No implicit 120 RPM capacity, percentage share, identity fallback, total model cap, or owner exception. Cloudflare’s native limiter cannot accept a dynamic per-model rate.

The large generated diff is the required Drizzle snapshot; the migration itself adds one nullable column.

## Checks

- Focused API persistence/reset, UI serialization, Durable Object, catalog, and fallback integration tests
- Gen and Enter typechecks
- Biome 2.3.10

Fixes #13030
Supersedes #13125


YoannDev90:
Nice — this is a cleaner shape than my draft, and thanks for the co-author credit. The nullable `perUserRpm` is definitely simpler than a derived share. Two points I'd love to settle before this ships, both grounded in what motivated #13125:

**1. Blank = no limit leaves the failure mode from #13030 wide open.**

```ts
if (!endpoint || endpoint.perUserRpm == null) return;
```
(`gen.pollinations.ai/src/utils/community-model-rate-limit.ts`)

A model whose owner leaves the field empty gets zero Pollinations-side protection. I expect most owners will leave it empty:
- the field is optional, and the dashboard gives no hint that leaving it blank means "unprotected";
- the cost of being wrong (upstream quota exhaustion) is invisible until the moment the model starts failing — there's no feedback loop that teaches the owner to set it.

And when it happens, it's not one user getting throttled — the shared upstream quota is exhausted for *everyone*, which is exactly #13030's scenario (502 / upstream 429 for all callers).

This is not hypothetical for me: my model was deliberately taken down by a request flood that exhausted my upstream quota — every subsequent call failed for all users. Same failure class.

**Would you consider a default-on total cap layered underneath?** The idea:
- same Durable Object, but one bucket keyed by `endpoint.id` alone (no user in the key), so it caps *total* requests to the upstream per window, regardless of who sends them;
- it enforces the owner's declared upstream capacity, falling back to a conservative default (e.g. 120 RPM) when unset — so the protection exists even for owners who never touch the form;
- enforcement happens *before* the upstream is ever hit: excess requests get a clean `429` + `Retry-After` + `community_model_rate_limit` errorCode, the upstream never sees the flood, and the model never gets flagged as failing;
- the gateway's `429` goes through the existing `UpstreamError` path with a custom `errorCode` (`shared/error.ts:50`), so clients can distinguish "your budget exhausted" (`RATE_LIMITED`) from "this model is at capacity".

This complements `perUserRpm` rather than competing with it: per-user limits handle fairness between legit users; the total cap protects the upstream itself. Per-identity limits can't stop a flood with rotated IPs or fresh accounts — only a model-level cap can.

**Side note** (no blocker): once a total cap exists, per-user could be *derived* from it — a share of the declared upstream capacity — instead of being a separate field. `perUserRpm` would then become an optional override for owners who want stricter per-user limits than the share gives (e.g. a 1000 RPM model where you still want 10 RPM per user). That's the one case a fixed share can't express.

I can implement the total cap on this branch if the principle works for you.

**2. Upstream 429 being remapped to 502 hurts the very owners this protects.**

```ts
const remapTo502 = new Set([401, 402, 403, 404, 409, 415, 429]);
```
(`shared/error.ts:376`)

A `502` means "bad gateway / we couldn't reach the backend" — but the upstream *answered* 429, it just ran out of quota. Remapping it to 502 conflates two very different states:
- **model at capacity** (transient, retry later, not the caller's fault, *not* an outage);
- **model down** (real availability problem, deserves alerting).

Consequence: a community model whose upstream quota is temporarily exhausted looks identical to a dead model in health metrics — and that's what flagged my model as failing after the flood. The stability drop was a measurement artifact, not a real defect.

Since `docs/apidocs-recipes.md:184` already tells clients to watch `429`/`503` + `Retry-After`, keeping the upstream `429` as a `429` (preserving the original `upstreamStatus` and adding a distinguishing `errorCode`) is:
- more honest — the upstream answered, and "too many requests, back off" is what it said;
- action-ready — clients already know to honor `Retry-After`;
- clean for observability — `handleError` doesn't emit server-error telemetry for `429` (`< 500`, `shared/error.ts:257`), so a busy model stops counting as an incident.

Happy to pull `429` out of that remap set if you agree.

Happy to keep #13125 open meanwhile — I'll close it once we've settled the direction here.

PR #13266, opened 2026-08-10
feat: deduplicate video generation requests

voodoohop:
- Adds one SQLite Durable Object per existing media cache key for video GETs.
- Re-checks R2 inside the object, elects one leader, and runs the current video generator unchanged from an alarm.
- Waits for the R2 write before releasing callers; joined callers return `X-Cache: HIT` / `COALESCED`, avoiding duplicate billing and community rewards.
- Fails closed on alarm redelivery instead of submitting twice, then expires terminal state with the 30-day R2 lifecycle.
- Keeps the synchronous 200 response contract, provider implementations, cache-key identity, and seed behavior unchanged. No 202 API, provider job IDs, or new generation deadline.

Testing:
- `biome 2.3.10 check`
- `tsc --noEmit`
- Full Vite Worker build
- Added Durable Object and cache interaction tests. The local Workers Vitest harness exits before collecting both these tests and an unchanged baseline test (`usage-headers.test.ts`).

### DIFF
--- a/gen.pollinations.ai/src/durable-objects/MediaGeneration.ts
+++ b/gen.pollinations.ai/src/durable-objects/MediaGeneration.ts
@@ -1,0 +1,152 @@
+import { DurableObject } from "cloudflare:workers";
+import { UpstreamError } from "@shared/error.ts";
+import { R2_LIFECYCLE_TTL_MS } from "@shared/r2-storage.ts";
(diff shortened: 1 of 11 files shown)


voodoohop:
Closing in favor of a fresh, API-generic design. This implementation coupled durable orchestration to the media handler and would be the wrong base for generic request deduplication.

PR #13274, opened 2026-08-10
fix: distinguish fallback spillover from degradation

ElliotEtag:
- Adds `primary_5xx`, `primary_retried_503s`, and `fallback_rescues` to `model_health`.
- Keeps DEGRADED based only on final user-visible outcomes.
- Reports non-final primary 503 pressure separately as SPILLOVER.
- Preserves the current hide/relist monitor workflow and community rate-limit exclusions.

Root cause: the image-specific monitor query counted non-final Z-Image Vast 503 attempts even when Fal handled the final response.

Checks:
- `git diff --check`
- Proposed Tinybird SQL executed against live data: final 5xx remained zero while retried primary 503s were reported separately

Tinybird has not been deployed. Rollout: validate/deploy staging, merge, deploy production, then verify the monitor auto-refreshes from `main` on its next cycle.


voodoohop:
@ElliotEtag don't the consumers of this pipe like model monitor etc need to be adapted?

Issue #10944, opened 2026-05-14
Unify size handling when input image is provided to /v1/images/edits


voodoohop:
Current behavior in `gen.pollinations.ai` (`createAndReturnImages.ts`) defaults missing `size` to `1024x1024` even in edit mode, which biases edit-model output toward square regardless of input aspect.

- Edit mode is invoked when `image` is present in the request body
- When the caller omits `size`, gateway plugs in `1024x1024` and then resolves to a model-supported size (e.g. for gpt-image-2: picks closest of `1024x1024 / 1536x1024 / 1024x1536` based on the passed ratio — which is always 1:1)
- Result: portrait/landscape inputs get reshaped to square on every edit

The user expectation in an edit workflow is that the model should respect the input image's aspect (and ideally its resolution) by default, since this is *editing* not *generating*.

Open question — best dev UX:
- Skip passing default size to upstream model when input image is provided (let model decide)
- Or auto-derive size from the input image's actual dimensions
- Or pass aspect-snapped size only (let model pick resolution)

Only override when the caller explicitly passes `size`.

Surfaced while building `apps/voice-edit` — every voice-pointer edit was getting reshaped to square instead of preserving input aspect.

cc @voodoohop

aikhusus2025-ctrl:
[solusi-issue-10944.md](https://github.com/user-attachments/files/30931942/solusi-issue-10944.md)

voodoohop:
- Resolved by #13459.
- Image edits now preserve the source aspect ratio when `size` is omitted, while explicit sizes remain unchanged. Thanks!

PR #13293, opened 2026-08-11
fix: preserve source image aspect ratio for image edits

### LINKED ISSUE #12583: [QUEST] Preserve aspect ratio for image edits
Original issue: #10944

## Goal

## Acceptance criteria

## Quest

### Reward

5

davealan74:
Fixes #12583
References #10944

## What

When a `/v1/images/edits` request omits `size`, the response previously defaulted to the model's square resolution, discarding the source image's portrait or landscape ratio.

## How

- `deriveSourceImageSize` fetches the first source image and reads its pixel dimensions using the existing `readImageDimensions` utility (PNG, JPEG, GIF, BMP, WebP).
- `scaleToEditSize` scales those dimensions to ~1 MP while preserving the aspect ratio, snapping each side to the nearest 16-pixel multiple — the same granularity providers expect.
- The derived size string is passed through `resolveParams` identically to an explicit caller-supplied `size`. Falls back silently to model defaults if the image cannot be fetched or its dimensions cannot be read.
- Requests with an explicit `size` are not affected.

## Changes

- `gen.pollinations.ai/src/routes/images.ts` — two new helpers (`scaleToEditSize`, `deriveSourceImageSize`) wired into `handleImageEdit`; new imports for `fetchUserImage` and `readImageDimensions`.
- `gen.pollinations.ai/test/image/imageEdits.test.ts` — unit tests for `scaleToEditSize`: portrait/landscape preservation, square passthrough, 16-px snap invariant, ~1 MP area target.

## Validation

- Biome: no issues on modified files.
- Tests: unit tests for the pure scaling logic pass. Full vitest suite requires `workerd` (glibc 2.35+) not available in this environment; CI will verify.

### DIFF
--- a/gen.pollinations.ai/src/image/models/novaCanvasModel.ts
+++ b/gen.pollinations.ai/src/image/models/novaCanvasModel.ts
@@ -2,7 +2,11 @@ import debug from "debug";
 import { getImageEnv } from "../env.ts";
 import { HttpError } from "../httpError.ts";
 import type { ImageParams } from "../params.ts";
-import { base64ToBuffer, downloadUserImage } from "../utils/imageDownload.ts";
+import {
+    base64ToBuffer,
+    downloadUserImage,
+    readImageDimensions,
+} from "../utils/imageDownload.ts";
 
 const logOps = debug("pollinations:nova-canvas:ops");
 const logError = debug("pollinations:nova-canvas:error");
@@ -67,11 +71,6 @@ export async function callNovaCanvasAPI(
         throw new HttpError("AWS credentials not configured", 500);
     }
 
-    const { width, height } = clampDimensions(
-        safeParams.width || 1024,
-        safeParams.height || 1024,
-    );
-
     // Check if image input is provided for editing mode
     const rawImageUrl = safeParams.image
         ? Array.isArray(safeParams.image)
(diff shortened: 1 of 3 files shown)


ElliotEtag (changes requested):
- Blocker: deriving `size` in the route makes `resolveParams` emit width/height, so `ImageParamsSchema` sets `dimensionsExplicit=true`. For Seedream 4.0 edits this selects the custom-size path, where both sides must be 1024–4096px. A 768×1024 source derives to 880×1184 and therefore returns 400 instead of using the working preset/match-input path. This also changes routing for other `dimensionsExplicit` consumers. Please carry derived dimensions without marking them caller-explicit, or derive the ratio where the source is already downloaded.

- Performance: affected adapters download or decode the first image once here for dimensions and again through `toDataUri`/`downloadUserImage`. The first fetch has no request signal or application timeout, and its swallowed failure can make a bad URL fail twice sequentially.

- Tests: the explicit-size test never calls production code; it only compares numbers parsed from a literal. Please add handler/model-path coverage for no-size portrait and landscape edits, explicit-size preservation, JSON and multipart inputs, and provider constraints. The PR checks currently do not run Vitest.

- The scaler also needs provider-aware bounds: for example, 4000×100 becomes 6480×160.

davealan74:
Addressed all four points in the follow-up commit:

- **Blocker + performance**: removed `scaleToEditSize`, `deriveSourceImageSize`, and the broken `@/userImage.ts` import entirely from the route layer. `handleImageEdit` now passes `size` directly from the request (or `undefined`) unchanged, so `dimensionsExplicit` is only ever set when the caller explicitly supplies a `size`.

- **Ratio preservation moved to model layer**: `novaCanvasModel` now reads image dimensions from the buffer it already downloads for the variation request. When `!dimensionsExplicit`, it passes those native dims (clamped to Nova Canvas constraints) instead of the 1024×1024 default. Seedream already uses `match_input_image` via `resolveAspectRatio` when `hasImage && !aspectRatio`; OpenRouter Seedream Pro already derives ratio from its own download — both unchanged.

- **Tests**: replaced the fake `scaleToEditSize` unit tests with `resolveParams` tests that exercise production code: no width/height emitted when size is omitted (confirming `dimensionsExplicit` stays false), portrait/landscape explicit-size propagation, and OpenAI quality-string mapping.

- **Provider bounds**: moot once the route-layer scaler is removed; `clampDimensions` in Nova Canvas already enforces 320–4096px and 4MP.

voodoohop:
- Thank you for working on #12583 — your contribution helped shape the final solution in #13459.
- Added a claimable 5-Pollen thank-you reward to your account. You can claim it from the [Quests page](https://enter.pollinations.ai/quests).

Issue #12889, opened 2026-07-30
Server-mint request_id instead of trusting the client header


voodoohop:
`requestId()` is mounted bare at `gen.pollinations.ai/src/index.ts:108`. At the pinned hono 4.10.6 the middleware accepts an inbound `X-Request-Id` verbatim (≤255 chars, `[\w\-=]`) and echoes it back, so any caller can pin unrelated requests to one id or replay one they saw.

Two consequences:

- **Cross-user read on Nova Reel.** `image/models/novaReelModel.ts:198` interpolates the id into a fixed S3 prefix and reads `{prefix}/output.mp4` back. Two callers sending the same id share the key.
- **Attempt grouping is corruptible.** After #12819, `request_id` is the only key joining a fallback attempt row to its settlement row.

Fix: a ~6-line `shared/middleware/request-id.ts` that always mints server-side and still sets the response header, mounted in gen and enter. Do *not* use `requestId({ headerName: "" })` — the same flag governs the inbound read and the echo, so it silently drops a header documented at `APIDOCS.md:2027`.

Not a blocker for #12819: no consumer there joins on `request_id`.

PR #13299, opened 2026-08-12
feat: deduplicate generation requests

voodoohop:
- Coordinates identical finite-generation cache misses through one Durable Object per shared cache identity; callers wait for and read the same completed R2 object.
- Covers non-streaming text/chat, embeddings, image, video, 3D, OpenAI images, image edits, and finite audio endpoints without provider-specific job IDs or generation deadlines.
- Authorizes every caller independently; only the detached generation performs billing and community rewards, independently of cache-write success.
- Keeps failure handling simple: no 202, unknown state, tombstones, deployment reconciliation, or direct-generation fallback.
- Preserves request IDs, balance attribution, fallback telemetry, multipart replay, and provider error responses across detached execution.
- Verified with Biome, TypeScript, the full gen suite (90 files; 910 passed, 6 skipped), and a live staging matrix including disconnect/rejoin, cache replay, and exactly-once billing.

Stacked on #13269.


voodoohop:
## Adversarial review — admission/execution split

Reviewed through `357a51eb4`. The split (`3c3ec7e14` + `84a1a465f`) went through 6 attack angles with independent refutation of every claim — 25 candidate findings, 23 refuted. The four later commits are reviewed in the addendum below. 59 focused tests pass at HEAD.

**The refactor is behaviour-preserving on the billing path.** Exactly one deduction per generation, in the executor. No dropped `waitUntil`, no unregistered cache write, no admission/executor route divergence. Both remaining findings are telemetry, and both are inherited from the PR base rather than introduced by these two commits.

### 1. Coordinated generations emit no row under the caller's request id

- `execute-generation.ts` — the executor mints its own id via `requestId()`; `shared/middleware/request-id.ts:17` is `crypto.randomUUID()`.
- `GenerationJob` (`generation-deduplication.ts:44-48`) has no id field, so the admission id never crosses the hop, and the executor's id never comes back — `x-request-id` is in neither `ERROR_HEADERS` nor R2's `CACHED_HEADER_NAMES`.
- Admission emits nothing for a coordinated request: success → `X-Cache: HIT` → early return at `track.ts:360-363`; failure → `COALESCED-ERROR` → early return at `track.ts:291`.
- Effect: a user reports "request `3f2a…` returned 502", and `generation_event_v2` has zero rows for that id — for both successes and failures. The id they hold is valid, it just has no row.
- Adding `x-request-id` to `EXECUTOR_HEADERS` will not help: `requestId()` deliberately never trusts an inbound header (`shared/middleware/request-id.ts:8-19`, per #12889).
- Fix: add an explicit id to `GenerationJob` and have the executor's `track` adopt it.
- Scope: streaming bypasses the DO, and pre-dedup rejections still emit normally. Rows keep `userId`, model, path, IP and hashed `cacheKey`, so user+time+model lookup still works. Billing unaffected.

### 2. Every coordinated row ships `balances: {}`

- `balance.balanceCheckResult` has one writer: `generation-access.ts:66`, inside `checkBalance`, reachable only via `generationAccess` — which the executor chain does not include.
- `track.ts:212` reads it at emit time, and the executor's row is the only row.
- `selectedMeterId`/`selectedMeterSlug` stay correct (overwritten from the committed `payerBucket` at `track.ts:436-441`); only the `balances` map degrades. No Tinybird pipe reads it; two Grafana panels do. Deduction uses live D1.
- Fix: in the executor, after `balance`, call `createBalanceCheckResult(await balance.getBalance(userId), paidOnly)` for telemetry only. Do not add full `generationAccess` — that re-runs the 402 gate admission already enforced.
- The comment at `track.ts:206-207` is now stale: it says `balanceCheckResult` is set once `balance` has run, and the executor runs `balance` with it still unset.

### Checked and cleared

`Promise.allSettled` at the settlement snapshot (all 8 `waitUntil` sites register before `.fetch()` resolves); cache-key agreement across the hop; `frontendKeyBilling` reconstructing the same `PollenRateLimiter` instance; `replaceCapturedResponse` header precedence; error responses never persisted to R2. `c.set("generationCache", …)` in the executor is **not** dead — `track.ts:215-225` reads it for the emitted `cacheKey`.

Per-request Hono construction measured at ~75–86 µs / ~40 KiB, in front of an R2 head, a DO read, a multi-second provider call and a multi-MB put — and likely cheaper than the base, which dispatched through the whole ~168-entry public app. Not worth changing; the singleton alternative reintroduces the `ExecutionContext` side channel `84a1a465f` removed.

### Notes

- `84a1a465f`'s net −28 production lines are real, but they come from deleting `generation-executor-context.ts`, the `captureError` rewrite, and the media-cache wrapper removal. The cache consolidation itself is **+13** (`generation-cache.ts` +53/−40).
- CI has never run on this PR — `ci-pull-request-checks.yml` only triggers on PRs targeting `main`/`master`, and the base is `codex/consolidate-generation-cache`.
- `seed=-1` no longer bypasses the cache, so it returns one frozen result per prompt while `APIDOCS.md:416,452,556` still documents it as random. Open, reported separately.

---

## Addendum — `1f177323a`, `95cd60ddc`, `acf0fe340`, `ac60f136d`

Verified correct:

- **`EXCLUDED_PARAMS.includes(key.toLowerCase())`** (`utils/media-cache.ts:46`) closes the admission/executor key drift: `sanitizeUrl` strips `key` case-insensitively, so a request carrying `?Key=x` previously made the two sides compute different keys — executor writes one, caller reads another, 503 after billing, deterministic on retry. Now closed. Side effect: capitalized `?NoFeed=`/`?Key=` variants change key once, so a few stale objects go unused.
- **`alarmRunning` removal** is dead-code, not a behaviour change — `startAndWait` no longer references it, and Cloudflare does not invoke `alarm()` re-entrantly for one DO.
- **Hand-written `CoordinatorStub` replaced** with the binding's own RPC type (`generation-deduplication.ts`) — removes a type that could silently drift from the DO class.
- **`X-Cache-Type: EXACT` on text hits** (`utils/text-cache.ts:168`) aligns text with media. No telemetry change — `track.ts:946` already fell back to `"EXACT"` for cache hits — and dedup still overrides with `GENERATED`/`COALESCED` for coordinated responses.
- **`app.ts` deleted**, `app` localised to `index.ts`, `vitest.config.ts` alias removed with no dangling entry. `hono-openapi` added to `dedupe` — same fix class as the earlier hono duplication.

### 3. The settlement fix is incomplete (low)

`execute-generation.ts` now does `await settlement` before throwing on `!cacheWrite` and on a `cacheWrite` rejection — a real fix for abandoned billing promises. But `const settlement = Promise.allSettled(promises)` is created *after* `captureError` and `drainResponse`.

- `drainResponse` reads the tee'd stream that feeds the R2 write, so an R2 failure can reject it.
- If it does, `executeGeneration` throws before `settlement` exists → the alarm's `settlement` stays `undefined` → its `finally { if (settlement) await settlement }` awaits nothing → the already-registered billing promise is abandoned.
- Same class as the bug fixed two lines below, two lines earlier.
- Harm direction is safe (no charge, caller gets 503), but a deduction can partially apply without its row.
- Fix: create `settlement` immediately after `.fetch()` returns.

### 4. The `seed` doc wording undersells the change

`schemas/image.ts:51` and `schemas/text.ts:19` now say `-1` "selects a random seed for a new generation; repeated identical requests may return the cached result."

- It *will*, not *may* — deterministically, and `refreshR2ObjectTtl` resets the lifecycle on access, so a regularly-hit prompt never expires.
- `-1` is now functionally identical to any fixed seed whose value happens to be unpredictable, and only the very first caller for a given prompt+params gets the randomness.
- Neither description tells callers what to do instead: supply your own varying seed.

Documenting rather than restoring the bypass is a legitimate call, but the wording should say plainly that `-1` no longer produces variation across requests.

PR #13319, opened 2026-08-13
fix: preserve input image aspect ratio when size is omitted on edits

### LINKED ISSUE #12583: [QUEST] Preserve aspect ratio for image edits

pepi21:
Fixes #12583 — references #10944.

## What this does

When `/v1/images/edits` (or any generation with an input image) is called without an explicit `size`, the gateway derives default dimensions from the source image's own proportions instead of forcing the model's square default. Implements option 2 from #10944 at the gateway level, so every edit path is covered by one mechanism.

## Why another PR for this quest

Reviewing the linked PRs, two gaps seemed worth closing:

1. **The JSON request format never triggers route-level fixes.** `CreateImageEditRequestSchema` defaults `size` to `"1024x1024"`, so on the JSON path `size` is *never* omitted by the time it reaches `handleImageEdit` — a route-level "derive when size is omitted" only fires for multipart uploads. The acceptance criteria ask for coverage of the relevant edit request formats; this PR removes the schema default for edits (generations unchanged) so both JSON and multipart behave identically, with tests for both.
2. **`dimensionsExplicit` must stay `false` for derived sizes** (as noted in the issue discussion): a size derived from the source image is still a default, not caller intent. Deriving inside the pipeline after `ImageParamsSchema` keeps the existing signal untouched, so seedream-4's custom-dimension routing and video models behave exactly as before.

## How

- `shared/schemas/openai.ts`: edits use a new `imageEditSizeField` without the `1024x1024` default. Generations keep theirs.
- `gen.pollinations.ai/src/image/params.ts`: new `deriveDimensionsFromInputImage` — probes the first input image with the existing `readImageDimensions`/`downloadUserImage` utils and scales its proportions into the model's default pixel budget (`getDefaultSideLength(model)²`), so an edit without an explicit size never costs more than today's default. Small sources keep their exact resolution (editing shouldn't upscale). Any probe failure falls back silently to the current square default — no request that works today can start failing.
- `gen.pollinations.ai/src/image/handler.ts`: applied in `generateImageOrVideoResponse` for non-video models only, before pricing input is computed. Provider-specific size constraints keep being enforced by the existing per-model snapping (gpt-image-2's 16px/3:1/4K clamps, gpt-image-1's three fixed sizes via `closestByRatio`, seedream presets, …), which already do the right thing once they receive a non-square ratio.

Explicit `size`/`width`/`height` always win (`dimensionsExplicit` short-circuits the derivation).

## Tradeoff worth flagging

For URL-sourced input images this adds one fetch at the gateway (guarded by the existing `fetchUserImage` SSRF/size limits). Data URIs — what multipart uploads become — are just decoded in memory.

## Tests

8 new cases in `test/image/params.test.ts`: portrait/landscape sources, per-model pixel budgets (flux vs seedream5), explicit-size override, unreadable-image fallback, no-image no-op, and both edit schema formats (JSON size omitted / explicit). Ran locally against current `main`:

- `vitest test/image/` — 226/226 passing
- `vitest test/index.test.ts` (includes the `/v1/images/edits` route test) — 52/52 passing
- `tsc --noEmit` clean, Biome clean on modified files

### DIFF
--- a/gen.pollinations.ai/src/image/handler.ts
+++ b/gen.pollinations.ai/src/image/handler.ts
@@ -29,7 +29,11 @@ import {
 import { getImageEnv, syncImageEnv } from "./env.ts";
 import { HttpError } from "./httpError.ts";
 import { setKleinVpcBinding } from "./models/fluxKleinModel.ts";
-import { type ImageParams, ImageParamsSchema } from "./params.ts";
+import {
+    deriveDimensionsFromInputImage,
+    type ImageParams,
+    ImageParamsSchema,
+} from "./params.ts";
 import { sanitizeString, sleep } from "./util.ts";
 import {
     CONTENT_POLICY_ERROR_CODE,
@@ -478,7 +482,13 @@ export async function generateImageOrVideoResponse(
 ): Promise<Response> {
     syncImageEnvironment(c.env);
     const originalPrompt = decodePrompt(prompt || "random_prompt");
-    const safeParams = parseImageParams(c, body);
+    let safeParams = parseImageParams(c, body);
+    // Editing an image without an explicit size should keep its proportions,
+    // not reshape it to the model's square default. Video models keep their
+    // aspectRatio/resolution semantics untouched.
+    if (!isVideoModel(safeParams.model)) {
+        safeParams = await deriveDimensionsFromInputImage(safeParams);
+    }
     c.var.track.setPricingInput({
         resolution: safeParams.resolution,
         hasImage: (safeParams.image?.length ?? 0) > 0,
--- a/gen.pollinations.ai/src/image/params.ts
+++ b/gen.pollinations.ai/src/image/params.ts
@@ -2,6 +2,10 @@ import { IMAGE_SERVICES, type ImageModelName } from "@shared/registry/image.ts";
 import type { ModelDefinition } from "@shared/registry/registry.ts";
 import { z } from "zod";
 import { getDefaultSideLength } from "./models.js";
+import {
+    downloadUserImage,
+    readImageDimensions,
+} from "./utils/imageDownload.ts";
 
 const allowedModels = Object.keys(IMAGE_SERVICES) as [
     ImageModelName,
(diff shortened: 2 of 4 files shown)


ElliotEtag:
- Equal-area scaling does not guarantee the current provider tier or cost. For a 2:1 source, a 2048² budget derives roughly 2896×1448. That moves seedream5 from 2K to 3K, seedream 4.0 from 2K to 4K, and seedream-pro onto its 4K path, even though those edit handlers already preserve the source ratio natively.

- Raw derived dimensions also bypass real backend constraints. Small sources such as 200×200 are rejected by Klein and Z-Image, which require at least 256px per side; extreme inputs can exceed their 4096px maximum. This conflicts with the acceptance criterion that outputs respect provider-supported size constraints and with the stated no-regression guarantee.

- Derivation currently runs for every non-video model, including text-only flux, zimage, dreamshaper, and Ideogram paths that discard the image. Those requests now fetch/decode an unused image and alter text-to-image output dimensions. Tests using flux therefore validate the helper but not a working edit path.

- Please gate this to edit-capable models that need gateway-derived dimensions, skip native match-input models, and normalize through each target models supported sizes. Add handler-level tests for tiny, huge, extreme-ratio, portrait, and landscape inputs.

- The new probe fetch should also have a bounded timeout; URL inputs otherwise add an unbounded serial fetch before generation.

Circuit-Overtime:
- The underlying issue is still present on current main, but this PR should not be merged or rebased as-is.
- Elliot’s blockers are confirmed: derivation runs for every non-video model, raw equal-area dimensions can change provider tiers or violate bounds, native match-input edit paths are overridden, and URL inputs add an extra unbounded fetch.
- The focused helper tests pass (13/13) and Biome is clean, but they do not exercise actual edit-capable handler/provider paths, so they do not establish the acceptance criteria.
- All four touched files have changed on main; this branch is 47 commits behind and currently conflicts. Issue #12583 also has multiple competing open implementations, and maintainers have asked contributors to stop submitting duplicates.
- Recommendation: close this duplicate rather than invest in a broad rebase. Preserve the useful schema change that distinguishes omitted edit size, then implement one maintainer-selected, provider-aware solution with handler-level constraint and timeout coverage.

On your views @ElliotEtag and @pepi21

voodoohop:
- Thank you for working on #12583 — your contribution helped shape the final solution in #13459.
- Added a claimable 5-Pollen thank-you reward to your account. You can claim it from the [Quests page](https://enter.pollinations.ai/quests).

PR #13322, opened 2026-08-13
feat(gen): add ?community query parameter to model discovery endpoints

fisventurous:
adds optional `?community` query parameter to model discovery endpoints so client apps and bots can filter out community models when they only want core official models

changes
\- new `filterEntriesByCommunityParam` helper in `proxy.ts`, applied inside `modelsListHandler` and the inline `/v1/models` handler
\- `?community=false` (or `0`) excludes community models, `?community=true` (or `1`) returns only community models
\- omitting the parameter returns all models (fully backward compatible)
\- updated OpenAPI descriptions on `/v1/models`, `/models`, `/text/models`, `/image/models` to document the parameter
\- added query parameter docs section in `models.md`
\- integration test covering `/models`, `/text/models`, `/v1/models`, `/image/models` with `true`, `false`, `1`, `0`, and omitted variants


ElliotEtag:
- `?community=tru` currently behaves like `?community=false`: every value except `true` or `1` sets `wantCommunity` to false. That silently returns the official-only catalog for unsupported values.
- Please define a shared query schema such as `z.enum(["true", "false", "1", "0"]).optional()` and attach `validator("query", ...)` to all eight model-discovery routes. This validates the contract and exposes `community` as a structured OpenAPI query parameter. Since validation can return 400, include the 400 response in the route docs.
- Please also use an image-modality community fixture and assert that `/image/models?community=true` includes it and `community=false` excludes it. The current fixture is text-only, so the image assertion does not exercise removal of a community image model.

fisventurous:
thanks for the review, all three points addressed in the latest push:

\- added `ModelListQueryParamsSchema` with `validator("query", ...)` across all 8 discovery routes so invalid queries like `?community=tru` return 400
\- exposed `community` as a structured query param in OpenAPI docs and added 400 response docs
\- added a community image fixture to test `/image/models` filtering and asserted 400 on invalid query values

Circuit-Overtime:
- Reviewed the implementation and prior discussion; all eight model-discovery routes use the shared validated query schema, and the image-community fixture covers the requested modality case.
- Fixed the failing Biome formatting and the missing `SELF` import that caused 15 type errors in the new integration test.
- Reused the schema type and completed the model docs with `/video/models`.
- Updated the fork with current `main`; no unresolved review threads.
- Local validation: Biome, `git diff --check`, and gen TypeScript typecheck pass. Refreshed CI is running before merge.

PR #13332, opened 2026-08-13
fix(polli-cli): validate numeric flags (--width/--height/--duration/--speed) before hitting the API

TruftedBug89:
## Summary

The `polli` CLI passes numeric flags straight through to the API without validation. Non-numeric or out-of-range values produce confusing raw server errors instead of a clear CLI message:

```console
$ polli gen image "a cat" --width abc
Generating image...
error: 400 Bad Request: {"success":false,"error":{"message":"Query parameter validation failed","code":"BAD_REQUEST","details":{"fieldErrors":{"width":["Invalid input: expected number, received NaN"]}}}}
```

The request is still sent (and billed) before the user learns the flag was wrong.

## Changes

- **`gen image` / `gen video`**: `--width` / `--height` are validated locally as integers in `[1, 4096]` before any network call.
- **`gen audio`**: `--duration` must be an integer in `[1, 300]`; `--speed` must be a number in `[0.25, 4]`.
- Adds a small shared helper `src/lib/validate.ts` (`parsePositiveInt` / `requirePositiveInt`) used by the commands.
- Adds the package's first unit tests (`test/validate.test.ts`, vitest) covering the helper.

## Before / after

```console
$ polli gen image "a cat" --width abc
error: --width must be an integer, got "abc"      # exit 1, no network call
```

## Checks

- `npm test` — 7/7 pass (vitest)
- `npx biome check` — clean on all changed files
- `npm run build` — tsup build succeeds

## Notes

- The API server does its own validation; this PR only makes the CLI fail fast with a friendly message instead of forwarding garbage and surfacing a raw JSON error.

### DIFF
--- a/packages/polli-cli/src/commands/gen/audio.ts
+++ b/packages/polli-cli/src/commands/gen/audio.ts
@@ -10,6 +10,7 @@ import {
 } from "../../lib/output.js";
 import { playAudio, playerMissingHint } from "../../lib/play.js";
 import { readStdin } from "../../lib/stdin.js";
+import { requirePositiveInt } from "../../lib/validate.js";
 
 export function createAudioCommand() {
     return new Command("audio")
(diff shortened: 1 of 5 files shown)


Circuit-Overtime (changes requested):
@TruftedBug89 please make the changes

- Blocking: packages/polli-cli/src/commands/gen/audio.ts validates --duration with requirePositiveInt and min: 1, so valid fractional durations are rejected.
- The audio API contract accepts 0.5–300, and eleven-sfx specifically supports 0.5–30. The CLI skill also documents that range. Running polli gen audio "short sound" --model eleven-sfx --duration 0.5 now exits locally with "must be an integer".
- Use finite-number validation for --duration with the API-wide 0.5–300 bounds, retain integer validation for width/height, and add a 0.5 duration regression test.
- Local checks otherwise pass: 7 tests, build, and Biome.

PR #13358, opened 2026-08-14
fix: use OpenAI-compatible instructions for Qwen TTS

voodoohop:
- Renames the Qwen TTS request field from `instruct` to OpenAI-compatible `instructions`.
- Sends the same plural field to DashScope as `parameters.instructions`.
- Updates JSON, multipart, simple-audio, and OpenAPI schemas.
- Updates the existing Qwen integration test to verify the exact upstream payload.

Tests:
- `npx vitest run test/index.test.ts -t Qwen --reporter=verbose`
- `npm run typecheck`
- `npx biome check gen.pollinations.ai/src/routes/audio.ts gen.pollinations.ai/src/routes/generation-handlers.ts gen.pollinations.ai/test/index.test.ts`


ElliotEtag:
- Verified upstream field name: DashScope docs only document `instructions` (plural) for `qwen3-tts-instruct-flash`, and the Python SDK maps `text`/`voice` into `input` and leftover kwargs (incl. `instructions`) into `parameters` ([api_request_factory.py#L187](https://github.com/dashscope/dashscope-sdk-python/blob/master/dashscope/api_entities/api_request_factory.py)). So `parameters: { instructions }` is the correct wire format — and the old `parameters: { instruct }` was likely ignored upstream. This PR fixes style instructions not working at all, not just a rename.
- `test/index.test.ts` replaces the only Qwen TTS test instead of adding one. Lost coverage: GET `/audio/{prompt}?model=qwen-tts`, the `qwen-tts` → `qwen3-tts-flash` model-ID mapping, and the empty-`parameters` case. Keep the old test alongside the new one.
- Old clients sending `instruct` now get it silently stripped (no 400) → un-styled audio instead of an error. Acceptable given the param never worked upstream; noting the break is silent.

Docs: https://www.alibabacloud.com/help/en/model-studio/qwen-tts

PR #13371, opened 2026-08-14
fix: distinguish invalid api keys from missing credentials in 401s

ElliotEtag:
- A request with an expired, revoked, or deleted API key returned the same 401 message as a request with no credentials at all ("Authentication required. Please provide an API key…"), which misreads as an auth-plumbing bug instead of a dead key.
- gen's auth middleware now notes when a key was extracted from the request but failed authentication, and `requireAuthorization`/`requireUser` return 401 "Invalid API key. The provided key was not recognized; it may be expired, revoked, or mistyped." with a pointer to https://enter.pollinations.ai/keys.
- The invalid-key message also takes precedence over route-specific "please provide a key" messages (realtime), since those are written for the no-credentials case.
- Adds a worker test asserting the invalid-key 401 body; existing 401 (no credentials), 402 (paidOnly balance), and 403 (permissions/banned) behavior is unchanged.

Context: a deleted testing key surfaced as "paidOnly models return 401 while openai works" — the non-paid models were only succeeding via pre-auth text-cache hits, and the generic 401 message hid that the key itself no longer existed.


voodoohop:
- Looked at how OpenRouter, Anthropic, Stripe, OpenAI, and OAuth describe authentication failures.
- For our API, missing and invalid keys need the same client action, so both now return one `401`: `A valid API key is required.`
- Removed invalid-key detection, route-specific auth messages, and the duplicate `requireAuthorization()` guard. Protected routes now use one `requireUser()` path.

Issue #13385, opened 2026-08-14
feat(polli-cli): add a simple API-complete agent surface


Circuit-Overtime:
## Problem

The CLI has good shortcuts for common generation and account workflows, but it drifts whenever the APIs add endpoints. Agents must fall back to curl and manually reproduce key lookup, bearer auth, stdin/file handling, binary output, and error behavior.

Current gaps include 3D, embeddings, advanced audio transforms/timestamps, realtime, earnings, key-specific usage, richer usage pagination, community catalog filtering, and the My Models image fields tracked in #12853. Adding a dedicated command for every route would make the CLI large and keep the drift problem.

## Scope

- Add `polli api <path>` for relative `gen.pollinations.ai` HTTP paths only.
- Reuse stored, environment, or `--key` bearer authentication; allow public calls without a key.
- Support GET/POST/PUT/PATCH/DELETE, JSON from `--data`, `@file`, or stdin, and multipart form fields including `@file` uploads.
- Print JSON/text to stdout; require `--output` for binary responses and emit machine-readable metadata under `--json`.
- Reject absolute URLs so a Pollinations key cannot be sent to an arbitrary host.
- Add the community image registration fields from #12853 to `polli my-models`.
- Document the stable agent contract and representative raw API recipes.

## Acceptance criteria

- Common typed commands remain unchanged.
- Agents can call every non-WebSocket Pollinations API without curl.
- Invalid paths, methods, JSON, form values, and binary-without-output fail clearly with non-zero status.
- Secret values never appear in normal output or errors.
- Unit tests cover request parsing, auth behavior, JSON/text output, multipart files, and binary output.

## Follow-ups

Add typed shortcuts only where repeated use justifies them (3D and embeddings first). Realtime remains SDK-native because it is a stateful WebSocket protocol.

Addresses #12853.

voodoohop:
- Split this into smaller, independently reviewable PRs.
- Prefer complete, typed `polli` commands as the CLI coverage strategy. Raw API access is already straightforward over HTTP, so a generic `polli api` passthrough needs a separate, concrete justification.
- Keep My Models/image/endpoint-agent coverage in one focused PR, then add missing command families such as embeddings, 3D, usage, and earnings separately.
- This keeps scope and tests clear while moving toward full API coverage without coupling unrelated surfaces.

PR #13386, opened 2026-08-14
docs(polli-cli): document registering an image my-model

Circuit-Overtime:
Documents how to register a community **image** model from the CLI.

- `test --modality image` before `create`, and what the probe reports back
- how `--input-modalities text,image` advertises image-edit support, surfaced in the `inputs` column
- the two billing modes: `request` (per generated image) vs `tokens` (per 1M, verified against `COMMUNITY_IMAGE_TOKEN_PRICE_FIELDS`)
- `--modality` is fixed at creation

Reduced to the docs. The `my-models.ts` flags land in #13491; the generic `polli api` pass-through is a separate product decision tracked in #13385.

Co-authored-by: @Circuit-Overtime


Circuit-Overtime:
@voodoohop I am setting this as ready to review please do let me know if you like the direction or if you need further simplification on this one

voodoohop:
Read the whole PR. Two separate outcomes.

**my-models half — superseded.** #13491 is queued to merge with the same flags plus `completionImagePrice` in the response type and a combined `image_price` column. Your two `modelBody` tests are in it, credited to you via `Co-authored-by` — they were the first tests in `packages/polli-cli`, which had `vitest run` wired and no test files.

**SKILL.md my-models block — we want this.** It's the one part nothing else covers: the `test --modality image` example, the full create example with `--image-pricing` and `--input-modalities`, and the two billing modes. Checked it against `COMMUNITY_IMAGE_TOKEN_PRICE_FIELDS` — "token price flags" is correct, `tokens` mode really does use prompt-text / prompt-image / completion-image per 1M. One ask: lift it out of the permissions paragraph, which already carries permissions plus quest claiming. A short paragraph next to the image examples reads better.

**On `polli api` — not convinced yet.** If you know the endpoint you can curl it today. The CLI's value is the dedicated commands: local flag validation before a request goes out, and `--help` discoverability the raw API can't give you. The one real argument we see is that an agent shelling out to `polli` never receives the raw key, whereas curl means handing the secret to the agent. That's interesting but not decided.

So: reduce this PR to the SKILL.md block and we'll merge it. Make the pass-through case in #13385 and we'll settle it there.

voodoohop:
Went ahead and did the reduction rather than leaving it on you — this PR is now the SKILL.md docs only, merged up to main, 11 added lines. Your content, restructured into its own `### Register an image my-model` section instead of extending the permissions paragraph, and credited via `Co-authored-by`.

Checked the billing sentence against `COMMUNITY_IMAGE_TOKEN_PRICE_FIELDS` before keeping it — "token price flags" is right, `tokens` mode really does price prompt-text / prompt-image / completion-image per 1M.

The `polli api` work isn't lost, it's just not decided. Argue it in #13385 and open a fresh PR if we land on yes.

PR #13387, opened 2026-08-14
fix: request reasoning summaries from Azure Responses API

tomdacatto:
## Summary

- body.reasoning previously set only effort, so Azure Responses did not generate reasoning-summary text even though the model produced billed reasoning tokens.
- Adds summary: auto to the reasoning request field so the existing streaming and non-streaming relay can expose the summary as reasoning_content.
- Current Microsoft Azure documentation supports auto for GPT-5 reasoning summaries. GPT-5 models do not support concise, so auto is intentional and no fallback is needed.
- Enabling summaries can add output tokens and latency for callers that explicitly set reasoning_effort.

## Validation

- npx vitest run test/text/azureResponsesClient.test.ts — 12/12 passed
- npm run typecheck — passed
- npx biome check on both changed files — passed
- GitHub checks — passed

### DIFF
--- a/gen.pollinations.ai/src/text/azureResponsesClient.ts
+++ b/gen.pollinations.ai/src/text/azureResponsesClient.ts
@@ -186,7 +186,7 @@ function buildBody(messages: ChatMessage[], options: TransformOptions): Json {
         typeof options.reasoning_effort === "string" &&
         REASONING_EFFORTS.has(options.reasoning_effort)
     ) {
-        body.reasoning = { effort: options.reasoning_effort };
+        body.reasoning = { effort: options.reasoning_effort, summary: "auto" };
     }
     if (Array.isArray(options.tools) && options.tools.length) {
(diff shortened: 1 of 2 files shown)


Circuit-Overtime (approved):
- Thank you for the focused fix and for clearly documenting the behavior and billing tradeoff.
- The request shape is correct: current Microsoft Azure Responses documentation supports reasoning.summary = "auto", and the existing streaming/non-streaming relay already maps the resulting summary into reasoning_content.
- Local validation passed: 12/12 focused tests, TypeScript typecheck, and Biome. The branch is current with main, mergeable, and all GitHub checks are green.
- Non-blocking documentation note: the PR description mentions "concise" as a fallback, but Microsoft documents that GPT-5 models do not support concise; "auto" used by this patch is the correct value. No code change is needed.
- Approved and ready to merge.

PR #13400, opened 2026-08-15
fix: add missing space between model name and description in toggle confirmation dialog

MarcosFRG:
## What
Fixes a missing space in the community endpoint toggle confirmation dialog where the model ID and the status description ran together (`modelId?It will…` → `modelId? It will…`).

## Where
`enter.pollinations.ai/frontend/src/components/community-endpoints/community-endpoint-toggle-confirmation.tsx`

## How
Added `{" "}`.

## Why
Pure rendering bug — the `?` literal was glued to the `{endpoint?.disabled` conditional expression with no whitespace, so the description text appeared immediately after the question mark with no separation.


MarcosFRG:
😅

PR #13434, opened 2026-08-15
fix: agent MCP servers reject Gradio/SSE endpoints with 502

ElliotEtag:
Reported on Discord: adding an MCP server to a prompt agent fails with `502 Bad Gateway: openai error: MCP HTTP Transport Error: POSTing to endpoint (HTTP 405): Method Not Allowed` (tested with `https://victor-websearch.hf.space/gradio_api/mcp/sse`).

Two causes, both fixed:

- The runtime only spoke Streamable HTTP (`type: "http"`), so legacy HTTP+SSE endpoints — the URL Gradio advertises for every space — got a JSON-RPC POST and answered 405. The runtime now retries with the SSE transport when the initialize POST returns 400/404/405 (the MCP back-compat flow); other errors still propagate unchanged.
- `normalizeCommunityEndpointBaseUrl` stripped trailing slashes from stored MCP URLs, and Gradio's Streamable HTTP endpoint (`/gradio_api/mcp/http/`) 307-redirects without the slash — which the runtime blocks. So even the correct URL couldn't work. MCP server URLs now keep their trailing slash.

Behavior changes: saved MCP URLs preserve a trailing slash; MCP connect falls back to SSE transport on 400/404/405.

Verified against the reported space: both `/gradio_api/mcp/sse` (via fallback) and `/gradio_api/mcp/http/` (direct) now list tools. Tests cover slash preservation, the SSE fallback tool loop for all compatibility statuses, and no-fallback on auth errors.


voodoohop:
Merged `main` into this branch and hand-resolved the conflict — it's now 0 behind, 26/26 tests pass. Core SSE fallback is right (trigger set matches the MCP back-compat flow, 401/403 correctly excluded); three things needed reconciling with what shipped after this branch was cut.

- **`loadMcpTools` kept main's per-server isolation.** This branch had one try/catch around the whole loop doing `close(); throw`, which is the pre-#13436 shape — one incompatible server would 502 every agent request again. Running main's own tests against this branch's code reproduced it (`keeps running when an MCP server fails to load`: 200 → 502). The http→sse retry is now nested *inside* the per-server try.
- **`mcpTransportFetch` follows redirects again** (#13457). Beyond re-blocking ordinary redirects, the thrown `Error` carries no `statusCode`, so `isMissingHttpTransport` never matched it and the SSE fallback never fired for the redirect case this PR set out to fix. Note the wrapper is also a required workerd shim — `@ai-sdk/mcp` sends `redirect: "error"`, which workerd rejects outright.
- **The 401 test asserted a whole-request 502**, which isolation removed. It now asserts the agent still serves, and the retry decision is covered by a unit test on `isMissingHttpTransport` (400/404/405 → retry; 401/403/429/5xx → not). I'd originally asserted "no GET on the MCP url" as the discriminator — that's wrong, `HttpMCPTransport.openInboundSse` issues a GET itself.

Trailing-slash change is untouched and correct — it wraps the shared normalizer locally rather than editing it, so community endpoints are unaffected. Its comment was updated: redirects are followed now, so the slash saves a round trip rather than being load-bearing.

Current production state of the reported agent, for context — the owner has moved to `/http` and added more servers:

```
victor-websearch      /gradio_api/mcp/http    -> 400   (SSE fallback would catch this)
google-langextract    /gradio_api/mcp/http    -> 200
burtenshaw-llm-council /gradio_api/mcp/htp    -> 404   (typo)
doc-mcp               /gradio_api/mcp/htttp   -> 404   (typo)
```

Two of the four are user typos with no feedback anywhere, which argues for URL validation alongside this.

PR #13459, opened 2026-08-16
fix: preserve aspect ratio for image edits when size is omitted

### LINKED ISSUE #12583: [QUEST] Preserve aspect ratio for image edits

Creatneworld:
Fixes #12583
References #10944

- Preserves the source image's aspect ratio when an image edit omits `size`.
- Leaves caller-provided dimensions unchanged.
- Removes the square default only from JSON edit requests; image generation still defaults to `1024x1024`.
- Runs inference only for image models that accept image input.
- Fits derived dimensions into the model's existing size budget, aligned to 16px with a 256px minimum, so large or extreme inputs are not forwarded raw.
- Reuses the existing guarded image downloader and dimension reader; no cache, provider table, or new public parameter.

Validation:

- 7 focused aspect-ratio and schema tests
- 31 related parameter, routing, and image-decoder tests
- Gen TypeScript check
- Biome on all modified files

### DIFF
--- a/gen.pollinations.ai/src/image/handler.ts
+++ b/gen.pollinations.ai/src/image/handler.ts
@@ -38,7 +38,12 @@ import {
     contentPolicyMessage,
     firstContentPolicyMessage,
 } from "./utils/contentModeration.ts";
-import { bufferToUint8Array, detectMimeType } from "./utils/imageDownload.ts";
+import {
+    bufferToUint8Array,
+    detectMimeType,
+    downloadUserImage,
+    readImageDimensions,
+} from "./utils/imageDownload.ts";
 import { setImagesBinding } from "./utils/imageTransform.ts";
 import { buildTrackingHeaders } from "./utils/trackingHeaders.ts";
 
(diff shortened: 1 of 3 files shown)


Creatneworld:
Hi @voodoohop! This implementation matches the merge direction suggested in the issue thread (g33ky00's review): the \dimensionsExplicit\ signal from \ImageParamsSchema\ is already on main, and this PR adds exactly the route-level source-size detection on top of it.

Key points:
- \
esolveEditDimensionsForImage\ runs after the schema transform, so a size derived from the source image **keeps \dimensionsExplicit: false\** — seedream-4 and friends stay on the non-custom path.
- Explicit \size\ requests and video models (I2V start-frame) are untouched.
- Provider constraints still apply downstream (gpt-image-2 clamp, gpt-image-1 closestByRatio).
- 6 new tests + 37 related tests pass; Biome clean.

Happy to adjust if you'd prefer a different placement.

voodoohop:
- Merged — thank you for the clean foundation.
- You are assigned to #12583, so the 5-Pollen quest reward will be available from the [Quests page](https://enter.pollinations.ai/quests).

PR #13471, opened 2026-08-16
feat: add account settings

ElliotEtag:
Adds self-service account settings and deletion to the Enter dashboard.

This PR builds on the merged #13564, which introduced the GitHub-scoped quest reward keys and database migrations. This PR contains the account UI, deletion flow, legal copy, and integration coverage.

- Adds an Account page with GitHub profile details, a copyable Pollinations ID, and a danger zone for account deletion.
- Adds Account and Sign out actions with package UI icons, including closing the mobile drawer after navigation.
- Uses Better Auth hard deletion for the account, sessions, credentials, balances, API keys, and user-owned resources.
- Enforces a session created within the previous ten minutes before deletion.
- Preserves only the orphaned GitHub-scoped reward idempotency records supplied by #13564, preventing a replacement account from receiving the same quest reward again. Replacement accounts see those quests in the normal claimed state without receiving a claimable reward or including the old pollen in their totals.
- Adds concise Terms and Privacy acknowledgement at sign-in and aligns the legal text with active-system deletion, temporary cache retention, billing records, and backups.
- Tests deletion cascades, reward retention, re-registration with the same GitHub identity, duplicate reward prevention, fresh-session enforcement, and the 50-Pollen purchase quest.

Addresses #13547


YoannDev90:
Can you add `Fixes #13547 ` ?

YoannDev90:
#13547

voodoohop:
Reverted the retained-quest display in 1fa1a43. Not a bug report — the code worked, it's a scope call.

- The unique reward key from #13564 already blocks re-earning. Delete, re-register, redo the quest: `onConflictDoNothing` swallows it, `recorded: 0`, no pollen. That protection needs no code here.
- What the revert drops is only the label on a card that user can no longer complete. They'll see it as `open` forever. One person, one wrong label, no money at risk — not worth logic in three files.
- Mainly: I don't want new API surface for this. `previouslyEarnedQuestIds` on `/quests/rewards` is permanent — code we can delete, a shipped response field we live with. Same objection applies to putting it on `/quests/check` instead.
- Also reverted the widened reward queries in `/account/quests` and `/quests/rewards` to the plain `userId` filter, and the frontend state that threaded the field. Restores the `Promise.all` in `/account/quests`.

−77 net lines. 35/35 quest + account-delete tests pass.

Rest of the PR is unaffected — account page, deletion flow, legal copy all stand.

PR #13479, opened 2026-08-17
feat: add typed endpoint agents

voodoohop:
- Adds `agent.kind` with prompt and endpoint config schemas behind one stable agent ID
- Routes prompt agents through Enter and endpoint agents through their saved OpenAI-compatible target
- Keeps listing visibility, pricing, RPM, fallbacks, encrypted bearer, and delegation independent from execution kind
- Migrates existing delegating endpoints into endpoint-agent records without moving credentials or spend authority
- Adds dashboard, CLI, API, runtime, and integration coverage for both kinds

Tests:
- `npx biome check --write` on all changed source files
- SQLite migration fixture passed, including target move and credential preservation
- Full local typecheck/test unavailable because the worktree dependency install ran out of disk; CI is authoritative


voodoohop:
Superseded by #13529. Same outcome, smaller shape: the discriminator lives on `community_endpoint.kind` rather than on the agent table, so the agent row stays prompt-only and no config union is needed.

PR #13488, opened 2026-08-17
fix: scope delegating endpoint cache

voodoohop:
- Scopes every delegating endpoint cache entry to the calling API key.
- Uses the managed agent ID or external endpoint ID as the stable cache namespace.
- Adds coverage for delegating and ordinary external community endpoints.

Checks:
- `npm run typecheck` in `gen.pollinations.ai`
- Targeted Vitest could not start in the current workspace: the Cloudflare worker pool exited before test discovery (0 tests); the added test is included for CI.


voodoohop:
Absorbed into #13529. `kind = "agent"` now covers delegating endpoints, so they get a per-caller cache partition; a listing with no agent id names its own scope instead of `agent:null`.

PR #13489, opened 2026-08-17
fix: validate delegating endpoint updates

voodoohop:
- Rejects priced updates for listings that delegate generation
- Rejects fallback models on delegating listings
- Allows invalid stored prices to be repaired by setting them to zero
- Adds integration coverage for repair and fallback rejection

Tests:
- `npx biome check --write enter.pollinations.ai/src/routes/community-endpoints.ts gen.pollinations.ai/src/community-endpoints.test.ts`
- `npm run typecheck` (gen)
- Targeted Vitest exits before test discovery in the local Cloudflare pool
- Enter typecheck is blocked by the worktree missing the existing `stripe` package

### DIFF
--- a/enter.pollinations.ai/src/routes/community-endpoints.ts
+++ b/enter.pollinations.ai/src/routes/community-endpoints.ts
@@ -20,6 +20,7 @@ import {
     communityModelId,
     isCommunityEndpointOwnerAllowed,
     isCommunityFallbackPricingAllowed,
+    isFreeCommunityEndpoint,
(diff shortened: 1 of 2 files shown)


voodoohop:
Mostly absorbed into #13529, which rejects non-zero prices and fallbacks on an agent listing at create and update. Not carried over: this PR also blocked *unrelated* updates while stored prices were non-zero, forcing repair of legacy delegating rows. Reopen if that matters — repair is still possible there by posting zero prices.

PR #13416, opened 2026-08-15
feat: support speech-to-text (transcription) modality for community models

YoannDev90:
## Summary
- Adds a `transcription` modality to community endpoints (My Models): register an OpenAI-compatible upstream for `/v1/audio/transcriptions`.
- Enter: `/test` probes the endpoint with a sample audio file and reports `billableUsage.promptAudioSeconds`.
- Gen: `/v1/audio/transcriptions` proxies to the community upstream and attaches `x-usage-prompt-audio-seconds` + `x-model-used` usage headers.
- Billing mirrors first-party whisper/scribe: billed against `promptAudioPrice` (shared `promptAudioPrice` column) per audio second; duration read from `usage.duration` or `usage.seconds`, missing duration bills zero.
- Frontend: transcription option in the My Models dialog, base price key row, placeholders, fallback options.
- `normalizeCommunityEndpointInputModalities` now defaults transcription endpoints to `["audio"]` instead of `["text"]`.

## Tests
- Enter service probe tests (4): duration/seconds/zero/no-text.
- Gen unit tests (5): URL derivation, supported-endpoint restriction, definition/price, billing forwarding.
- Gen integration test: register → probe → gen transcription call → usage headers → `/audio/models` + `/v1/models` listing.
- `npx vitest run` on affected files green; typecheck green both services.

## Related
Fixes #13225

PR #13519, opened 2026-08-18
refactor: one timeout for community endpoints, 300s

voodoohop:
- Adds `COMMUNITY_ENDPOINT_TIMEOUT_MS = 300_000` to `shared/community-endpoints.ts`.
- Image generation used **120s**, the enter registration probe **90s**, with no shared source. Both now read the constant.
- 300s matches what the text paths already run: `generateTextPortkey.ts:30` and `azureResponsesClient.ts:16` are both 290s.

Why keep a timeout at all: Workers impose **no wall-clock limit** on an invocation — [the runtime caps CPU time, not duration, and there is no per-subrequest timeout](https://developers.cloudflare.com/workers/platform/limits/). So without an explicit signal a hung community endpoint hangs for as long as the caller holds the connection. The 524 proxy read timeout is an *origin* limit and does not bound a Worker.

The timeout is also load-bearing for failover: it throws 502, and 502 is in `FALLBACK_ON_STATUS_CODES` (`gen.pollinations.ai/src/fallback.ts:22`), so a hung endpoint fails over to the next declared community fallback.

Follow-up: `gen.pollinations.ai/src/audio/communityEndpoint.ts` on #13416 declares its own `REQUEST_TIMEOUT_MS = 300_000` — same value, should point at this constant once that lands.

Note `gen.pollinations.ai/src/image/availableServers.ts:191` warns that a timeout signal once broke every pool request despite passing locally. That was about *adding* one where there was none; this only changes existing values. Still worth a smoke test of a community image generation after deploy.

### DIFF
--- a/enter.pollinations.ai/src/services/community-endpoint-openai.ts
+++ b/enter.pollinations.ai/src/services/community-endpoint-openai.ts
@@ -1,4 +1,5 @@
 import {
+    COMMUNITY_ENDPOINT_TIMEOUT_MS,
     type CommunityEndpointImagePricing,
     communityChatCompletionsUrl,
     communityEndpointErrorDetail,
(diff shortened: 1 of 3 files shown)


voodoohop:
Folded into #13518 — the timeout constant lands together with the download-helper unification so both callers change once.

PR #13522, opened 2026-08-18
fix(stats): median Tinybird cost so flat-rate preflight isn't 402'd

### LINKED ISSUE #12747: Pre-flight balance check uses Tinybird avg_cost_usd instead of model-defined pricing — blocks flat-rate community models with inflated estimates

iotserver24:
Fixes #12747

## Why the estimate was ~33.34 (not 0.01)

`checkBalance()` reads Tinybird `public_model_stats.avg_cost_usd`, which was a **7-day mean** of `total_price`.

Request-priced community images used to bill:

`completionImagePrice * OpenAI output_tokens`

For `vendouple/zimage` that is **0.01 × ~3334 ≈ 33.34** — the number in the 402. Billing already logs `{ completionImageTokens: 1 }` for `imagePricing !== "tokens"`, so **new** events are ~0.01. The **mean** stays poisoned until old rows age out, and token-priced models have the same outlier problem with no flat rate to fall back on.

## Change

1. **Tinybird** `public_model_stats.pipe`: `avgIf` → `medianIf` (column name stays `avg_cost_usd`). Typical charge, not a few misbilled rows.
2. **Preflight** still uses Tinybird. It only uses the defined per-request price when Tinybird is missing or **>10×** that price (clearly garbage). Token-priced models keep Tinybird.

Needs a Tinybird `deployment create` on staging then prod for the pipe to go live.

## Tests

- community flat-rate 0.01 + Tinybird 33.34 + budget 5 → allowed (Tinybird treated as outlier)
- same model + Tinybird 0.012 → trusts Tinybird
- same model + budget 0.005 → 402 quoting `~0.0100`
- `openai` still uses Tinybird

`npx vitest run src/byop-markup.test.ts` — 22 passed.


voodoohop:
@iotserver24 in this case id like to figure out why the estimate is wrong and fix it there. because it will still stay inaccurate if a model doesnt have a flat rate. like figuring out why the estimate is broken would be fixing the root cause of the problem

iotserver24:
@voodoohop agreed — the defined-price bypass was papering over a bad Tinybird number instead of explaining it.

**Why `avg_cost_usd` was ~33.34 for a 0.01 model**

`public_model_stats` does `avgIf(total_price, …)` over 7 days. For request-priced community images, billing used to do:

`completionImagePrice * OpenAI output_tokens`

So `vendouple/zimage` at **0.01 /image** × **~3334 output_tokens** ≈ **33.34**. That is the figure in the 402.

That multiply is already gone on the request path (`communityImageUsage` returns `{ completionImageTokens: 1 }` for `imagePricing !== "tokens"`). New events log ~0.01. The **mean** is still poisoned until those old rows fall out of the 7-day window, and the same mean-of-outliers problem hits token-priced models that have no flat rate to fall back on.

**What this PR does now**

1. **Tinybird:** `avgIf` → `medianIf` in `public_model_stats.pipe` (column name stays `avg_cost_usd` so callers do not break). Typical charge, not a few misbilled rows.
2. **Preflight:** still reads Tinybird. It only swaps in the defined per-request price when Tinybird is missing or **>10×** that price — i.e. when the average is clearly garbage, not as the default.

Token-priced models keep Tinybird (now a median). Needs a Tinybird `deployment create` on staging then prod for the pipe change to go live — I did not deploy it.

`npx vitest run src/byop-markup.test.ts` — 22 passed.

PR #13525, opened 2026-08-18
feat: add the fallback merge helper

voodoohop:
- Adds `shared/registry/merge-fallbacks.ts`: one generic `mergeFallbacks(base, fallbacks)` plus the `FallbackDefinition` contract
- A fallback target is either a **string** naming an existing top-level model, or an **object** stating only its difference from the parent (`id`, `provider`, `cost`, …) — one ordered list, so try-order reads in one place
- Object targets inherit the parent's public identity and are stamped `hidden: true`; `aliases`/`fallbacks`/`hidden` are `Omit`ed from the target type so they can't be hand-set
- Mechanism only — no consumer in this PR. First consumer is the image catalog in #13521, which rebases onto this
- Verified: gen + enter typechecks, biome, and a runtime check that `["flux", {id: "x-fal", …}]` yields `fallbacks: ["flux","x-fal"]` with `flux` left public and untouched

### DIFF
--- a/shared/registry/merge-fallbacks.ts
+++ b/shared/registry/merge-fallbacks.ts
@@ -1,0 +1,89 @@
+import type { ModelDefinition } from "./registry";
+
+/**
+ * A fallback route, stated as the difference from the model it stands in for.
+ *
+ * Everything left out is inherited from that model, so the public identity a
+ * caller sees — title, brand, modalities — can never drift from the model they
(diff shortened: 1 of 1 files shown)


voodoohop:
Not intended as a separate PR — the helper stays in #13521, where its consumer lives. No content lost: the file on that branch is byte-identical.

PR #13529, opened 2026-08-18
feat: one endpoint kind, two ways to run an agent

voodoohop:
Successor to #13479, on the data model that table already implies: `community_endpoint` stores the endpoints, so it stores what kind of endpoint each one is.

**Schema**
- `community_endpoint.kind` (`model` | `agent`) — what a listing *is* to callers, and the only thing deciding whether a call is authorized with an `ag_` run token that spends the caller's balance.
- `agent_id` stays pure internal routing: set → prompt agent on Enter's runtime; null → the listing's own `base_url` (an agent on the owner's server, or a plain model).
- `agent` table is unchanged and stays prompt-only — no `kind`, no config union.

**Migration 0050** backfills `kind='agent'` from `agent_id IS NOT NULL` and from `delegates_generation = 1 AND modality = 'text'`. Image rows stay `model`: the image path never minted a run token, so their flag was inert, and promoting them would break routing.

**Behaviour**
- Both agent kinds look identical from outside — same list, same `agent: true` in the catalog, same `ag_` token. Only routing differs.
- Free invariant unchanged: `mintDelegatedToken` still throws unless every price is 0.
- An agent listing never stores a bearer; the run token replaces it.
- Dashboard gets a "Runs on" toggle; `polli my-models create --kind agent`.

`delegates_generation` is superseded by `kind` but **not dropped here** — gen and enter deploy separately, so the drop is a follow-up once gen ships without reading it.

**Verification**
- gen: 1017 passed / 6 skipped
- enter: 803 passed, 5 failures confirmed pre-existing on main
- `tsc --noEmit` clean for gen and enter (src + frontend); biome clean


voodoohop:
Superseded by #13623.

That PR keeps the same premise — the table that stores the listings stores what kind each one is — but carries it through: `type` gets a third value (`hosted_agent`) and a `payload` column holds only what each kind has. The four-entry `agentRejections` array this PR left in place becomes structurally impossible there.

Also: this branch's migration is numbered 0050, which main has since taken (along with 0051).

Ported across: the kind-backfill test, extended to cover payloads.

voodoohop:
Closing as superseded — see #13623.

Issue #10733, opened 2026-05-07
Expose BYOP app details from /account/key


voodoohop:
- Add BYOP app attribution to `GET /account/key` for keys minted through BYOP.
- Include the public app key/client id, app name, and app user identity when available.
- Avoid `owner` wording; use `user` or `appUser` since this represents the app account.
- Keep non-BYOP keys simple with `byopApp: null`.
- Prefer reusing fields already loaded by auth middleware; avoid adding a new query in the route if possible.

## Acceptance criteria

- [ ] `GET /account/key` includes BYOP app attribution for keys minted through BYOP.
- [ ] Response includes the public app key/client id, app name, and app user identity when available.
- [ ] Field naming avoids `owner`; uses `user` or `appUser` for the app account.
- [ ] Non-BYOP keys keep `byopApp: null`.
- [ ] Implementation reuses fields already loaded by auth middleware where possible, avoiding a new query in the route.

### Reward

7

voodoohop:
Checked against #10422 and current origin/main.

- /account/key still returns key metadata only.
- No byopApp/client app details are exposed in the response yet.
- Leaving open.

davealan74:
PR opened at #13520 — adds `byopApp` object to `GET /account/key` with `clientKeyId`, `name`, and `appUser` for BYOP-minted keys; `byopApp: null` for non-BYOP keys. No new DB query; reuses the three fields already loaded by the auth middleware join.

PR #13520, opened 2026-08-18
feat(enter): expose BYOP app details from GET /account/key

davealan74:
Fixes #10733

## Changes

- **`enter.pollinations.ai/src/routes/account.ts`**: Replaces the flat `byopClientKeyId` field on `GET /account/key` with a `byopApp` object (`clientKeyId`, `name`, `appUser`) for BYOP-minted keys; `byopApp: null` for non-BYOP keys. Updates the OpenAPI schema accordingly. No new DB query — all three fields (`byopClientKeyId`, `byopClientName`, `byopClientUserId`) are already loaded by the auth middleware via a left-join on the BYOP client key.
- **`enter.pollinations.ai/test/account-key.test.ts`**: Updates the agent-run-token test to assert `byopApp: null` (non-BYOP parent key). Adds a dedicated test verifying `byopApp: null` for regular secret keys.

## Acceptance criteria

- Adds `byopApp` object to `GET /account/key` for BYOP-minted keys
- Includes `clientKeyId` (public app key/client id), `name` (app name), and `appUser` (app user identity)
- Field naming avoids `owner`; uses `appUser`
- Non-BYOP keys return `byopApp: null`
- Reuses fields already loaded by auth middleware — no new route query

PR #13539, opened 2026-08-18
Addressing Issue #10733 (Expose BYOP app details from /account/key #10733)- #13538

NamanSoni78:
Added BYOP app attribution to GET /account/key for keys minted through BYOP.
Includes public app key id, app name, and app user identity (user object).
Keeps non-BYOP keys simple with byopApp: null and reuses auth middleware fields.

PR created by @NamanSoni78
This PR is for Closer of issue "Ship bounty #10733: Expose BYOP app details from /account/key"

### DIFF
--- a/enter.pollinations.ai/src/routes/account.ts
+++ b/enter.pollinations.ai/src/routes/account.ts
@@ -1609,6 +1609,32 @@ export const accountRoutes = new Hono<Env>()
                                         .describe(
                                             "Publishable app key that minted this key via the BYOP authorize flow. Server-attested; clients cannot forge.",
                                         ),
+                                    byopApp: z
+                                        .object({
+                                            keyId: z
+                                                .string()
+                                                .describe(
+                                                    "Public app key / client id",
+                                                ),
+                                            name: z
+                                                .string()
+                                                .nullable()
+                                                .describe("App name"),
+                                            user: z
+                                                .object({
+                                                    id: z
+                                                        .string()
+                                                        .describe(
+                                                            "App user identity",
+                                                        ),
+                                                })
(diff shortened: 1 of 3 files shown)


voodoohop:
Thanks for this @NamanSoni78 — real work, and it arrived the same day as #13520, which we merged for #10733.

Why we went with the other one:

- It **replaces** `byopClientKeyId` with `byopApp` instead of adding alongside it. The repo leans hard on clean breaks over carrying two fields with the same data (see the YAGNI section in `AGENTS.md`).
- Flat `appUser: string` over `user: { id: string }` — the extra nesting didn't earn its keep for a single id.

Worth saying though: keeping the old field was the **safer** read, and you were more right than the PR we merged. `media.pollinations.ai` consumes `byopClientKeyId` from `GET /account/key` to stamp `appKeyId` on upload rows, and #13520 missed that consumer — we had to add the media update before merging, or every BYOP upload would have silently recorded a null app id, with no test catching it. Your instinct not to break the existing field was correct; the resolution was to update the consumer rather than keep both.

We've awarded you **3.5 Pollen** for the contribution — it'll show up under your quest rewards at https://enter.pollinations.ai.

Thanks for jumping on this one, and sorry for the duplicated effort — that's on our side for leaving #10733 unassigned. Hope to see you on the next one.

NamanSoni78:
@voodoohop  Thank you! I’m really happy to have contributed, and I’m glad to work with the Pollinations AI community. I’ll definitely keep participating and look forward to tackling more issues and contributing wherever I can. 🚀

Thanks again for the opportunity and for the feedback!

PR #13593, opened 2026-08-19
fix(econ): stop booking community sale price as provider cost

### LINKED ISSUE #12895: Community model cost is recorded as the full sale price, so margin reads negative
We record `total_cost` = the full sale price for community models, but we never pay a provider for them — the owner runs the endpoint on their own credentials. Our only outflow is the 75% owner reward.

The economics reader then subtracts that payout from revenue *and* books it as cost, so it is counted twice.

**Worked example — 1.0 Pollen community generation**

| Column | Recorded | True |
|---|---|---|
| `total_cost` | 1.0 | 0 |
| `total_price` | 1.0 | 1.0 |
| `community_model_reward_amount` | 0.75 | 0.75 |

- `shared/community-endpoints.ts:425` sets `cost: communityPriceDefinition(...)` with `priceMultiplier: 1` (`:426`), so cost and price are numerically identical by construction. No `provider === "community"` carve-out exists anywhere in the cost path.
- `"community"` is in `POLLEN_PRICED_VENDORS` (`apps/operation/economics/web/src/lib/insights.ts:822`), forcing `calib = 1`, so `trueCostPaidUsd` = the full 1.0.
- `retainedPaidUsd = soldPaid − eco` = 1.0 − 0.75 = **0.25** (correct).
- `marginUsd = retainedPaidUsd − providerUsage` = 0.25 − 1.0 = **−0.75** (`insights.ts:1067-1070`).

**What's wrong on the dashboards**

- Every community model shows margin −0.75 per 1.0 sold instead of +0.25, and counts toward `underwaterCount` (`insights.ts:1187`).
- `trueMultiplier` reads 0.25× when there is no provider cost to multiply against.
- Cash Margin / Cash Margin % tiles (`views/EconTab.tsx:68-82`) and the `EconTable` margin column (`components/EconTable.tsx:591-594`) inherit it.
- `apps/operation/economics/ingest/connectors/community.md:48` books `cost_paid + cost_quests` as the community provider-cost witness and mirrors it into `op_cloud`, so the full gross price lands in Total spend / Cash P&L as an expense — for a payout made in Pollen scrip, not cash.
- Lower stakes: `weekly_usage_stats.pipe:21` (`sum(total_cost)`), the KPI gross-margin tooltip (`apps/operation/kpi/src/components/KPITrendTable.jsx:168`), and the "Provider cost for usage served through paid Pollen" label in `DataQualityTab.tsx:416`, which is simply false for community.

**BYOP already solves this**

BYOP puts the third-party payout *outside* the base price (`total_price = dev_price + devCredit`, `shared/billing/track-helpers.ts:163-165`), so `total_cost` stays a genuine provider cost and `op_pollen_populate.pipe:21` recovers the payout as `total_price − dev_price`. Community reuses the revenue half of that convention (`model_paid` nets out of revenue) but not the cost half.

**Fix options**

- Zero `cost` for community definitions and let `community_model_reward_amount` be the sole representation of the payout. Blocked as a one-liner: `getPriceDefinitionForModel` derives price *from* cost, so `cost: 0` would also zero the price. Needs a price definition independent of `cost × multiplier`.
- Or carve community out in `op_pollen_populate.pipe` and treat `cost_paid` as 0 for `provider = 'community'`, leaving the registry alone.

Introduced by `5b9034745 feat: add community model rewards`; the economics reader was never updated to match. Not blocking #12819 — that PR's per-request split is correct on its own terms; this is the underlying cost model.

iotserver24:
Fixes #12895

Community models record `total_cost` = the sale price, but Pollinations never pays an upstream provider. The only outflow is the 75% owner reward (`model_paid`). The economics reader then subtracted that payout from revenue **and** treated the sale price as provider cost, so a 1.0 Pollen generation showed margin **−0.75** instead of **+0.25**.

## Changes

1. **`op_pollen_populate.pipe`** — `cost_paid` / `cost_quests` skip `provider = community`. Price, requests, and `model_paid` are unchanged.
2. **Economics reader** — ignore community pollen cost when computing `trueCostPaidUsd` / margin, so already-materialized rows with the old meter do not stay underwater.
3. **Community connector** — do not mirror `cost_paid` into `op_cloud`. The payout is eco, not a cloud invoice.

Worked example in `insights.test.ts`: `price_paid=1`, `cost_paid=1`, `model_paid=0.75` → retained **0.25**, margin **+0.25**, not underwater.

`npx vitest run src/lib/insights.test.ts` — 55 passed.

The pipe change needs a Tinybird `deployment create` (staging then prod) before `op_pollen` itself stores zeros. The reader fix is live as soon as this ships.


voodoohop:
@iotserver24 I added a small follow-up:

- Normalizes historical community provider cost in op_pollen_api, so every consumer gets the same data.
- Uses the owner reward as community COGS in the weekly gross-margin KPI.
- Removes the duplicate economics-reader workaround.

Reason: the sale price is not an upstream provider bill. The owner payout is recorded separately, leaving 0.25 Pollen retained from a 1.00 Pollen sale. All 55 focused tests pass.

PR #13600, opened 2026-08-20
fix: replace z.custom() with proper zod schema for OpenAPI generation

rekty:
- Replaces the non-serializable `z.custom<ModelMessage>()` field with a JSON Schema-compatible permissive schema.
- Restores `/api/docs/open-api/generate-schema` without changing runtime validation behavior.
- Unblocks generated API reference updates.

Fixes #13558

PR #13609, opened 2026-08-20
fix: keep generated API docs complete

ElliotEtag:
- Fix the custom managed-agent request schema that crashes Enter OpenAPI generation
- Fail the merged docs schema when the required account schema is unavailable
- Require account, community-model, and agent paths before generating APIDOCS.md
- Include OpenAPI tag prose in generated Markdown and clarify App Key usage
- Add regression coverage for agent routes and partial-schema failures

Tests:
- enter.pollinations.ai: docs redirect + prompt agent runtime (22 passed)
- gen.pollinations.ai: docs routes (7 passed)


voodoohop:
- #13600 is merged, and this branch is now updated from `main`.
- I resolved the schema overlap by retaining `z.unknown()` with the localized `ModelMessage[]` cast.
- The remaining type-safety, fail-closed schema handling, and regression coverage make sense as the follow-up.

PR #13615, opened 2026-08-20
fix: explicit max_tokens defaults for OpenRouter reasoning models

### LINKED ISSUE #13594: Qwen reasoning models default to Chutes' 1024 max_tokens, causing blank output and billed tokens
**Problem**
Qwen reasoning models (`qwen3.8-27b`, `qwen3.8-max`) are returning blank answers while still consuming/billing completion tokens. This happens when the caller does not explicitly set `max_tokens`.

**Root cause**
Pollinations only sets `max_tokens` defaults for Claude/Bedrock models in `modelConfigs.ts`. For the Qwen OpenRouter routes, no default is injected, so the upstream provider's default applies. Chutes AI — the provider serving these models — defaults `max_tokens` to **1024**.

Reasoning models consume that entire 1024-token budget on the internal thinking/reasoning trace before emitting any visible response, so the user gets a blank answer but is still billed for ~1024 completion tokens.

Evidence from Chutes docs for `Qwen3.8-27B-FP8`:
- `max_tokens` default: **1024**
- OpenRouter model page lists 65,536 as the *maximum supported*, but the *default* in effect is Chutes' 1024.

**Affected models**
- `qwen3.8-27b` (has `createReasoningEffortTransform("toggle")` but reasoning defaults to ON when not specified)
- `qwen3.8-max` (no reasoning transform, reasoning always on)

**Expected behavior**
Reasoning models should not silently inherit a 1024-token provider default that guarantees blank output. Pollinations should set an explicit `max_tokens` default for Qwen reasoning models (similar to what already exists for Claude), ensuring enough headroom for the reasoning trace plus the final answer.

**Suggested fixes**
1. Set an explicit `max_tokens` default for Qwen reasoning models in `modelConfigs.ts` (e.g. 8192–16384, or align with the provider's reasoning max).
2. Consider defaulting `reasoning_effort` to `"none"` for Qwen when the user does not request reasoning, so non-reasoning calls behave predictably.
3. Investigate whether usage should be refunded/changed when the model produces zero visible output due to budget exhaustion.

**Impact**
Users calling `qwen3.8-27b` or `qwen3.8-max` without explicit `max_tokens` get empty responses and are charged for completion tokens that produce no usable output.

---
**Author:** `Yerminator (@the_yerminator)` (UID: `1123534156626939945`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1539690547982106664/1539693941828354060)

---

### Reward

3

gggff123:
- Sets a 64,000-token default ceiling for seven OpenRouter reasoning models.
- Prevents low provider defaults from consuming the response budget on reasoning before visible output.
- Keeps caller-provided `max_tokens` authoritative.
- Mirrors the existing explicit-default pattern used by Claude/Bedrock models.

Closes #13594

### DIFF
--- a/gen.pollinations.ai/src/text/configs/modelConfigs.ts
+++ b/gen.pollinations.ai/src/text/configs/modelConfigs.ts
@@ -156,7 +156,7 @@ export const portkeyConfig: PortkeyConfigMap = {
         createOpenRouterModelConfig({
             model: "z-ai/glm-5.3",
             defaultOptions: {
-                max_tokens: 16384,
+                max_tokens: 64000,
                 provider: {
                     only: ["z-ai/fp8"],
                     allow_fallbacks: false,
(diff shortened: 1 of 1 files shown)


NamanSoni78:
Nice Work 😁

gggff123:
@voodoohop and @ElliotEtag  If you are free Please review this .

NamanSoni78 (approved):

PR #13584, opened 2026-08-19
Addressing issue #13558 "/api/docs/open-api/generate-schema 500s — z.custom cannot be represented in JSON Schema #13558 Open"

NamanSoni78:
Addressing issue #13558 
Fixes HTTP 500 on GET /api/docs/open-api/generate-schema by replacing z.custom() with a JSON Schema compatible Zod schema shape in PromptAgentRequestSchema and hiding the internal agent runtime route from OpenAPI docs via describeRoute({ hide: true }).
@voodoohop  please check it
Pull Requested by @NamanSoni78

PR #13617, opened 2026-08-20
feat: support per-user RPM for catalog models

voodoohop:
- Reuses one `perUserRpm` model field for catalog and community models.
- Enforces limits across every modality and fallback attempt.
- Exposes configured limits through model-list endpoints.
- Leaves models uncapped when the field is unset.
- Adds focused limiter and fallback coverage.


NamanSoni78:
Sir @voodoohop if you are free then please review pr #13584

Issue #10, opened 2021-07-16
[Bug] After pressing cancel button, inputs at pollinations.ai disappeared


gokaykucuk:
After running and receiving the video output of one run, I pressed the "stop" button to run another input. At this point the website became unresponsive. After refreshing the page, all the "input"s disappeared from pollinations.ai and the system got into an unusable state.
```
https://pollinations.ai/ipfs/QmaXCvUxwvQiPrLmBeTEiqaA9N2Y319pqVwo92aet1TQr2
```
After some inspection, it seems like the output folder has "status='running'" while input has "formAction=cancel".

gokaykucuk:
Healed by itself, closing

PR #13620, opened 2026-08-20
fix: prevent zero billing when OpenRouter omits usage

ElliotEtag:
## Problem

Some successful paid OpenRouter requests omit usage or return an all-zero object. Pollinations then records zero provider cost and charges zero Pollen.

## Production evidence

Tinybird production snapshot on 2026-08-21:

- Seven-day all-zero usage: `gemini-fast` 3,734/645,788 successes (142 fully missing), `gemini` 388/3,406 (166 missing), and `gemini-flash-lite-3.5` 43/4,250.
- The rolling-hour symptom was limited to those three pinned Gemini routes. Non-Gemini routes had only isolated seven-day events: Mistral 12, Llama Scout 2, Mercury 2.

## What this PR does

- Recovers exact usage from OpenRouter generation metadata before settlement.
- Retries at 0s, 5s, and 15s for metadata indexing.
- Marks unrecovered requests `usage_missing` instead of silently treating them as billed at zero.
- Excludes genuinely free community routes and does not backfill history.

## Relationship to the source fix

[Gateway #10](https://github.com/pollinations/gateway/pull/10) is the source fix: the generic adapter strips OpenRouter provider pins, while the native adapter lacked parameter parity. Once released, Pollinations can enforce Vertex pins centrally.

This PR is optional billing-boundary defense-in-depth, not the routing fix. It protects paid requests if an upstream still returns unusable usage after Gateway #10.

## Checks

- Biome, TypeScript, and all 39 tracking-observability tests pass.

### DIFF
--- a/gen.pollinations.ai/src/middleware/track.ts
+++ b/gen.pollinations.ai/src/middleware/track.ts
@@ -128,6 +128,11 @@ type ResponseTrackingData = {
     errorTracking?: ErrorData;
 };
 
+const OPENROUTER_GENERATION_ENDPOINT =
+    "https://openrouter.ai/api/v1/generation";
+const OPENROUTER_USAGE_RECOVERY_DELAYS_MS = [0, 5_000, 10_000] as const;
+const OPENROUTER_GENERATION_LOOKUP_TIMEOUT_MS = 5_000;
+
 export type TrackVariables = {
     track: {
         modelRequested: string | null;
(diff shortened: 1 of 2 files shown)


ElliotEtag:
Closing as superseded by the source fix in pollinations/gateway#10. After that gateway change merges, the remaining Pollinations change should only upgrade the gateway and switch the shared OpenRouter factory to the native adapter.

PR #13413, opened 2026-08-15
chore: consolidate prompt agent migration

voodoohop:
- Regenerates the final prompt-agent schema as a single Drizzle migration
- Folds MCP header storage into migration 0048
- Removes intermediate migrations and snapshots 0049–0050
- Keeps the 0048 filename so the existing staging migration history remains compatible

Validation: fresh local D1 full migration chain, zero Drizzle drift, 22 focused agent tests, and no pending staging migrations.

Issue #12346, opened 2026-07-12
feat: per-tool pricing for community prompt agents


voodoohop:
Follow-up to #12336 (tool pricing was removed there in `14eb75f1c`; restore-and-improve).

- Owner-declared per-tool fees (`toolPrices`), billed from `usage.tool_call_counts` (already reported unbilled by the agent template)
- Restore: `tool_prices` column, `shared/registry/community-billing.ts`, `fees` in model-info//models, route/CLI (`--tool-price`)/dashboard fee inputs — see the removal diff for the exact surface
- Design decision: bill only successful tool calls; failed calls still feed errors back to the model, optionally reported as unbilled attempts
- Fees only apply to public agents (private = owner pays own costs, no fees)

Depends on #12336.

PR #13623, opened 2026-08-20
feat: one typed listing table — type + payload envelope

voodoohop:
## Summary

- Unifies proxy models, prompt agents, and endpoint agents in one typed `community_endpoint` table.
- Gives every listing a common `base_url` and `upstream_model`; prompt-agent storage uses a fail-closed `.invalid` runtime placeholder that resolves from the environment at the API/runtime boundary.
- Uses strict Zod schemas for external writes and kind-specific payloads: only proxies store credentials, pricing, RPM, or fallbacks; only prompt agents store prompt/MCP config.
- Removes the separate agent table and migrates linked and standalone prompt agents without exposing credentials or prompts.
- Updates Enter, Gen, dashboard, and `polli` CLI together as a clean API break.

## Related work

- Product direction: #11373; broader runtime research: #10628.
- Community listing foundations: #11567 (table/routing) and #12333 (visibility), both merged.
- Early agent-kind and prompt-agent paths: #12253 and #12266, consolidated by #12278.
- Prompt-agent foundation: #12336 (merged); migration consolidation #13413 was folded into it.
- Endpoint-agent schema lineage: #13479 → #13529 → this PR.
- Separate follow-up, intentionally not implemented here: prompt-agent tool pricing #12346.

## Verification

- CLI tests: 4/4.
- Gen routing, run-token, and fallback tests: 27/27; Gen typecheck clean.
- SQLite migration validation covers linked and standalone prompt agents, endpoint agents, proxies, constraints, and indexes.


voodoohop:
Pushed `bda3f85d87` — one real test break, found running enter's suite locally.

`test/integration/account-delete.test.ts` removed the `agent as agentTable` import and the agent fixture insert, but kept the assertion:

```
ReferenceError: agentTable is not defined
  96| expect(await db.select().from(agentTable)).toHaveLength(0);
```

0053 does `DROP TABLE agent`, so the assertion has nothing left to check — removed it. File passes (2/2).

Worth flagging why CI was green with this on the branch: enter's suite doesn't run in CI, and `enter/tsconfig.json` excludes `test/**` from typecheck, so neither the reference error nor the missing import was visible. Full local run is now **6 failed / 812 passed (818)**, and all 6 are the known pre-existing set that fails on clean main too (`account-key` ×1, `docs-redirect` ×1, `integration/account-keys` ×3 — plus this one, now fixed).

Separately verified fixed since the earlier review pass: the client strips `modality` before the PATCH (`community-endpoints.tsx:177`), the gen catalog's per-arm guard is `if (!payload) return []` rather than inverted, and `d1-tinybird-sync.ts` now reads prices out of the payload while deliberately not exporting the payload itself.

PR #13637, opened 2026-08-20
ci: one production deploy workflow, one migration step

voodoohop:
- Three workflows (`deploy-enter-cloudflare`, `deploy-gen-cloudflare`, `deploy-media-cloudflare`) shared the `pollinations-cloudflare-production` concurrency group, so they serialized in arbitrary order. Each independently ran `wrangler d1 migrations apply` against the same `production-pollinations-enter-db`.
- Measured skew: 2-6 min between the schema moving and the last worker's code going live, ~3 min of it pure concurrency-release latency between separate workflow runs.
- Now one workflow: a single `migrate` job gates `deploy-enter`, `deploy-gen`, `deploy-media` running in parallel. Order is deterministic and the gap drops to roughly the deploy step itself.
- Per-service path filtering preserved via `dorny/paths-filter@v3`, pinned with `base: ${{ github.ref }}`. That is required on long-lived branches: without it the action diffs `merge-base(main, production)..production`, which is empty after a promotion, so every filter reports false and the deploy jobs skip while `migrate` still applies the schema.
- `workflow_dispatch` gains a `service` selector: `all` (default), `enter`, `gen`, `media`.
- `docs-regenerate-api-reference.yml` now watches `Deploy / Cloudflare production`, with a `gate` job that regenerates only when the run's `deploy-gen` job succeeded — the run-level conclusion does not track gen.
- Runbooks updated: `operations/deployment/README.md`, `AGENTS.md`, `enter.pollinations.ai/AGENTS.md`, `packages/mcp/README.md`.
- Trade-off, stated plainly: the migration no longer runs after gen's `Test` step. Previously it did only when gen happened to be dequeued first (1 of 3 orderings) — when enter or media went first the migration was already applied untested. So this makes behaviour uniform, not strictly safer. Gating `migrate` on gen's tests is a possible follow-up.

CI cannot exercise the production push topology; the `base` behaviour is reasoned from the action's source.


voodoohop:
Fixed in 57ecea5be1.

**P1 — confirmed, and worse than a skip.** Verified against `dorny/paths-filter@v3` source: with `base` unset it resolves `base = defaultBranch` (`main`), `head = github.ref` (`production`), so `isBaseSameAsHead` is false and it falls through to `getChangesSinceMergeBase(main, production)`. On a `main`→`production` promotion the merge base *is* main's tip and the merged tree matches main's, so the diff is empty. Every filter reports false — but `migrate` has no `if`, so it still runs. The failure mode is migrations applied with zero workers deployed, which is strictly worse than the skew this PR set out to fix. Now `base: ${{ github.ref }}`, which routes through the `beforeSha` branch and matches the old `on.push.paths` semantics.

**P2 docs** — added a `gate` job that reads the triggering run's jobs API and regenerates only when `deploy-gen` concluded `success`. Fixes both directions: enter/media-only runs no longer regenerate, and a gen success survives an enter failure.

**P2 runbooks** — updated `operations/deployment/README.md`, `AGENTS.md`, `enter.pollinations.ai/AGENTS.md`, `packages/mcp/README.md`. No stale references remain.

Note: `.github/docs/DEPLOYMENT.md` was renamed to `operations/deployment/README.md` on `main`; patched the live path.

Agreed CI can't exercise the production push topology — the `base` fix is reasoned from the action's source, not from a green run.

voodoohop:
One more, found on re-audit: `migrate` had no `if`, so it ran whenever the workflow fired regardless of what the filters resolved to.

That is what made the missing `base` a production hazard instead of a no-op — filters false, every deploy job skips, schema moves anyway. Gated it on at least one service output being true (c1e3125af9), so "migrations applied, no worker updated" is now unreachable by construction rather than merely unlikely.

Also verified, no change needed: `paths-filter` copes with the default shallow checkout — `ensureRefAvailable` does `git fetch --depth=1 origin <sha>` for both endpoints and diffs the trees, so `fetch-depth: 0` is not required.

Remaining known gap, unchanged and deliberate: `migrate` is not gated on gen's `Test` step. Pre-PR, gen's tests gated the migration only when gen was dequeued first; enter-first and media-first both migrated untested. Making it uniform was the tradeoff. Adding a `test-gen` job ahead of `migrate` would strictly improve on all three, at ~2 min added latency for every production deploy — happy to add if wanted.

PR #13662, opened 2026-08-21
fix: render activity timestamps in user local timezone Fixes #13651

### LINKED ISSUE #13651: Activity dashboard shows incorrect timestamps due to timezone handling
@voodoohop and @ElliotEtag 
### Description

The timestamps displayed in the Activity page (`https://enter.pollinations.ai/activity`) appear to be incorrect.

In both:
- **Events over time**
- **Last events**

the displayed event times do not match my actual local time.

I am located in India (IST, UTC+5:30), and I suspect this might be related to timezone handling or incorrect conversion from UTC to the user's local timezone.

### Expected behavior

The activity timestamps should either:
- Automatically display in the user's local timezone, or
- Clearly indicate the timezone being used (for example UTC).

### Actual behavior

The timestamps appear shifted and do not match the actual event time.

### Steps to reproduce

1. Open `https://enter.pollinations.ai/activity`
2. Check the timestamps under:
   - Events over time
   - Last events
3. Compare the displayed time with the actual event time in the local timezone.

### Environment

- Location: India (IST / UTC+5:30)
- Browser: Chrome
- Device: Desktop

### Possible cause

This may be related to:
- Missing timezone conversion on the frontend
- Dates being rendered directly in UTC
- Incorrect use of browser locale/timezone formatting

Thanks for looking into this!

karanray06:
Fixes #13651

Root cause: The frontend correctly received UTC timestamps but display formatters hardcoded `timeZone: 'UTC'` in their `Intl.DateTimeFormat` calls, overriding the browser's automatic timezone conversion. This fix removes the override to default to the process' local timezone, updates the UI label, and adds unit tests.

### DIFF
--- a/enter.pollinations.ai/frontend/src/components/activity/activity-helpers.tsx
+++ b/enter.pollinations.ai/frontend/src/components/activity/activity-helpers.tsx
@@ -84,34 +84,4 @@ export function downloadFile(url: string): void {
     anchor.remove();
 }
 
-export function formatActivityChartDate(
-    date: Date,
-    isHourly: boolean,
-): { label: string; fullDate: string } {
-    return {
-        label: isHourly
-            ? date.toLocaleTimeString("en-US", {
-                  timeZone: "UTC",
-                  hour: "2-digit",
-                  minute: "2-digit",
-                  hour12: false,
-              })
-            : date.toLocaleDateString("en-US", {
-                  timeZone: "UTC",
-                  month: "short",
-                  day: "numeric",
-              }),
-        fullDate: date.toLocaleDateString("en-US", {
-            timeZone: "UTC",
-            weekday: "short",
-            year: "numeric",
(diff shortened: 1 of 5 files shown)


voodoohop:
- Thanks @karanray06 for working through the Activity timestamp paths.
- The picker, API windows, and chart buckets are intentionally UTC; changing only the displayed labels makes a selected UTC day span two local calendar dates.
- #13804 keeps those semantics consistent and makes the UTC behavior explicit with one wording change.

karanray06:
> - Thanks @karanray06 for working through the Activity timestamp paths.
> - The picker, API windows, and chart buckets are intentionally UTC; changing only the displayed labels makes a selected UTC day span two local calendar dates.
> - #13804 keeps those semantics consistent and makes the UTC behavior explicit with one wording change.

Okay @voodoohop

PR #13665, opened 2026-08-21
fix: #13660 - Activity timezone display

### LINKED ISSUE #13660: Activity dashboard shows timestamps in UTC instead of the user's local timezone
## Problem

The Activity dashboard at `https://enter.pollinations.ai/activity` appears to display timestamps in UTC instead of the user's local timezone.

The issue is visible in:
- **Events over time**
- **Last events**

For users in timezones such as **IST (UTC+5:30)**, the displayed event times do not match their local time, which can make the activity history confusing.

## Expected Behavior

Timestamps should either:

- Be converted and displayed in the user's local/browser timezone, or
- Clearly indicate that the displayed timestamps are in UTC.

## How to Reproduce

1. Open `https://enter.pollinations.ai/activity`.
2. Use a browser configured to a timezone other than UTC, such as `Asia/Kolkata`.
3. Trigger an activity/event and note its actual timestamp.
4. Compare the timestamp shown in **Last events** and **Events over time** with the corresponding API timestamp.
5. Verify whether the displayed time matches the user's local timezone.

## Suggested Investigation

- Check the timestamp returned by the Activity API.
- Trace how the timestamp is parsed and formatted in the frontend.
- Check whether the UI explicitly formats timestamps as UTC.
- Compare the behavior of **Last events** and **Events over time**.

## Acceptance Criteria

- Activity timestamps are displayed consistently in the user's local timezone, **or**
- The UI clearly indicates when timestamps are displayed in UTC.
- Both **Last events** and **Events over time** follow the same timezone behavior.
- Existing timestamp formatting elsewhere in the application is not affected.

pegalink:
Fixes #13660

- Activity dashboard timestamps (Last events, Events over time) and Model Monitor "data as of" time were
  hardcoded to render in UTC via `timeZone: "UTC"`, so users outside
  UTC (e.g. IST) saw times that didn't match their local clock.
- `formatTimestamp` (last-events-panel.tsx) and `formatActivityChartDate`
  (activity-helpers.tsx) now format without a `timeZone` override, so
  `toLocaleString`/`toLocaleDateString` render in the viewer's browser
  timezone instead.
- Both Last events and Events over time now behave consistently.
- Updated the graph caption from "Times are shown in UTC" to "Times
  are shown in your local timezone."
Note: the usage/earnings graph buckets are still aggregated
server-side on fixed UTC day/hour windows — only the displayed label
is now localized, not the bucket boundary itself. For timezones
behind UTC this can occasionally shift which local date a bar's label
shows relative to the UTC window it sums. Fully localized bucketing
would need backend aggregation changes, out of scope here.

### DIFF
--- a/enter.pollinations.ai/frontend/src/components/activity/activity-helpers.tsx
+++ b/enter.pollinations.ai/frontend/src/components/activity/activity-helpers.tsx
@@ -88,21 +88,21 @@ export function formatActivityChartDate(
     date: Date,
     isHourly: boolean,
 ): { label: string; fullDate: string } {
+    // No `timeZone` override on any of these — each bucket boundary is a
+    // fixed instant in time, and formatting it without a timezone renders
+    // it in the browser's local timezone.
     return {
         label: isHourly
             ? date.toLocaleTimeString("en-US", {
-                  timeZone: "UTC",
                   hour: "2-digit",
                   minute: "2-digit",
                   hour12: false,
               })
             : date.toLocaleDateString("en-US", {
-                  timeZone: "UTC",
                   month: "short",
                   day: "numeric",
               }),
         fullDate: date.toLocaleDateString("en-US", {
-            timeZone: "UTC",
             weekday: "short",
             year: "numeric",
             month: "short",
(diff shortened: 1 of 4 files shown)


voodoohop:
- Thanks @pegalink for keeping the implementation small.
- The period picker and bucket boundaries remain UTC, so browser-local labels alone would not represent a local calendar period consistently; the model-monitor edit is also outside this issue.
- #13804 resolves the confusion with one explicit Activity-page wording change.

PR #13692, opened 2026-08-22
fix: invalidate stale KPI cache entries

voodoohop:
- Bumps the KPI Tinybird cache key version
- Invalidates existing six-hour edge entries
- Makes the new 30-minute TTL effective immediately


voodoohop:
Closing in favor of clearing the existing KPI cache operationally; the 30-minute TTL change is already deployed.

PR #13706, opened 2026-08-22
feat: add pollenType field to API keys

sharktide:
## Summary
- Adds `pollenType` column to apikey table to restrict which pollen bucket an API key can use
- Adds `questPollenOnly` column to prevent quest models from using paid pollen
- Updates `canCoverEstimatedCharge()` and `atomicDeductUserBalance()` with `keyPollenType` parameter
- Passes pollen type through track-helpers and generation-access
- Updates API routes for create/update with pollenType and questPollenOnly
- Adds `PollenTypeInput` component and updates frontend key dialogs
- Fixes isPaidOnly to take precedence over keyPollenType
- Fixes `>` to `>=` for restricted bucket checks
- Includes Drizzle migration

## Precedence rules
1. paidOnly: true → always pack
2. questPollenOnly: true + quest model → always tier
3. Per-model override (PR #13707) → use that
4. Key-level pollenType → use that
5. Default: existing logic


Circuit-Overtime:
- Rebased onto current `main` and pushed to `sharktide/feat/key-pollen-type`.
- Preserved the newer overdrawn-wallet and strict paid-only behavior from `main`.
- Fixed the Drizzle snapshot parent and paid-only balance reporting.
- Added focused preflight and atomic-deduction coverage for Quest/Paid key restrictions.
- Verified: 15 billing tests, Gen typecheck, Biome, and all GitHub checks pass.

PR #13707, opened 2026-08-22
feat: add per-model pollen type overrides in permissions.models

sharktide:
Stacked on: #13706

## Summary
- Adds `ModelPermissionEntry` type (string | {id, pollenType}) for per-model pollen type overrides
- Adds `resolveModelPollenType()` and `extractAllowedModelIds()` helpers
- Updates `normalizePermissions()` and `canonicalizeModelPermissionEntries()` for mixed arrays (prefers object overrides over strings when same model appears as both)
- Updates `filterPermissionsToVisibleModels()` to handle union types
- Updates `generation-access.ts` and `track.ts` to use per-model overrides
- Updates Zod schemas to accept union types for allowedModels
- **Fixes**: All model-access consumers (auth.ts, model.ts, realtime.ts, proxy.ts) now use `extractAllowedModelIds()` to handle object-form permissions correctly


Circuit-Overtime:
- Restacked directly on the updated #13706 branch and pushed to `sharktide/feat/model-pollen-override`.
- Added the missing object-form `allowedModels` API schema and preserved overrides through dashboard edits.
- Updated `/account/key` to return object-form model permissions and added regression coverage.
- Verified: 24 focused permission/billing tests, Gen typecheck, and Biome pass.

PR #13708, opened 2026-08-22
feat: add questPollenOnly key-level toggle for quest models

sharktide:
Stacked on: #13707 (which is stacked on #13706)

## Summary
- Adds `questPollenOnly` boolean toggle on API keys
- When enabled, quest models (paidOnly: false) can only use quest pollen and never fall back to paid
- Updates `generation-access.ts` with full pollen type resolution precedence
- Updates `track.ts` deduction to respect the flag
- Updates API routes schemas and handlers
- Adds `QuestPollenOnlyToggle` component to frontend
- Updates key dialogs and types

## Precedence rules (final state)
1. `paidOnly: true` → always pack
2. `questPollenOnly: true` + quest model → always tier
3. Per-model override in permissions.models → use that
4. Key-level pollenType → use that
5. Default: existing logic


Circuit-Overtime:
- Restacked directly on the updated #13707 branch and pushed to `sharktide/feat/quest-models-no-paid`.
- Added the missing create/update/list, D1 persistence, auth loading, and `/account/key` plumbing.
- Centralized precedence so paid-only > quest-only lock > per-model override > key-level restriction is identical across preflight, realtime, and deduction.
- Verified: 24 focused permission/billing tests, Gen typecheck, and Biome pass.
- Product distinction: this flag differs from `pollenType: "quest"` only by overriding a per-model `paid` setting. Keep #13708 only if that hard-lock behavior is intentional. CI is refreshing; not merged.

PR #13715, opened 2026-08-22
feat: add Browser MCP worker

voodoohop:
- Adds a private stateless MCP for rendered Markdown, screenshots, and PDFs
- Uses Cloudflare Browser Run Quick Actions and Media upload RPC
- Reports measured Browser Run time for Pollen settlement at the Gen proxy
- Extracts the trusted usage-header writer shared with FFmpeg
- Adds focused transport, URL, upload, usage, and failure tests


voodoohop:
Consolidated into #13716 so the MCP Worker and registry integration land in one PR.

PR #13717, opened 2026-08-22
feat: add web search MCP worker

voodoohop:
- Adds one stateless `searchWeb` tool backed by the Pollinations `openai-search` model
- Forwards the caller credential to the normal Gen model path, so existing Pollen billing and telemetry remain authoritative
- Returns concise Markdown answers with deduplicated citation links
- Keeps the Worker private for service-binding access through the MCP registry
- Adds transport, authorization forwarding, citation, failure, batch, and dry-run coverage

### DIFF
--- a/apps/web-search-mcp/deploy.json
+++ b/apps/web-search-mcp/deploy.json
@@ -1,0 +1,7 @@
+{
+  "target": "worker",
+  "credentials": "myceli",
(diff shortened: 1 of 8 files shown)


voodoohop:
Consolidated into #13718 so the MCP Worker and registry integration land in one PR.

PR #13719, opened 2026-08-22
feat: add Transcription MCP worker

voodoohop:
- Adds one stateless `transcribeAudio` tool for public HTTPS audio
- Reuses the shared public-URL validator and validates redirects
- Forwards the caller credential to the normal Gen transcription API for Pollen billing and telemetry
- Supports model, language, and prompt inputs with a 50 MB source limit
- Adds transport, multipart forwarding, URL safety, upstream failure, and dry-run coverage


voodoohop:
Consolidated into #13720 so the MCP Worker and registry integration land in one PR.

PR #13721, opened 2026-08-22
feat: add Vision MCP worker

voodoohop:
- Adds one stateless `analyzeImage` tool for visual questions and OCR
- Passes public HTTPS image URLs to Pollinations multimodal text models
- Forwards the caller credential to the normal Gen text path for Pollen billing and telemetry
- Reuses the shared public-URL validator and keeps the Worker private
- Adds transport, multimodal payload, URL safety, upstream failure, and dry-run coverage


voodoohop:
Consolidated into #13722 so the MCP Worker and registry integration land in one PR.

PR #13723, opened 2026-08-22
feat: add Python MCP worker

voodoohop:
- Adds one stateless `runPython` tool in a fresh network-disabled Cloudflare Container
- Limits execution to 10 seconds and returns bounded stdout/stderr
- Reports measured active Container time for Pollen settlement through Gen
- Destroys the per-call Container on success and failure
- Adds transport, execution, failure billing, batch, bundle, and Docker image configuration coverage


voodoohop:
Consolidated into #13724 so the MCP Worker and registry integration land in one PR.

PR #13725, opened 2026-08-22
feat: add private media upload entrypoint

voodoohop:
- Adds a private Worker entrypoint for streamed service-owned media uploads
- Stores unlisted outputs in the existing Media R2 bucket with immutable metadata
- Validates declared sizes before writing and caps uploads at the configured limit
- Adds a staging Media Worker environment for service-binding integration
- Covers streamed storage and invalid-size rejection


voodoohop:
Superseded by #13672, which now targets main and includes the media upload entrypoint, FFmpeg Worker, registry, billing, and Gen proxy.

PR #13731, opened 2026-08-22
feat: add Transcription MCP

voodoohop:
- Adds a private Transcription MCP Worker\n- Registers Transcription for Gen routing and agent selection\n- Preserves caller billing through the downstream audio API\n- Adds bindings and proxy/Worker tests


voodoohop:
Consolidated into #13744 with the Web Search and Vision MCP servers.

PR #13732, opened 2026-08-22
feat: add Vision MCP

voodoohop:
- Adds a private Vision MCP Worker\n- Registers Vision for Gen routing and agent selection\n- Preserves caller billing through the downstream multimodal model\n- Adds bindings and proxy/Worker tests


voodoohop:
Consolidated into #13744 with the Web Search and Transcription MCP servers.

PR #13733, opened 2026-08-22
feat: add Python MCP

voodoohop:
- Adds a private network-disabled Python Container MCP Worker\n- Registers Python for Gen routing and agent selection\n- Bills measured container runtime in Pollen\n- Adds bindings and proxy/Worker tests


voodoohop:
Consolidated into #13745 with the Browser and Storage MCP servers.

PR #13534, opened 2026-08-18
Addresses #13515 (Phase 1 — bearer auth, core S3 proxy). SigV4 and  Pollen billing metering are follow-up work, not included here.

karanray06:
Implemented Phase 1 of the S3-compatible API for media storage. Added documentation for the new S3 API in the media-storage docs. Updated the integration and S3 test suites to enforce strict secret key authorization for write and delete operations, ensuring publishable keys remain read-only for uploads. Resolved TypeScript compilation errors regarding D1 database bindings and implicit any types across the test suites and S3 implementation.

Checks: Verified that the S3 API documentation accurately reflects Phase 1 capabilities. Confirmed all TypeScript compilation errors are resolved via tsc. Passed all 29 integration tests and verified the entire vitest suite runs green. Ensured proper key identity mapping for isolation and delete authorization tests.

Addresses #13515 (Phase 1 — bearer auth, core S3 proxy). SigV4 and Pollen billing metering are follow-up work, not included here.

PR #13735, opened 2026-08-23
feat: add stateless Storage MCP

voodoohop:
- Adds a stateless Storage MCP with list, read, write, and delete tools for small text files\n- Routes it through the authenticated Gen MCP registry and forwards the caller credential for downstream billing\n- Keeps storage caller-scoped so agent run tokens inherit their parent user namespace\n- Adds Worker, Gen proxy, typecheck, and dry-run coverage\n\nDepends on #13534 and #13672. #13534 now passes its focused S3 suite 33/33 after the root-listing fix; it still needs environment-correct agent-token verification and downstream Pollen metering before Storage can ship.

### DIFF
--- a/apps/storage-mcp/deploy.json
+++ b/apps/storage-mcp/deploy.json
@@ -1,0 +1,7 @@
+{
+  "target": "worker",
+  "credentials": "myceli",
+  "install": "npm ci",
+  "build": "npm test",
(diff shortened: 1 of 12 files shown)


voodoohop:
Consolidated into #13745 with the Browser and Python MCP servers.

PR #13741, opened 2026-08-23
feat: connect Discord accounts

voodoohop:
- Adds Discord account linking while keeping GitHub as the only sign-in provider
- Enforces one connected Discord identity per Pollinations account
- Shows the linked Discord profile and Pollinations server membership with a session-only cached lookup
- Adds a 1-Pollen quest for connected users confirmed in the Pollinations Discord
- Adds an explicitly gated production sync for the three Discord secrets


enter.pollinations.ai/src/services/discord.ts
```diff
@@ -0,0 +1,64 @@
...
+    const response = await fetch(
+        `https://discord.com/api/v10/guilds/${POLLINATIONS_DISCORD_GUILD_ID}/members/${account.accountId}`,
+        {
+            headers: {
+                Authorization: `Bot ${config.botToken}`,
+            },
+        },
+    );
+
+    if (response.status === 404) return { member: false, joinedAt: null };
```
ElliotEtag: **[P2] Distinguish Discord error code `10007` (unknown member) from other 404 responses such as `10004` (unknown guild). Treating every 404 as `member:false` converts deployment or bot-access failures into incorrect quest eligibility instead of surfacing an upstream error. Parse the JSON error body and return false only for `10007`; propagate other codes as failures.

enter.pollinations.ai/src/routes/account.ts
```diff
@@ -793,6 +806,50 @@ export const accountRoutes = new Hono<Env>()
...
+                        },
+                    },
+                },
+                401: { description: "Unauthorized" },
+                404: { description: "Discord account not connected" },
+                502: { description: "Discord membership check failed" },
+            },
+        }),
+        async (c) => {
+            await c.var.auth.requireAuthorization();
```
ElliotEtag, line 842: **[P2] Keep this bot-backed lookup session-scoped or require an explicit account permission. The router accepts API keys, so this route also admits delegated agent-run tokens whose account scope is intentionally stripped because third parties may hold them. Each call exposes membership and `joinedAt` and makes a live shared-bot request. The API-key verifier used by this Enter route has rate limiting disabled; the separate publishable-key generation limiter does not protect it. Reject API keys or require `account:profile`, then cache/throttle the lookup and honor Discord 429/`Retry-After`.

enter.pollinations.ai/src/auth.ts
```diff
@@ -110,6 +121,16 @@ export function createAuth(env: Cloudflare.Env, ctx?: ExecutionContext) {
                 enabled: true,
             },
         },
+        account: {
+            encryptOAuthTokens: true,
```
ElliotEtag: **[P3] Account for existing plaintext GitHub OAuth tokens before enabling encryption globally. Better Auth will attempt to decrypt those legacy values, causing token and account-info operations to fail until reauthentication. Migrate, deliberately invalidate, or document that transition and cover it with a legacy-row test. Existing sessions and the newly linked Discord flow are unaffected.

enter.pollinations.ai/frontend/src/routes/_dashboard.account.tsx
```diff
@@ -32,13 +42,82 @@ export const Route = createFileRoute("/_dashboard/account")({
...
+            }
+            const account = data?.find(
+                (account) => account.providerId === "discord",
+            );
+            if (!account) {
+                setDiscordConnection(null);
+                return;
+            }
+
+            const [{ data: info }, membershipResponse] = await Promise.all([
```
ElliotEtag: **[P3] Recover the Discord controls after errors and rejected requests. The `listAccounts` error path and a rejection here can leave `discordConnection` undefined, while a rejected link/unlink call can leave `connectionPending` true, keeping the button disabled until reload. Clear loading state in `finally` and load membership independently so its failure cannot prevent disconnecting an already-linked account.

ElliotEtag (dismissed):
Requesting changes for the Discord membership authorization and shared-bot quota exposure. Additional correctness, OAuth-token transition, and frontend recovery findings are inline.

enter.pollinations.ai/src/services/discord.ts
voodoohop: Fixed. Discord response bodies are parsed and only error code 10007 returns `member: false`; other 404s, including 10004, surface as upstream failures. The integration test covers the unknown-guild case.

enter.pollinations.ai/src/routes/account.ts
voodoohop: Fixed. The membership route now rejects API keys and is session-only, caches successful membership results for 60 seconds, and returns Discord 429 responses with `Retry-After`. Per the product decision, no additional local cooldown is added.

enter.pollinations.ai/src/auth.ts
voodoohop: Fixed by removing global `encryptOAuthTokens`. This feature no longer changes storage behavior for existing plaintext GitHub OAuth tokens, so no migration or invalidation is required.

enter.pollinations.ai/frontend/src/routes/_dashboard.account.tsx
voodoohop: Fixed. Account loading and link/unlink pending state now recover in `finally`, and membership loading is independent, so a failed lookup cannot prevent disconnecting an already linked account.

PR #13743, opened 2026-08-23
feat: connect GitHub repositories through MCP

voodoohop:
- Updates the existing GitHub App connection to current main; Better Auth handles linked user tokens, encryption and refresh without a new table.
- Adds GitHub to the MCP list and agent picker, with Connect/Manage/Disconnect under account settings.
- Proxies a small read-only tool set to the official hosted GitHub MCP through the existing Enter binding; no new Worker or container.
- Uses the caller’s connection, keeps GitHub credentials server-side, rejects public keys and their derived agent tokens, and blocks upstream redirects.
- Verified: 35 Enter tests, 9 gateway tests, 2 code-agent SDK tests, Enter/Gen typechecks, frontend build and scoped Biome.

Setup: configure the GitHub App client ID, secret and slug per environment. Request read access for repository contents, issues and pull requests; Administration write is not needed. No secrets were changed or deployments run. An authenticated hosted-MCP smoke test remains pending app configuration.

Related: #13790.


voodoohop:
Architecture note for the later consumer: keep this PR as the GitHub connection/authorization layer, but avoid coupling its repository operations specifically to memory. #13790 tracks a backend-neutral Git storage MCP where memory is one convention; a connected GitHub repository and a Cloudflare Artifacts repository could implement the same list/read/write/history contract. Agents with sandboxes can use a short-lived remote directly, while prompt agents need the MCP tool surface.

PR #13745, opened 2026-08-23
feat: add browser, Python, sandbox, and storage MCPs

voodoohop:
- Adds Browser, Python, Sandbox, and Storage MCP servers
- Gives each caller or managed-agent run a short-lived Cloudflare development sandbox with shell and file tools
- Uses private media uploads for Browser screenshots and PDFs
- Registers and routes the servers through the Gen MCP proxy
- Bills trusted usage receipts through the shared Pollen flow

FFmpeg was extracted and merged separately in #13915.


voodoohop:
Related follow-up: #13790 proposes a Git-native storage MCP backed by Cloudflare Artifacts.

This should complement rather than expand the current small-file Storage MCP in this PR:

- Keep this PR focused on caller-scoped object storage.
- Prototype Artifacts separately for versioned files, commits, history, and forks.
- Put memory conventions above the generic Git/file tools instead of making a memory-only backend.
- Preserve a backend-neutral tool contract so connected GitHub repositories can later use the same surface.

The main unknown is write complexity: current Artifacts binding/REST docs expose reads, history, repo lifecycle, forks, and tokens, but not a direct write-and-commit method; the prototype needs to test Git Smart HTTP from a Worker.

https://developers.cloudflare.com/artifacts/

voodoohop:
- Hosted MCP registry and Gen proxy merged in #13672.
- FFmpeg was extracted and merged in #13915.
- Git-backed storage continues in the focused #14102.
- Python duplicates the general Sandbox capability.
- Browser and Sandbox need separate current-main implementations; retaining the 72-file stale bundle would mix obsolete registry, billing, deployment, and SDK code.

PR #13747, opened 2026-08-23
refactor: extract community reward selection

voodoohop:
- Extracts the community reward recipient decision from `track.ts` into a pure `selectCommunityModelReward` helper.
- Preserves public endpoint, cross-owner public fallback, and same-owner private fallback behavior.
- Keeps `servedPrice` as the reward base, including the existing undefined-price fallback behavior.
- Adds a focused regression assertion and leaves deduction, telemetry, and side-effect ordering unchanged.

Checks:
- Biome (3 files)
- Gen TypeScript typecheck
- `tracking-observability.test.ts` (40/40)


voodoohop:
- Superseded by #13756, now reopened as the grouped behavior-preserving Gen fallback-attribution refactor.
- #13751 remains separate because it is an observable correctness change.
- Closing this smaller PR without merging.

PR #13763, opened 2026-08-23
refactor: clarify community endpoint policy

voodoohop:
- Derives create/update modality, pricing, visibility, RPM, and advertised policy in one pure decision path
- Leaves fallback lookup, credential encryption, publish authorization, and SQL visibly ordered in the handlers
- Preserves private zero-pricing, paid-only resets, image-pricing transitions, and payload rewrite boundaries
- Adds integration coverage for create/update policy transitions

Checks:
- `npm run typecheck` (`enter.pollinations.ai`)
- focused community endpoint suite: 29 passed
- `npx biome check --write` on changed files
- `git diff --check`

### DIFF
--- a/enter.pollinations.ai/src/routes/community-endpoints.ts
+++ b/enter.pollinations.ai/src/routes/community-endpoints.ts
@@ -482,6 +482,102 @@ const UpdateEndpointSchema = z
     })
     .strict();
 
+type ProxyUpdateInput = z.infer<typeof ProxyUpdateSchema>;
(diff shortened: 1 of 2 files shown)


voodoohop:
Superseded by #13767, reconstructed on current main and expanded into a cohesive modularization: schemas, pure proxy policy, fallback rules, and response presentation are now separated from HTTP orchestration. This avoids carrying the unrelated encrypted staging-file commit through the old branch.

PR #13773, opened 2026-08-23
fix: prevent caching of authenticated account responses

### LINKED ISSUE #13771: [QUEST] Prevent caching of authenticated account responses
### How this quest works

### Goal

### Outcome

### Existing code

### Reward

4

davealan74:
Fixes #13771

## Changes

- Adds a single `.use()` after-handler middleware on `accountRoutes` that sets `Cache-Control: private, no-store, max-age=0` and `Pragma: no-cache` for every route under `/account`, including nested `/my-models` routes — applied once, not per-handler.
- Removes the existing per-handler `c.header("Cache-Control", ...)` from the `/keys` list endpoint, which was the only route that set it before.
- Adds `test/integration/account-cache-headers.test.ts` with two focused tests: one for a top-level account route (`/account/profile`) and one for a nested route (`/account/my-models`), both asserting the correct `Cache-Control` and `Pragma` values.

### DIFF
--- a/enter.pollinations.ai/src/routes/account.ts
+++ b/enter.pollinations.ai/src/routes/account.ts
@@ -796,6 +796,11 @@ const usageResponseSchema = z.object({
  */
 export const accountRoutes = new Hono<Env>()
     .use(auth({ allowApiKey: true, allowSessionCookie: true }))
+    .use(async (c, next) => {
+        await next();
+        c.header("Cache-Control", "private, no-store, max-age=0");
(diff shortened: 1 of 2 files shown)


voodoohop:
- Thanks @davealan74 for the thoughtful implementation. We selected and merged #13852 for #13771, so I’m closing this alternative to keep one path.

voodoohop:

voodoohop:

PR #13796, opened 2026-08-23
feat: create endpoint agents in dashboard

voodoohop:
- Adds an endpoint-agent create mode to the existing community endpoint dialog
- Adds a separate Agents-section action that calls the endpoint-agent registration API
- Keeps bearer credentials, modality, pricing, fallbacks, and endpoint probes out of the agent form

Related: #13781

Tests:
- `npx vite build --config frontend/vite.config.ts`
- `npx vitest run test/integration/community-endpoint-policy.test.ts` (6/6)
- Biome and `git diff --check`


Circuit-Overtime:
The create mode correctly reuses the existing listing form while submitting only endpoint-agent fields to the strict `/endpoint-agents` route; bearer credentials, modality, pricing, fallbacks, and endpoint probes stay outside this flow. The resulting listings are also partitioned into the Agents section and retain the existing edit/delete path.

PR #13797, opened 2026-08-23
fix: fall back to X-Request-Id header in PollinationsError.requestId

### LINKED ISSUE #13794: [QUEST] Preserve request IDs in JavaScript SDK errors
### How this quest works

### Goal

Preserve the server-generated request ID on JavaScript SDK errors so users can include it in support reports.

### Outcome

- `PollinationsError.requestId` uses the error body's request ID when present.
- It falls back to the response's `X-Request-Id` header when the body omits it.
- The behavior is covered by focused tests and one short usage example.

Minimal approaches are preferred. Do not change server responses or add success metadata, interceptors, logging, tracing, or compatibility layers.

### Existing code

- The server already emits `X-Request-Id` on every response.
- [`pollinationsErrorFromResponse`](https://github.com/pollinations/pollinations/blob/main/packages/sdk/src/error-response.ts) currently reads only the JSON body.

### Parent issue

#11203

### Reward

3

davealan74:
Fixes #13794

When the error body omits a `requestId` field, `pollinationsErrorFromResponse` now reads the `X-Request-Id` response header as a fallback, so `PollinationsError.requestId` is always populated when the server provides an ID.

- `packages/sdk/src/error-response.ts` — adds the header fallback (one line change)
- `packages/sdk/src/client.test.ts` — 3 focused tests: body ID wins, header fallback, neither→undefined

### DIFF
--- a/packages/sdk/src/client.test.ts
+++ b/packages/sdk/src/client.test.ts
@@ -334,3 +334,53 @@ describe("Pollinations.imageEdit — response resolution (characterization)", ()
         });
     });
 });
+
+describe("Pollinations request ID preservation", () => {
+    it("uses requestId from error body when present", async () => {
+        const client = newClient();
+        fetchMock.mockResolvedValue(
+            makeResponse(
+                { error: { message: "bad", code: "BAD_REQUEST", requestId: "body-req-id" } },
+                { ok: false, status: 400, headers: { "X-Request-Id": "header-req-id" } },
+            ),
+        );
+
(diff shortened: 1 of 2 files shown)


voodoohop:
- Thanks @davealan74 for the thoughtful implementation. We selected and merged #13819 for #13794, so I’m closing this alternative to keep one path.

voodoohop:

voodoohop:

PR #13850, opened 2026-08-24
fix(sdk): preserve request ID from X-Request-Id header fallback

rekty:
Implements Quest #13794. PollinationsError.requestId now uses the error body's requestId when present, and falls back to the X-Request-Id response header when the body omits it.

- Add header fallback in pollinationsErrorFromResponse()
- 5 focused tests: body ID, header fallback, no ID, flat body, non-JSON


voodoohop:
- Thanks @rekty for the thoughtful implementation. We selected and merged #13819 for #13794, so I’m closing this alternative to keep one path.

voodoohop:

voodoohop:

PR #13852, opened 2026-08-24
fix(account): send no-store cache headers on all account routes

### LINKED ISSUE #13771: [QUEST] Prevent caching of authenticated account responses

TRDEFI:
﻿Closes #13771

Adds one middleware on `accountRoutes` (right after the auth middleware) that sets `Cache-Control: private, no-store, max-age=0` and `Pragma: no-cache` on every response, including the nested `/agents` and `/my-models` routes. Headers are set after `await next()` so nothing a handler does can skip the policy.

Also drops the per-handler `Cache-Control` line in the `/keys` list handler, since the router-level middleware now covers it.

A focused test covers one top-level response (`/account/profile`) and one nested response (`/account/agents`) for both headers.


Circuit-Overtime (approved):
- The router-level middleware covers both top-level and nested account routes and enforces the required headers after handlers complete.
- The redundant `/keys` header is removed, and the focused profile/agents coverage matches #13771.
- Biome and all current GitHub checks pass.
- Approved; update the branch from `main`, then this is ready to merge.

PR #13853, opened 2026-08-24
feat(sdk): add accountEarnings() for GET /account/earnings

### LINKED ISSUE #13769: [QUEST] Expose developer earnings in the JavaScript SDK
### How this quest works

Multiple contributors may submit PRs for this quest. We will select the best completed approach, preferring correctness, clarity, and minimal code. The selected contributor will be assigned before merge and receive the reward.

### Goal

Expose developer earnings through the JavaScript SDK as a thin typed client for the existing account API.

### Outcome

- The SDK provides `accountEarnings()` for `GET /account/earnings`.
- Existing time-window query options can be passed through.
- The API's `daily` and `perEntity` response is typed and returned without unnecessary transformation.
- The method is exported, briefly documented, and covered by a focused test.

Minimal approaches are preferred. Do not add backend, Tinybird, CLI, React, caching, aggregation, or compatibility layers.

### Existing API and references

- [`GET /account/earnings`](https://github.com/pollinations/pollinations/blob/main/enter.pollinations.ai/src/routes/account.ts)
- [Existing SDK account methods](https://github.com/pollinations/pollinations/blob/main/packages/sdk/src/client.ts)
- [Enter dashboard consumer](https://github.com/pollinations/pollinations/blob/main/enter.pollinations.ai/frontend/src/components/activity/use-earnings-data.ts)

### Parent issue

#12771

### Reward

5

TRDEFI:
﻿Closes #13769

Thin typed client for the existing developer earnings endpoint:

- `accountEarnings(options)` on the `Pollinations` client, passing through `days`, `granularity`, `period` and `format` as query params
- `EarningsOptions`, `DeveloperEarningsRow`, `EarningsSource` and `EarningsResponse` types matching the API's `daily` / `perEntity` shape, returned without transformation
- New types exported from the package root

Two focused tests in `client.test.ts`: one asserting the `?days=30` query string and untouched payload, one for the bare endpoint when no options are passed. `vitest run` and `tsc --noEmit` pass in `packages/sdk`.


Circuit-Overtime (changes requested):
- Thank you for keeping this wrapper focused and for matching the existing daily/perEntity response shape. The focused tests, typecheck, Biome, and GitHub checks pass locally.
- One runtime blocker remains: EarningsOptions exposes format: "csv", but accountEarnings() always calls getJson<EarningsResponse>(). With format: "csv", the endpoint returns text/csv, so the SDK attempts JSON parsing and rejects despite promising EarningsResponse. The minimal fix is to omit format from this JSON-returning method; a separate CSV API can be added only if needed.
- Please also serialize days when it is defined rather than truthy. accountEarnings({ days: 0 }) currently drops the value and silently requests the server default instead of preserving server validation. Add a focused regression for this case.
- After those two small changes and an update from main, this should be ready to merge.

voodoohop:
- We selected another implementation for this quest, so I’m closing this PR.

PR #13859, opened 2026-08-24
feat(shared): strip EXIF/IPTC/XMP/GPS from user-supplied input images

fisventurous:
strips EXIF, IPTC, XMP, GPS, and comment metadata from all user-supplied input images before they hit any upstream model provider or storage, while preserving embedded ICC colour profiles. this was missing entirely, meaning user photos with embedded GPS coordinates, camera serial numbers, and device timestamps were being forwarded raw to all providers, including community BYOM endpoints.

output images using `writeExifMetadata` are unaffected, as this only targets incoming data. the implementation uses pure binary marker slicing on `uint8array`, so there are zero new dependencies and it runs sub-millisecond on workers.

\- added `shared/image-strip.ts` to handle JPEG, PNG, and WebP metadata chunks. preserves JPEG APP2 `ICC_PROFILE` colour profiles and handles `0xFF` marker fill bytes while stripping non-ICC metadata.
\- fetched images are now sanitised in `gen.pollinations.ai/src/userImage.ts`, acting as a single choke-point covering vision, edits, image-to-image, video, 3d, embeddings, and community endpoints. 
\- uploaded images are also sanitised in `media.pollinations.ai/src/index.ts` right before the R2 write.
\- added `gen.pollinations.ai/test/image/imageStrip.test.ts` covering byte-level metadata removal, ICC retention, fill marker handling, clean-image passthrough, and malformed input resilience.


fisventurous:
@voodoohop

Circuit-Overtime:
Thanks for addressing a real privacy gap here. Stripping sensitive metadata before storage or provider delivery is worthwhile, but a few correctness issues need resolving before merge:

- fetchUserImage is not a universal provider boundary. Several adapters, including the OpenRouter video paths, forward safeParams.image URLs directly, so those inputs still reach upstream providers unchanged.
- The JPEG parser does not handle legal 0xFF marker fill bytes. If metadata was already removed, encountering one can stop parsing and return a truncated image.
- APP2 commonly contains ICC color profiles; removing every APP2 segment can change image color and goes beyond stripping EXIF, XMP, IPTC, GPS, and comments.
- Add focused fixtures for JPEG, PNG, and WebP metadata removal, clean-image byte preservation, malformed or truncated chunks, JPEG fill markers, ICC preservation, and the actual provider and R2 boundaries.

I would keep the shared zero-dependency approach, but place or verify sanitization at every real outbound/storage boundary before relying on the privacy guarantee.

fisventurous:
thanks for the thorough review, I have pushed an update addressing the parser and test coverage items:

the 0xFF fill byte bug is resolved. the JPEG parser now consumes consecutive 0xFF fill/padding bytes before reading marker codes so images with legal marker padding parse cleanly without truncation.

APP2 handling is now selective. segments beginning with the ICC_PROFILE\0 header are preserved to protect colour rendering profiles, while non-ICC APP2 segments (such as FPXR) and APP1, APP13, and COM segments continue to be stripped.

added `imageStrip.test.ts` with focused byte-level test fixtures covering JPEG metadata removal, 0xFF fill byte handling, ICC profile retention vs non-ICC APP2 stripping, PNG tEXt/eXIf chunk stripping, WebP EXIF removal with RIFF file size updates, without reallocation, and truncated input resilience.

regarding direct URL passthrough on adapters like openrouter video: `fetchUserImage` sanitises everything fetched or stored by the gateway (vision, image edits, image-to-image, 3D, embeddings, media uploads, and community models). converting remote URL-forwarding video adapters to inline data URIs would significantly inflate payload sizes and hit upstream request body constraints, so those remain as URL passthrough for now. thoughts?

Circuit-Overtime:
Thanks @fisventurous — the JPEG fill-byte handling, ICC preservation, and focused format fixtures address two important parts of the earlier review. Two correctness gaps remain before merge:

- `fetchUserImage` still is not the universal outbound boundary. [OpenRouter video](https://github.com/pollinations/pollinations/blob/94937785aa2f06d06c51eb1b64455fe8748e8772/gen.pollinations.ai/src/image/models/openRouterVideoModel.ts#L91-L95), [Wan 3](https://github.com/pollinations/pollinations/blob/94937785aa2f06d06c51eb1b64455fe8748e8772/gen.pollinations.ai/src/image/models/wan3FalVideoModel.ts#L63-L85), and [Rodin](https://github.com/pollinations/pollinations/blob/94937785aa2f06d06c51eb1b64455fe8748e8772/gen.pollinations.ai/src/model3d/models/rodinModel.ts#L16-L29) still forward the original URL. A controlled temporary URL containing the sanitized bytes is an alternative to large data URIs; either cover every outbound path or narrow the privacy guarantee accordingly.
- The malformed-input tests only cover short headers. In each parser, removable metadata can set `stripped = true`, a later truncated segment can `break`, and the function then returns only the accumulated prefix. Add JPEG/PNG/WebP cases with valid metadata followed by a malformed chunk and fail closed after complete structural validation; returning a truncated image corrupts input, while forwarding the original would leak the metadata.

The media-upload R2 boundary itself is now covered by the implementation.

Circuit-Overtime (approved):
Thanks for the update — I pushed commit `a2795ccf4` to the contributor branch to close the remaining implementation gaps.

- Routes caller-controlled image inputs through the shared sanitized byte path before OpenRouter image/video, xAI, fal video, Rodin, and Trellis requests.
- Sanitizes data-URI and URL image parts for text providers instead of only providers requiring base64.
- Makes JPEG/PNG/WebP parsing fail closed when metadata is followed by malformed or truncated structure, avoiding partial/corrupt output.
- Returns a clear 400 for malformed metadata-bearing uploads and adds boundary coverage for provider payloads and R2 storage.

The earlier blocking concerns are addressed; the implementation is ready for merge LGTM!

Circuit-Overtime:
Thanks again for the careful privacy work here. I restacked the branch on current `main` and resolved the text-image transform conflict by preserving sanitization for every caller-controlled URL/data URI; retaining `main`’s provider skip would have reopened the metadata leak. The parser, provider-boundary, and R2-storage coverage remains intact, with no further technical changes requested.

fisventurous:
!askpolli review the most recent change

fisventurous:
!askpolli review now?

fisventurous:
!askpolli removed the data URI fast path from `toDataUri` so all inputs route through `downloadUserImage` and have EXIF/GPS stripped before upstream submission. added test coverage for metadata-bearing data URIs across all provider models as mentioned. review again.

fisventurous:
!askpolli review now

fisventurous:
!askpolli check now

fisventurous:
!askpolli now tracks completion on the `0x3b` trailer byte and routes incomplete stripped streams through `malformedAfterMetadata` to fail-closed with `InvalidImageStructureError`, matching the behaviour in the other format parsers. added regression test coverage for stripped GIFs missing the trailer. review.

fisventurous:
!askpolli review again

voodoohop:
Thanks for the thorough work. Scope we want:

- **No extra work per request:** no downloading or base64-converting URL images just to strip them. If the caller already published a URL, stripping it protects little
- **Strip only where we already hold the bytes**, mainly media.pollinations.ai uploads (those become public URLs)
- **Use a maintained npm library** instead of hand-written parsers for 4 formats; less code to own, fewer false 400s on odd but valid images
- **Open question:** stripping by default vs opt-in. Opt-in is simpler to reason about; default-on at upload is fine if it stays cheap

A much smaller PR along these lines would be easy to review.

PR #13882, opened 2026-08-25
Never cache authenticated account responses

### LINKED ISSUE #13771: [QUEST] Prevent caching of authenticated account responses

dseeker:
Closes #13771.

Authenticated account responses under /account (profile, balance, usage, keys, and the nested /agents and /my-models routers) had no explicit cache headers, so a shared proxy or browser cache could serve one user's data to another. This adds a setPrivateNoStoreHeaders middleware, applied once via .use() on accountRoutes so it covers every route mounted under it including the nested routers, that sets Cache-Control: private, no-store, max-age=0 and Pragma: no-cache on every response. The one existing per-handler Cache-Control header in the /keys listing endpoint is now redundant and has been removed.

Added test/integration/account-caching-headers.test.ts with two focused tests, one for a top-level route (/account/profile) and one for a nested route (/account/my-models), asserting both headers are present. Ran the full local test suite and typecheck; no regressions.

### DIFF
--- a/enter.pollinations.ai/src/routes/account.ts
+++ b/enter.pollinations.ai/src/routes/account.ts
@@ -24,6 +24,7 @@ import { and, desc, eq } from "drizzle-orm";
 import { drizzle } from "drizzle-orm/d1";
 import type { Context } from "hono";
 import { Hono } from "hono";
+import { createMiddleware } from "hono/factory";
 import { HTTPException } from "hono/http-exception";
 import { describeRoute, resolver } from "hono-openapi";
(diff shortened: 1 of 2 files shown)


voodoohop:
- Thanks @dseeker for the thoughtful implementation. We selected and merged #13852 for #13771, so I’m closing this alternative to keep one path.

voodoohop:

voodoohop:

PR #13893, opened 2026-08-25
feat: add Seedance 2.5 reference images

ElliotEtag:
- Adds OpenRouter-compatible input_references for up to 30 images on Seedance 2.5, while keeping image reserved for first/last-frame control.
- Exposes the capability through the existing image/video routes and public model metadata; no alias, provider, price, duration, resolution, fallback, or secret changes.
- Rejects unsupported models, more than 30 references, and references combined with frame images before provider submission.
- Direct 480p and local 720p/audio E2E passed. The 30-reference stress request completed once upstream in 329s, cached after client timeout, retried in 0.773s, and debited exactly 0.9248 Pollen once.
- Both workers type-check; formatting and 174 focused/catalog/pricing tests pass. Draft because the maximum stress case exceeded the documented 300s synchronous window.

### DIFF
--- a/gen.pollinations.ai/src/image/handler.ts
+++ b/gen.pollinations.ai/src/image/handler.ts
@@ -528,7 +528,9 @@ export async function generateImageOrVideoResponse(
     c.var.track.setPricingInput({
         resolution: safeParams.resolution,
         quality: safeParams.quality,
-        hasImage: (safeParams.image?.length ?? 0) > 0,
+        hasImage:
+            (safeParams.image?.length ?? 0) > 0 ||
+            (safeParams.input_references?.length ?? 0) > 0,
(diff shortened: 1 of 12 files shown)


ElliotEtag:
Superseded by #13933, rebuilt from current main. The replacement preserves max_reference_images as the video frame cap, adds a separate conservative max_input_references field, limits the change to native GET routes, forwards URLs directly, and leaves /v1/images/generations unchanged.

voodoohop:
Thanks @ElliotEtag for contributing to #13901. We selected #14010 because it offered the clearest small API for reference images, videos, and audio, with correct billing and focused tests. A 1-Pollen contributor bonus is now available on your Quests page.

Issue #13904, opened 2026-08-25
[QUEST] Add a reasoning quickstart to the API docs


voodoohop:
### How this quest works

Multiple contributors may submit PRs. We will select the best completed approach, preferring correctness, clarity, and minimal code. The selected contributor will be assigned before merge and receive the reward.

### Goal

Make it immediately clear how to request reasoning through the Pollinations text API.

### Outcome

- Add one concise `POST /v1/chat/completions` reasoning example.
- Add one concise `POST /text` reasoning example.
- Use the existing `reasoning_effort` parameter and point readers to model metadata for reasoning support.
- Update the editable API-doc sources so the generated reference includes the guidance.

Minimal approaches are preferred. Do not change runtime behavior, manually edit generated `APIDOCS.md`, add a provider matrix, or introduce new parameters.

### Existing code

- [`reasoning_effort` request schema](https://github.com/pollinations/pollinations/blob/main/shared/schemas/openai.ts)
- [Text-generation docs source](https://github.com/pollinations/pollinations/blob/main/gen.pollinations.ai/src/docs/text-generation.md)
- [API code samples](https://github.com/pollinations/pollinations/blob/main/gen.pollinations.ai/src/routes/docs-samples.ts)

### Reward

5

davealan74:
PR opened at https://github.com/pollinations/pollinations/pull/13917

iotserver24:
Implemented **reasoning quickstart in the API docs** in [PR #13944](https://github.com/pollinations/pollinations/pull/13944). The PR is submitted and the merge conflicts have been resolved; CI checks are running. This issue will close automatically when the PR is merged.

voodoohop:
@xiaotian1171, we selected #13927 🎉

It is the cleanest solution: one editable docs file, both POST examples, and no duplicated or unrelated changes. Please reply here so GitHub lets us assign you the 5-Pollen reward.

xiaotian1171:
Thanks! Excited to receive the reward — ready for assignment.

PR #13917, opened 2026-08-25
docs: add reasoning quickstart to text API docs

davealan74:
Fixes #13904

- Adds a `### Reasoning` section to `gen.pollinations.ai/src/docs/text-generation.md` explaining the `reasoning_effort` parameter, its accepted values, and how to check model metadata (`/text/models` or `/v1/models`) to identify which models support it — with one JSON body example for the OpenAI-compatible endpoint and one GET URL example for the simple text endpoint.
- Adds a `Reasoning (cURL)` code sample to `gen.pollinations.ai/src/routes/docs-samples.ts` under `post /v1/chat/completions` (using `openai-large` + `reasoning_effort: "medium"`).
- Adds a `Reasoning (cURL)` code sample under `get /text/{prompt}` (same model, `reasoning_effort=medium` as a query parameter).

No runtime behavior changed. `APIDOCS.md` not touched — the generated reference will pick up these sources on the next doc-regen run.


davealan74:
Fixed the test timeout in 7ff168d.

- `recraft-v4.1-vector` has `paidOnly: true` in the registry — the 402 path was correct but unreachable within 5 s
- Root cause: `checkBalance` called `getModelStats` (Tinybird HTTP) before checking pack balance, so the test timed out waiting for Tinybird in CI
- Fix: for `paidOnly` models, check `balance.getBalance` first and throw 402 immediately if pack balance ≤ 0; skip the Tinybird call entirely in that case

voodoohop:
Thanks @davealan74 — we selected #13927 as the clearest minimal quickstart for #13904. A 2-Pollen contribution bonus is now available on your Quests page.

Issue #13905, opened 2026-08-25
[QUEST] Show model limits in the Models dashboard


voodoohop:
### How this quest works

Multiple contributors may submit PRs. We will select the best completed approach, preferring correctness, clarity, and minimal code. The selected contributor will be assigned before merge and receive the reward.

### Goal

Make important model limits visible without making the Models dashboard feel more crowded.

### Outcome

- Show the context window when `context_length` is available.
- Show the supported duration range or fixed duration for video models using the existing duration metadata.
- Preserve the fields through the existing catalog-to-dashboard mapping.
- Keep the presentation compact and accessible on desktop and mobile.

Minimal approaches are preferred. Do not change the API, registry values, pricing, sorting, filtering, or add a separate model-details view.

### Existing code

- [`/models` response mapping](https://github.com/pollinations/pollinations/blob/main/shared/registry/model-info.ts)
- [Dashboard catalog mapping](https://github.com/pollinations/pollinations/blob/main/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts)
- [Shared model rows](https://github.com/pollinations/pollinations/tree/main/enter.pollinations.ai/frontend/src/components/models)

### Reward

5

davealan74:
Opened PR #13919 with an implementation:

- Shows `"Xk ctx"` (e.g. `128k ctx`) with a tooltip for models that expose `context_length`
- Shows duration range `"X–Ys"` or fixed `"Xs"` with a tooltip for video models
- Threaded through `ApiModelInfo` → `ModelPrice` → `ModelRow` with no API/registry/pricing changes

iotserver24:
Implemented **model limits in the Models dashboard** in [PR #13944](https://github.com/pollinations/pollinations/pull/13944). The PR is submitted and the merge conflicts have been resolved; CI checks are running. This issue will close automatically when the PR is merged.

Creatneworld:
Hi @voodoohop — this quest candidate has a ready implementation: [PR #14052](https://github.com/pollinations/pollinations/pull/14052) surfaces context-window and video-duration limits as compact chips on the Models dashboard (4 files, ~107 insertions). The formatting check just failed on import ordering — I've fixed it with biome and pushed, CI is re-running. If this candidate is adopted as an official POLLEN-QUEST bounty, the PR is ready to review. Thanks!

voodoohop:
[...]

PR #13923, opened 2026-08-25
feat: show context window length in model cards

rekty:
## Summary

Show the context window length in model catalog cards when available.

## Changes

- Added `contextLength` to `ModelPrice` type
- Mapped `context_length` from API response through catalog-to-dashboard pipeline
- Display context length as human-readable text (e.g. "128K context", "1.0M context") below the brand link

## How it works

1. API response includes `context_length` field for supported models
2. `model-catalog.ts` maps it to `contextLength` on ModelPrice
3. `model-row.tsx` renders it as a compact text label

Context length is only shown when the API provides it — no empty states or fallbacks.

Closes #13905


voodoohop:

PR #13927, opened 2026-08-25
docs: add reasoning quickstart to text generation guide

xiaotian1171:
Closes #13904

## Changes
- Adds a concise **Reasoning** section to the text-generation docs source.
- One `POST /v1/chat/completions` example and one `POST /text` example, both using the existing `reasoning_effort` parameter.
- Points readers to model metadata for checking reasoning support. No runtime changes, no generated docs touched.

## Test plan
- Docs build renders the new section with both examples.


Circuit-Overtime:
Thanks for keeping this implementation documentation-only and compact. This is the right branch to continue for #13904.

Please tighten the model-support wording before merge:

- Model metadata exposes reasoning support, not whether every effort level is adjustable.
- none does not disable reasoning for every reasoning model.
- minimal is not accepted or preserved by every model.
- Unsupported effort handling varies between rejection, normalization, and omission, so avoid promising that non-reasoning models ignore it.

The two request examples can stay; the surrounding claims should describe only the stable public contract.

xiaotian1171:
Thanks for the quick review! Pushed ccdaf011b with the wording tightened along all four points: the docs now say metadata reports reasoning *support* rather than per-level adjustability, drop the promise around `none` disabling reasoning, and describe handling of unsupported values as varying by model (honor / reject / normalize / drop) instead of claiming they're ignored. The two request examples are untouched.

PR #13955, opened 2026-08-26
fix: show model limits in Models dashboard

iotserver24:
## Summary

Displays context length and video duration limits in the Models dashboard.

Fixes #13905

Co-authored-by: supercharge AI by iotserver24 <iotserver24@users.noreply.github.com>


Circuit-Overtime:
Technical review

- The patch removes existing tool-calling display, per-user limits, duration fields, and unrelated registry and proxy behavior.
- It introduces a new duration_seconds shape instead of preserving the existing min_duration, max_duration, and default_duration metadata required by #13905.
- #13960 stays within the catalog mapping and display path and supersedes this implementation.

voodoohop:

PR #13962, opened 2026-08-26
refactor: unify HttpError into UpstreamError hierarchy

### LINKED ISSUE #13551: Unify error classes across gen (HttpError vs UpstreamError)
`gen` throws two near-equally used error classes with no rule for which:

| class | defined | thrown | concentrated in |
|---|---|---|---|
| `HttpError` | `shared/http-error.ts` (extends `Error`) | 118 | `src/image/` (28 files) |
| `UpstreamError` | `shared/error.ts` (extends hono `HTTPException`) | 116 | `src/routes/`, `src/text/` |

Plus 76 bare `throw new Error(`, 20 `HTTPException`, and provider-specific `UserImageError` / `InferenceportError` / `ReplicateError` / `FalError`.

The split tracks subsystem lineage, not intent — new code inherits whichever its neighbour used. Concrete example: the three community endpoint callers do the same thing (`redirect: "manual"` + `AbortSignal.timeout` → 502 on connect failure) but `image/communityEndpoint.ts` throws `HttpError` while `audio/communityEndpoint.ts` throws `UpstreamError`, so their timeout guards can't be shared.

Investigate:
- Do the two classes carry different information, or is one a superset? (`UpstreamError` adds `requestUrl`, `upstreamStatus`, `responseBody`, `upstreamHeaders`, `errorCode`)
- Are they caught/serialized differently at the route boundary, and does either lose detail?
- Can `HttpError` fold into `UpstreamError`, or do both belong behind one type?
- Should bare `throw new Error(` in request paths be banned by a lint rule?

Deliverable: one documented rule for which class to throw, and a migration if the classes can merge.

rekty:
## Summary

Fixes #13551 — unifies `HttpError` and `UpstreamError` into a single class hierarchy.

## Problem

The codebase had two near-equally used error classes with no rule for which:
- `HttpError` (extends `Error`) — 126 uses in `src/image/`
- `UpstreamError` (extends Hono `HTTPException`) — 127 uses in `src/routes/`, `src/text/`

`UpstreamError` is the superset (carries `requestUrl`, `upstreamStatus`, `responseBody`, `upstreamHeaders`, `errorCode`). The split tracked subsystem lineage, not intent. Image handlers had to manually convert `HttpError` → `UpstreamError` at route boundaries, losing context.

## Solution

**`HttpError` now extends `UpstreamError`** instead of plain `Error`.

This is a backward-compatible, single-file change:
- All 126 `instanceof HttpError` checks still work
- All `instanceof UpstreamError` checks now also catch `HttpError`
- `HttpError` automatically gets the rich error envelope (upstreamHost, upstreamStatus, upstreamBody, upstreamHeaders) in the error handler
- Constructor maps positional args `(message, status, details, upstreamUrl, errorCode)` → `UpstreamError` options bag

## Migration Rule (documented in JSDoc)

- **New code** must throw `UpstreamError` directly
- **Existing** `HttpError` sites may keep using the class for backward compatibility
- Do **not** add new fields to `HttpError`; use `UpstreamError` options instead

## Files Changed

- `shared/http-error.ts` — `HttpError` extends `UpstreamError` with `@deprecated` tag

### DIFF
--- a/enter.pollinations.ai/test/account-usage.test.ts
+++ b/enter.pollinations.ai/test/account-usage.test.ts
@@ -324,6 +324,31 @@ test("GET /api/account/usage forwards table filters to the pipe", async ({
     mocks,
 }) => {
     await mocks.enable("tinybird");
+    mocks.tinybird.state.usageResponse = [
+        {
+            cursor_event_id: "event-1",
+            timestamp: "2026-04-14 12:10:00",
+            type: "generate.text",
+            model: "gpt-a",
+            api_key_id: "key_a",
+            api_key: "alpha",
+            api_key_type: "secret",
+            meter_source: "tier",
+            input_text_tokens: 10,
+            input_cached_tokens: 0,
+            input_audio_tokens: 0,
+            input_audio_seconds: 0,
+            input_image_tokens: 0,
+            output_text_tokens: 20,
+            output_reasoning_tokens: 0,
+            output_audio_tokens: 0,
+            output_audio_seconds: 0,
+            output_image_tokens: 0,
+            output_video_seconds: 0,
+            cost_usd: 1,
+            response_time_ms: 123,
+        },
+    ];
 
     const response = await SELF.fetch(
         "http://localhost:3000/api/account/usage?days=30&limit=15&api_key_ids=key_b,key_a,key_a&models=gpt-b,gpt-a,gpt-a",
--- a/gen.pollinations.ai/src/image/handler.ts
+++ b/gen.pollinations.ai/src/image/handler.ts
@@ -204,38 +204,33 @@ function safeUpstreamUrl(value: string | undefined): URL | undefined {
     }
 }
 
-function throwImageError(error: unknown): never {
-    if (error instanceof UpstreamError) throw error;
-
-    // Content-policy rejections from any provider (DashScope green-net, Replicate
-    // moderation, Vertex safety, Azure content safety) are client errors, not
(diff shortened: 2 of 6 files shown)


Circuit-Overtime:
Technical review

- Split out commit b216e3a8d; the Qwen provider and pricing change is unrelated to the error-hierarchy change.
- Making HttpError extend UpstreamError changes control flow in throwImageError. Its first UpstreamError guard now catches HttpError before moderation mapping, status classification, and response-body handling.
- Preserve the HttpError-specific behavior and add regression coverage for moderation, mapped status, and upstream response details.

rekty:
Hi @Circuit-Overtime, thanks for the technical review! Here's what I've addressed:

1. ✅ **Split out Qwen commit** — The Qwen provider/pricing change was already in a separate commit that has been removed from this branch. This PR now only contains the error hierarchy refactor.

2. ✅ **Control flow in throwImageError** — HttpError is now checked BEFORE UpstreamError in the guard chain. Since HttpError extends UpstreamError, this preserves moderation mapping, status classification, and response-body handling that the UpstreamError-only path skips.

3. ✅ **Regression tests added** —  covers: moderation wrapping, status classification (422→400), upstream response details preservation, content-policy context fields, plain UpstreamError pass-through, and fallback 500 wrapping.

4. ✅ **HttpError.name writable** — Subclasses like UserImageError can still override .

The branch has been rebased on latest main. All CI checks pass. Ready for review!

PR #13960, opened 2026-08-26
feat: show model limits in the Models dashboard (#13905)

wiradifit:
Closes #13905

## What

- **Preserve the fields through the existing mapping**: `context_length` and the video duration metadata (`min_duration` / `max_duration` / `default_duration`) already exposed by the registry-to-API layer (landed in #11145) now flow through `baseModelPrice` into `ModelPrice` (`contextLength`, `duration`).
- **Show them compactly**: one muted `text-xs` line under the model id, e.g.:

  ```
  400K context · 4–10s video
  ```

- Duration renders as a fixed value when min = max, a range otherwise; partial metadata degrades gracefully (`up to 15s`, `from 4s`, or default only).

## What this deliberately does not do

Per the quest constraints: no API changes, no registry value changes, no pricing/sorting/filtering changes, and no separate model-details view.

## Verification

- `npx vitest run test/cost-variants.test.ts test/effective-prices.snapshot.test.ts test/pricing-data.test.ts test/community-endpoint-envelope-migration.test.ts test/d1-tinybird-sync.test.ts` — **99 passed** (includes 3 new tests: context mapping + formatting pins, duration mapping + rendering, absence case renders nothing)
- `npx tsc --noEmit` — clean
- `npx biome check` on all touched files — clean
- `npm run build:frontend` — production build succeeds with the new row content

PR #13967, opened 2026-08-26
docs: add reasoning quickstart to API docs (#13904)

wiradifit:
Closes #13904

## What

- **`text-generation.md`** (editable doc source): one concise `### Reasoning` section — the `reasoning_effort` enum from `shared/schemas/openai.ts`, a JSON example for `POST /v1/chat/completions`, and an equivalent-body note + example for `POST /text`. Points readers at each model's metadata (`reasoning` capability in `/models`) for support and accepted levels.
- **`docs-samples.ts`**: one `cURL (reasoning)` code sample on `post /v1/chat/completions`, one on `post /text`.

## Accuracy check (done live before writing)

- Verified `reasoning_effort` exists with that exact enum in both endpoints' schemas via `gen.pollinations.ai/openapi.json`.
- Live-probed `POST /text` with `"reasoning_effort": "max"` on a model that doesn't accept it: upstream returns a 400. The doc says so explicitly instead of implying values are silently clamped/ignored — this is the trap most users would hit.

## Constraints honored

No runtime changes, no manual edits to generated APIDOCS.md, no provider matrix, no new parameters.

### DIFF
--- a/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
+++ b/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
@@ -57,6 +57,9 @@ export type ApiModelInfo = {
     tools?: boolean;
     reasoning?: boolean;
     context_length?: number;
+    min_duration?: number;
+    max_duration?: number;
+    default_duration?: number;
     voices?: string[];
     is_specialized?: boolean;
     paid_only?: boolean;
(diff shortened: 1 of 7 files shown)


Circuit-Overtime:
Technical review

- Rebase onto main and remove the #13960 ancestor so this PR contains only the documentation change for #13904.
- The Models API exposes whether reasoning is supported; it does not expose the effort levels accepted by each model. Correct that statement.
- Remove the hard-coded default-model effort range unless it comes from a stable public contract.
- Keep the final diff limited to the two reasoning quickstart files.

voodoohop:
Thanks @wiradifit — we selected #13927 as the clearest minimal quickstart for #13904. A 2-Pollen contribution bonus is now available on your Quests page.

PR #13975, opened 2026-08-26
docs: publish concise community-model naming guidance (#12885)

wiradifit:
Closes #12885

## What

- **`COMMUNITY_MODEL_NAMING.md`** — one concise guide covering the three public fields:
  - **Model ID**: permanent `{username}/{model-id}`; avoid mutable details (pricing, provider routing, context size, fast-changing versions); good/avoid examples
  - **Display Title**: changeable, plain-language; example table across direct upstream, custom fine-tune, router, and rebranded models
  - **Description**: transparent one-liner, especially for routers/rebrands
  - Quick-reference table (purpose + mutability per field)
- **Linked from `BRING_YOUR_OWN_MODEL.md`** (the existing publishing docs)
- **Linked from the Add model form** — the Model ID field helper now points to the guide (models only; agent IDs keep their existing helper)

Community models explicitly do **not** need the stricter first-party convention from #13587 — the guide says so.

## Constraints honored

No schema enforcement, dashboard warnings, aliases, migrations, automatic renaming, compatibility layers, or changes to existing model IDs.

## Verification

- `tsc --noEmit`: clean · biome: clean · markdown links resolve to repo paths

### DIFF
--- a/BRING_YOUR_OWN_MODEL.md
+++ b/BRING_YOUR_OWN_MODEL.md
@@ -43,6 +43,8 @@ Owners receive 75% of the Pollen spent on their models. Paid and Quest Pollen ea
 
 The upstream credential is used by Pollinations to proxy requests to your endpoint. Do not place it in a model name, description, public URL, or example.
 
+See the [Community Model Naming Guide](./COMMUNITY_MODEL_NAMING.md) for tips on choosing a stable model ID and clear display title.
+
 ## Register with the CLI
 
 The CLI manages text, image, and transcription model registrations. Sign in, test the endpoint, then create the model:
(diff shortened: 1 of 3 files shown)


Circuit-Overtime:
Thanks for keeping this focused to one guide and two links. This is the strongest implementation for #12885.

One documentation correction is needed before merge:

- The guide says a model ID cannot change after registration, but the current account update schema accepts name and persists it.
- Describe the ID as something publishers should keep stable because changing it breaks callers and saved URLs.
- If IDs should become immutable, that is a separate behavior change outside this documentation quest.

The remaining guide, examples, and link placement are concise and within scope.

voodoohop:
Thanks @wiradifit. We selected and merged #14050 because it reduced the result to one paragraph in the existing guide. Your concise field tables and examples helped shape the guidance. A 1-Pollen contribution bonus is available on your Quests page.

Issue #13800, opened 2026-08-24
[QUEST] Add a 12-hour delay to community-model price changes


voodoohop:
### How this quest works

Multiple contributors may submit PRs. We will select the best completed approach, preferring correctness, clarity, and minimal code. The selected contributor will be assigned before merge and receive the reward.

### Goal

Give users 12 hours of notice before a public community model changes price or becomes public.

### Outcome

- A model's first price can take effect immediately.
- Price changes remain immediate while a model is private.
- A public model keeps its existing price for 12 hours after its owner submits a new price, then switches automatically.
- Changing a model from private to public takes effect after 12 hours.
- Changing a model from public to private remains immediate.
- Pending public changes and when they take effect are visible to the owner and model users.

Keep the implementation minimal. Do not add a general notification system.

### Parent issue

#13546

### Reward

8

davealan74:
Opened PR #13807 implementing the 12-hour price-change delay.

Key behavior:
- First price on a new model is always immediate
- Price/paidOnly changes on public models queue for 12 hours; current price stays in effect until then
- Private→public transitions queue for 12 hours
- Public→private is immediate and clears any pending change
- Owners see pending changes and their `effectiveAt` timestamp in the API response
- Lazy evaluation in both enter and gen workers — no background job needed

Marcus-Mok-GH:
@voodoohop I've implemented the 12-hour delay for community model price changes as requested. The PR is ready for review: https://github.com/pollinations/pollinations/pull/13864

Marcus-Mok-GH:
Solved in #13864

Marcus-Mok-GH:
Solved in PR [#13864](https://github.com/pollinations/pollinations/pull/13864).

This PR adds:
- `pendingPayload`, `pendingVisibility`, `pendingAt` fields to the community endpoint schema
- `PRICE_CHANGE_DELAY_MS = 12 * 60 * 60 * 1000` constant
- Queue logic: stores pending price/visibility changes on public models
- Drizzle migration: `0055_price_change_delay.sql`

Expected effect: first price on new models is immediate; all other price/visibility changes on public models are queued for 12 hours before taking effect.

voodoohop:
- Completed in #13807: public price changes and private-to-public publishing now get a 12-hour notice window while current terms stay active.
- Pending terms and their effective time are visible to owners and model users.
- Thanks @davealan74 — the issue is assigned for the 8-Pollen quest reward.

[...]

PR #13985, opened 2026-08-26
fix: gate community model recreation on a model-id cooldown

tomdacatto:
## Summary
- Follow-up to #13807, which added a 12-hour delay on price/visibility changes for public community models but left two gaps: the frontend never read the new `pending` field, and delete+recreate under the same name bypassed the delay entirely (#13800).
- Frontend: `endpointToForm` now shows the staged `pending` value (price, paidOnly, visibility) instead of the stale current one when reopening the edit dialog, plus a "Changes queued" banner/badge so owners don't re-submit thinking a save failed.
- Backend: deleting a public proxy model snapshots its effective price/paidOnly/pricing-mode into a new `community_endpoint_cooldown` table keyed by model id (`owner/name`), deliberately with no FK to `community_endpoint` so it survives that row's deletion.
- Recreating under the same name while the cooldown is active clamps any price/paidOnly increase to the snapshot and stages the requested value as `pending`, exactly like a normal update would. Same/lower price, a different name, or an expired cooldown all stay immediate.

Note: the original spec for this fix proposed writing the cooldown snapshot when a price change is *staged* on update. That's inconsistent with its own test matrix — a public model that's never had an update staged would have no cooldown row at all, so delete+recreate at a higher price would still be immediate. Snapshotting on **delete** instead closes the loophole in every case (fresh create, staged-but-unmatured change, expired cooldown) without touching the existing update-path code or its 10 passing tests.

## Test plan
- [x] `npx vitest run test/integration/community-endpoint-price-delay.test.ts` — 17/17 pass (10 pre-existing + 7 new cooldown cases: higher-price recreate staged, lower-price recreate immediate, different-name immediate, expired-cooldown immediate, staged-change survives delete, private-delete writes no cooldown, image pricing-mode-switch guarded).
- [x] `npx vitest run` on the other `community-endpoint-*` suites — 33/33 pass.
- [x] `npx tsc --noEmit` — no new errors introduced (pre-existing errors are all in unrelated unbuilt `@pollinations/ui` exports / missing `ai` sdk types).
- [x] `npx biome check --write` on all changed files.


Circuit-Overtime:
Thanks @tomdacatto for following through on the delete/recreate gap and the missing pending-state UI. Two changes are needed before this is safe to merge:

- The [delete handler](https://github.com/pollinations/pollinations/blob/7150f7464e053767d74ea36f37f0679c7b209966/enter.pollinations.ai/src/routes/community-endpoints.ts#L973-L985) deletes the endpoint before writing its cooldown. A failed username lookup/insert, or a recreate racing between those statements, leaves the model deleted with no cooldown and reopens the bypass. Resolve the snapshot inputs first, then commit the cooldown upsert and endpoint deletion atomically (for example with a D1 batch), with a regression test for failure/race behavior.
- `visibilitySnapshot` and `visibilityExceeds` are unreachable state: cooldowns are only recorded for effectively public models, and the writer always stores `public`. Please remove that column/branch unless private deletions are intentionally meant to create cooldowns too. This keeps the fix scoped to price/`paidOnly` and avoids carrying a second visibility policy that cannot execute.

The pending-value form mapping itself matches the existing delayed-change response shape.

Circuit-Overtime (approved):
Thanks @tomdacatto — the two blocking findings from the earlier review are addressed:

- Snapshot inputs, including the GitHub username, are resolved before mutation; the cooldown upsert and endpoint deletion now commit in one D1 batch, closing both the partial-failure and concurrent-recreation windows.
- The unused visibility snapshot column and unreachable visibility staging branch have been removed, leaving the cooldown scoped to price and `paidOnly`.
- The regression coverage now confirms that failed snapshot resolution leaves the endpoint intact and that the subsequent atomic delete/recreate path remains protected.

The pending-value UI still matches the API response and the full feature diff is coherent. From the community-endpoint implementation side, this is ready to merge.

Non-blocking scope note: the one-line Tinybird mock change in `gen.pollinations.ai/test/generation-vcr.test.ts` is unrelated to this feature; it would be cleaner as a separate test-fix PR if it is still needed on current `main`.

voodoohop:
Thanks @tomdacatto — #14051 used the simpler pending-publication approach and now covers delete/recreate without a cooldown table. A 2-Pollen contribution bonus is available on your Quests page.

PR #13995, opened 2026-08-26
feat(models): show context window and video duration in Models dashboard

Apollohzl:
## Summary
Implements pollinations/pollinations#13905 — surface important model limits in the Models dashboard without adding clutter.

## What changed
- `shared/registry/model-info.ts` already exposes `context_length` and video duration metadata (`min_duration` / `max_duration` / `default_duration` / `allowed_durations`) in the `/models` response. The frontend catalog mapping was not threading these into the dashboard.
- `ApiModelInfo` (model-catalog.ts): added `min_duration`, `max_duration`, `default_duration`, `allowed_durations` so the existing duration metadata is captured from the catalog.
- `ModelPrice` (types.ts) + `baseModelPrice`: thread `contextLength` and the duration fields through to the dashboard model.
- New `ModelLimitChips` (model-status-chips.tsx): compact, tooltip-bearing chips showing:
  - Context window as `Xk ctx` / `XM ctx` (e.g. `128k ctx`) with an exact-token tooltip, when `context_length` is available.
  - Video duration as a range `X–Ys` or fixed `Xs` (e.g. `4–8s`) with a tooltip, for video models.
- Rendered in both `ModelRow` (desktop) and `MobileModelRow` (mobile) status rows, so it stays compact and accessible on both.

## Constraints honored
- No API, registry, pricing, sorting, or filtering changes.
- No separate model-details view added.
- Only the catalog → dashboard mapping was extended.

## Verification
- `biome check` (lint) passes on all five changed files with no errors.
- Full frontend build / type-check could not be executed locally in this environment; relying on CI.

### DIFF
--- a/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
+++ b/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
@@ -57,6 +57,10 @@ export type ApiModelInfo = {
     tools?: boolean;
     reasoning?: boolean;
     context_length?: number;
+    min_duration?: number;
+    max_duration?: number;
+    default_duration?: number;
+    allowed_durations?: number[];
     voices?: string[];
     is_specialized?: boolean;
     paid_only?: boolean;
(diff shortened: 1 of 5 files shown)


voodoohop:

PR #13999, opened 2026-08-26
docs: add reasoning quickstart to text generation docs

xxooxannn:
Fixes #13904

Added a quick Reasoning section to the text generation docs. Shows how to use `reasoning_effort` with `POST /v1/chat/completions` and `POST /text`, and points to the `reasoning` flag in `/text/models` so you can check which models actually support it.

- Text generation guide now has two minimal examples (chat completions + text)
- Updated `reasoning_effort` description to mention the `/text/models` flag
- Documented the `reasoning_effort` query param on `GET /text/{prompt}` which had no description before

No runtime changes, APIDOCS gets regenerated on deploy.

### DIFF
--- a/gen.pollinations.ai/src/docs/text-generation.md
+++ b/gen.pollinations.ai/src/docs/text-generation.md
@@ -9,6 +9,30 @@ Generate text responses using AI models. Fully compatible with the OpenAI Chat C
 
 **Available models:** {{TEXT_MODELS}}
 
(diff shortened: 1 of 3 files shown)


voodoohop:
Thanks @xxooxannn — we selected #13927 as the clearest minimal quickstart for #13904. A 2-Pollen contribution bonus is now available on your Quests page.

PR #14005, opened 2026-08-26
feat(models): add Try in Play action to model cards

### LINKED ISSUE #13903: [QUEST] Open a model directly in Play
### How this quest works

### Goal

Let users open any supported model from the model catalog and try it immediately in the existing Play page.

### Outcome

- Each model card has a compact icon action with a **Try in Play** tooltip and accessible label.
- The action opens `/play?model=<model-id>`.
- Play preselects that existing model and shows its model category.
- The behavior works for the model types already supported by Play.

Minimal approaches are preferred. Keep the model cards compact; do not add a full text button, a new playground, backend changes, or a broader Play redesign.

### Existing code

- [Model catalog rows](https://github.com/pollinations/pollinations/tree/main/enter.pollinations.ai/frontend/src/components/models)
- [Play page](https://github.com/pollinations/pollinations/blob/main/pollinations.ai/src/ui/pages/PlayPage.tsx)

### Reward

5

TRDEFI:
﻿Closes #13903

Each model row now shows a compact icon action that opens `/play?model=<id>` with a "Try in Play" tooltip and accessible label. `PlayPage` reads `?model=` from the URL and preselects it when it matches a known model, so the existing model category selection updates automatically. Works for all model types Play already supports.

Kept the card compact - icon-only, no new playground or backend changes.

### DIFF
--- a/enter.pollinations.ai/frontend/src/components/models/model-row.tsx
+++ b/enter.pollinations.ai/frontend/src/components/models/model-row.tsx
@@ -373,7 +373,7 @@ export const ModelRow: FC<ModelRowProps> = ({ model }) => {
                             </div>
                         )}
                     </div>
-                    <div className="mt-1.5 flex min-w-0 flex-wrap items-center gap-1.5">
+                    <div className="mt-1.5 flex min-w-0 flex-wrap items-center gap-2">
                         <ModelStatusChips
                             showNew={showNew}
                             showAlpha={showAlpha}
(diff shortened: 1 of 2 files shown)


voodoohop:
Thanks @TRDEFI — we selected #13940 for its compact shared UI and complete Play preselection for #13903. A 2-Pollen contribution bonus is now available on your Quests page.

PR #14053, opened 2026-08-27
feat: add Pollen gift codes

ElliotEtag:
- Adds guest Pollen gift purchases inside the existing Pollen-page UI: no buyer account required, single-use redemption into a signed-in recipient's Paid balance, and code delivery on screen and Stripe invoice.
- Shares checkout construction with ordinary purchases: existing payment methods, prices, fees, automatic tax and invoices. Keeps required guest name/address collection, Stripe-hosted payment protections and the existing signed-in card gate; removes guest failed-card fingerprint storage.
- Uses one gift table. Cloudflare limits checkout to 5/minute per hashed IP, receipts to 10/minute per hashed IP, and redemption to 10/minute per signed-in user (previously 10/10 minutes). Limits are approximate and location-local; D1 redemption/balance updates remain atomic. All environments have separate binding namespaces.
- Keeps successful-refund invalidation and one-time credit reversal. Failed refunds, disputes, and refunds arriving before activation require manual handling; partial refunds invalidate the whole gift. Existing staging schemas need reconciliation with the consolidated migration before rollout.
- Verified 78 gift/Stripe integration tests, TypeScript, staging build, all environment limit configurations and Biome. UX unchanged; alternative PR #14637 untouched. No deployment or Stripe configuration changes.


Circuit-Overtime:
Thanks for the follow-up. The two payment-loss gaps are now addressed:

- Refund state is persisted by PaymentIntent, so a successful refund received before checkout completion still voids the gift.
- Stripe dispute loss and reinstatement now debit and restore redeemed gifts exactly once, while preserving any remaining active refund/dispute loss.

Focused coverage includes out-of-order delivery and replayed dispute events. No further technical changes requested from my side.

ElliotEtag:
Blocking findings in the payment-loss follow-ups:

- `refund.failed` is routed to the gift handler, but the service ignores every status except `succeeded`. A failed refund therefore leaves the gift voided and redeemed Pollen debited even though Stripe returns our funds.
- Loss identity is only `paymentIntentId:kind`. Multiple refunds collapse into one row, so one failure can clear another successful refund. Use per-object keys: `refund:<refund.id>` and `dispute:<dispute.id>`.
- There is no terminal guard. Because Stripe does not guarantee event order, a stale loss event can reactivate a dispute after recovery. Store `terminal`; non-terminal events must never reopen it.
- `charge.dispute.created` always means lost funds here. Warning inquiries fire it before funds are withdrawn, so recipients are debited too early. Keep inquiries inactive until escalation.
- Ordinary pack refund/dispute events also create gift-loss rows. Check D1 gift state first, then Stripe gift metadata only on a miss so out-of-order gift events still work.

Required tests: failed-refund recovery, two refunds on one PaymentIntent, stale replay after dispute recovery, inquiry escalation, and non-gift events.

@voodoohop could you review this state machine before merge?

PR #14089, opened 2026-08-28
feat: add Pollinations OAuth to operations dashboards

ElliotEtag:
- Adds identity-only Pollinations OAuth 2 authorization code + PKCE to Enter; login uses a 60-second, single-use profile token and never creates a generation API key.
- Adds shared signed, origin-bound dashboard sessions; access requires the admin role in the Better Auth user database instead of a deployed email allowlist.
- Adds login/logout account menus to Economics, KPI, and Observability through one shared auth contract and three distinct public OAuth clients.
- Hardens return redirects, cancellation, token consumption, upstream error handling, Grafana header/logout behavior, and defaults Grafana users to Editor.
- Production callbacks are registered for both Pollinations and Myceli hosts. Merge prerequisite: secrets PR #14192 must merge first. Tests: Enter OAuth 21/21; auth 13/13 plus typecheck; Economics worker 9/9 plus typecheck; Economics and KPI builds; Observability Worker dry-run.


voodoohop:
From the weekly call: admin access to the ops dashboards should come from an admin role in the database, not an email allowlist. Requirements to be defined before this lands.

voodoohop (commented):
Better Auth should remain the single source of truth for admin access. The manual role parsing duplicates Better Auth and misses adminUserIds.

Minimal version:
- Keep the existing authorize → code → token → userinfo flow.
- Mark the three dashboards as trusted clients and skip consent.
- At token exchange, use Better Auth's server-side admin check; reject non-admins.
- Return the normal profile from userinfo.

No admin scope, role claim, dashboard permission, or new endpoints.

voodoohop:
- Superseded by #14255, which uses the official Better Auth OAuth provider instead of custom dashboard token handling.
- Menu reuse is preserved in #14478; the logout fix is included in #14255.

voodoohop:
- Closing because #14255 uses Better Auth’s official OAuth provider, so we do not have to maintain our own login tokens and token-exchange rules.
- Authentication stays separate from API-key delegation; dashboard backends decide who can access admin data.
- The menu reuse and logout fix are preserved in #14478 and #14255.

PR #14116, opened 2026-08-29
fix(news): clarify model ID change

voodoohop:
- Introduces the September 7 change before mentioning that users can adopt the new IDs today.
- Removes unexplained “canonical” and “publisher-qualified” terminology.
- Keeps the example and existing-ID compatibility guarantee.

Companion to #14115.


voodoohop:
Closing to keep the announcement cleanup in a single PR: #14115.

Issue #14119, opened 2026-08-29
[QUEST] Add Pollinations setup for Pi


voodoohop:
### How this quest works

Multiple contributors may submit PRs. We will select the best completed approach, preferring correctness, clarity, and minimal code. The author of the selected merged PR receives the reward; no issue assignment is required.

Experience with Pi is especially welcome. In your PR, briefly share your relevant background and how you verified the integration.

### Goal

Make Pi ready to use Pollinations through the Polli CLI.

### Outcome

- `polli harness pi on` leaves a new or existing user ready to run Pi with Pollinations.
- If Pi is missing, the official installation experience is offered.
- Polli login creates a dedicated key for Pi.
- Current compatible Pollinations models are available and a default can be selected.
- The Polli skill is available for Pollinations-specific capabilities.
- Existing Pi configuration and skills are preserved.
- `status` clearly reports whether setup is complete.
- `off` removes only Pollinations-owned changes.
- The Scalar documentation explains setup, status, model selection, and removal from a user perspective.
- The complete setup is verified end to end.

Pi intentionally does not include built-in MCP support. A new MCP extension or bridge is outside this quest.

### Resources

- [Polli harness foundation](https://github.com/pollinations/pollinations/pull/14107)
- [Pi repository and installation](https://github.com/earendil-works/pi)
- [Pi custom models](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/models.md)

### Reward

12 Pollen

davealan74:
Implemented in PR #14125.

**What was added:**
- `polli harness pi on|off|status` — connects Pi to Pollinations as a custom provider
- Writes `~/.pi/agent/models.json` (provider config), `auth.json` (dedicated child key), `settings.json` (default model + provider)
- Installs the `polli` skill to `~/.pi/agent/skills/polli/SKILL.md` (auto-discovered by Pi at startup)
- Offers install instructions when `pi` is not found in PATH
- Snapshot/restore system: `off` restores files byte-for-byte, or surgically strips only Pollinations entries if files were edited
- 14 new tests covering all acceptance criteria

iotserver24:
Pi harness is consolidated in PR #14129. See https://github.com/pollinations/pollinations/pull/14129

- Pi harness: Configures pollinations provider in models.json, auth key in auth.json, default model in settings.json.
- Security: API key is reused if already valid via shared resolveHarnessKey (no redundant key minting).
[...]

PR #14133, opened 2026-08-30
Dcode9/harness_pi

Dcode9:


voodoohop:
- Pi quest #14119 was completed by #14125.
- Thank you for the implementation and submission.

PR #14149, opened 2026-08-30
fix(community-models): apply price delay only to adverse changes

### LINKED ISSUE #13992: [PATCH]: apply the community-model delay only to adverse pricing changes (follow up to #13800)
## Problem

The 12-hour public community-model change delay introduced by #13800 currently treats consumer-beneficial pricing changes the same as adverse ones. A price reduction or disabling `paidOnly` does not need a notice period and delaying it prevents users from receiving the better terms immediately.

## Outcome

- A public-model update containing only price reductions takes effect immediately.
- Disabling `paidOnly` takes effect immediately.
- Any price increase or enabling `paidOnly` remains queued for 12 hours.
- A single update that mixes increases and reductions remains queued as one policy with one effective time.
- Private-model pricing and the existing visibility rules remain unchanged.
- The delete/recreate cooldown continues to prevent publishing higher terms immediately.
- Existing pending-change responses and UI continue to show the effective public terms and queued policy.

Keep the implementation within the existing pricing comparison and pending-payload flow; do not introduce field-level schedules or a notification system.

Follow-up to #13800.

rekty:
## Summary

Fixes #13992

The 12-hour public community-model change delay introduced by #13800 currently treats consumer-beneficial pricing changes the same as adverse ones. This patch fixes that.

### What changes

- **Price reductions** now take effect **immediately** (no 12-hour wait)
- **Disabling paidOnly** now takes effect **immediately**
- **Price increases** and **enabling paidOnly** remain queued for 12 hours as before
- **Mixed updates** (both increases and reductions in one update) are treated as adverse and queued as one policy

### Implementation

Added `isAdversePricingChange()` in `proxy-policy.ts` that returns `true` when:
- Any price field increased, OR
- paidOnly was enabled from `false` → `true`

The existing `delayPricing` gate in `community-endpoints.ts` now checks `isAdversePricingChange()` in addition to the existing conditions. Consumer-beneficial changes bypass the delay entirely.

### What is NOT changed

- Private-model pricing (unchanged)
- Visibility rules (unchanged)
- Delete/recreate cooldown (unchanged)
- Existing pending-change responses and UI (unchanged — the pending field still shows queued changes correctly)

### Tests

Added 3 new integration tests:
1. **Price reduction** → takes effect immediately, no pending
2. **Disabling paidOnly** → takes effect immediately, no pending
3. **Mixed increase + reduction** → queued as one policy (adverse)

### DIFF
--- a/enter.pollinations.ai/src/routes/community-endpoints.ts
+++ b/enter.pollinations.ai/src/routes/community-endpoints.ts
@@ -42,6 +42,7 @@ import {
     deriveCreateProxyPolicy,
     deriveUpdatedProxyPolicy,
     hasProxyPricingInput,
+    isAdversePricingChange,
     proxyPricingChanged,
     withoutProxyPricingChanges,
 } from "./community-endpoints/proxy-policy.ts";
@@ -821,7 +822,8 @@ export const communityEndpointsRoutes = new Hono<Env>()
                     targetVisibility === "public" &&
                     (currentVisibility === "public" ||
                         pendingVisibility === "public") &&
-                    hasProxyPricingInput(input);
+                    hasProxyPricingInput(input) &&
+                    isAdversePricingChange(targetBase, targetPolicy);
                 const pricingChanged = proxyPricingChanged(
                     targetBase,
                     targetPolicy,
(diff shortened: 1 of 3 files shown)


voodoohop:
Thanks for the work. We decided to keep one 12-hour policy for all public pricing changes. It is simpler to understand and avoids separate rules for price increases and reductions, so we won't merge this approach.

PR #14225, opened 2026-08-31
feat(sdk): use PKCE in PolliProvider

voodoohop:
- Replaces fragment token callbacks with OAuth authorization-code + PKCE.
- Keeps the existing `PolliProvider` API and stored token key unchanged.
- Uses a stable callback URL, then restores the app's original query/hash route.
- Covers state validation, code exchange, route restoration, and React StrictMode.

Tests: `npm test`, `npm run typecheck`, `npm run build`, Biome.


Circuit-Overtime:
Thanks for moving PolliProvider to authorization-code + PKCE while keeping the public API and stored token key stable.

One race needs addressing before merge: login() stores the pending state/verifier, then awaits crypto.subtle.digest before redirecting. If login() is invoked again during that await (for example, a double click), the second call overwrites storage while either promise can still win the redirect. The callback can then carry one state/challenge while storage contains the other verifier, causing an invalid-state or PKCE failure.

Circuit-Overtime (approved):
Thanks for the clean PKCE migration. The pending-login race is now closed by the in-flight guard, and the regression coverage confirms repeated login calls cannot replace the active transaction state. The token exchange contract, callback restoration, storage compatibility, and StrictMode behavior all look correct. Approved.

PR #14252, opened 2026-09-01
refactor: use Better Auth OAuth provider for dashboards

voodoohop:
- Replaces the custom dashboard login grant in #14089 with Better Auth’s official OAuth provider
- Uses standard authorization-code + PKCE, token, and userinfo endpoints
- Limits access to existing admins and three trusted dashboard clients with exact redirects
- Keeps `/api/oauth/*` focused on API-key delegation
- Adds end-to-end provider tests and API-key migration coverage

Stacked on #14089.


voodoohop:
Superseded by #14255, which keeps the official Better Auth provider but moves the admin check to the shared dashboard backend.

Issue #14121, opened 2026-08-29
[QUEST] Add Pollinations setup for OpenClaw


voodoohop:
### How this quest works

Multiple contributors may submit PRs. We will select the best completed approach, preferring correctness, clarity, and minimal code. The author of the selected merged PR receives the reward; no issue assignment is required.

Experience with OpenClaw is especially welcome. In your PR, briefly share your relevant background and how you verified the integration.

### Goal

Give OpenClaw users one reliable Polli CLI path for installing and using Pollinations.

### Outcome

- `polli harness openclaw on` leaves a new or existing OpenClaw installation ready to use Pollinations.
- If OpenClaw is missing, the official installation experience is offered.
- Polli login creates a dedicated key for OpenClaw.
- Current compatible Pollinations models are available without maintaining a second hardcoded model catalog.
- The user can select a default model.
- Pollinations capabilities are available through the Polli skill.
- The resulting setup works with the current OpenClaw release.
- Existing agents and unrelated configuration are preserved.
- `status` clearly reports whether the integration is ready.
- `off` removes only Pollinations-owned changes.
- The existing Pollinations OpenClaw integration is used as reference, and no competing or outdated setup path remains.
- The Scalar documentation explains setup, status, model selection, and removal from a user perspective.
- The complete setup is verified end to end.

Avoid introducing another manually maintained fallback matrix or model registry.

### Resources

- [Polli harness foundation](https://github.com/pollinations/pollinations/pull/14107)
- [Existing Pollinations OpenClaw integration](https://github.com/pollinations/pollinations/tree/main/apps/openclaw)
- [Existing OpenClaw setup script](https://github.com/pollinations/pollinations/blob/main/apps/openclaw/setup-pollinations.sh)
- [OpenClaw installation](https://docs.openclaw.ai/install)
- [OpenClaw model providers](https://github.com/openclaw/openclaw/blob/main/docs/concepts/model-providers.md)
- [OpenClaw skills](https://docs.openclaw.ai/skills)

### Reward

12 Pollen

davealan74:
Opened PR #14126 with the OpenClaw harness implementation.

`polli harness openclaw on` fetches live Pollinations models, mints a dedicated API key, writes the provider block to `~/.openclaw/openclaw.json`, and installs the Polli skill. `off` restores the original config byte-for-byte. 7 new tests alongside the existing suite (32/32 passing).

gggff123:
Opened PR #14130  with polli harness implementation.

[...]

PR #14253, opened 2026-09-01
feat(cli): add OpenClaw harness

xiaotian1171:
Fixes #14121

- Adds `polli harness openclaw on|status|off` using OpenClaw's native `models.providers.pollinations` block, mirroring the existing `apps/openclaw` setup script but filling the model list from the live catalog instead of a second hardcoded table.
- Checks that OpenClaw is installed before login, key creation, or configuration, and shows the official install command.
- Defaults to `kimi` like the existing setup script; an unrelated `agents.defaults.model.primary` is preserved and reported, and repeated `on --model` switches only within Pollinations models.
- Installs the Polli skill under `skills/polli/`; `off` restores the previous config byte-for-byte and removes only Pollinations-managed entries. Follows the same inline-JSON, no-new-shared-modules pattern as the merged `prime.ts`.

Note: #14121 was closed as completed, but `main` has no OpenClaw profile yet (`harnesses/index.ts` lists dsh/opencode/pi/prime). This PR fills that gap.

Tests: 16 focused lifecycle tests; full CLI suite 83/83 pass; biome clean on modified files. Verified against a real OpenClaw 2026.8.1 install: `on` registers 68 live-catalog models, `openclaw models list` shows them all with `kimi` as default, and `off` restores the config byte-for-byte.

### DIFF
--- a/CODING_HARNESSES.md
+++ b/CODING_HARNESSES.md
@@ -3,6 +3,7 @@
 Use `polli harness` to connect a supported coding harness to Pollinations. It handles Polli login, a dedicated API key, model setup, and any Pollinations capabilities supported by that harness.
 
 > **Available now:** DeepSeek Harness, OpenCode, Pi, and Prime Agent are integrated `polli harness` profiles. OpenClaw is coming soon.
+> **Available now:** DeepSeek Harness, OpenCode, Pi, Prime Agent, and OpenClaw are integrated `polli harness` profiles.
 
 ## Use a harness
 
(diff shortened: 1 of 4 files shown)


voodoohop:
- Selected #14343 as the smaller complete base.\n- Reused this PR’s real-release testing approach.\n- A 3-Pollen bonus is ready to claim for that contribution.\n- Closing this alternative so the quest has one maintained setup path.

Issue #14232, opened 2026-09-01
[QUEST] Build a Pollinations GIMP 3 image plug-in with BYOP


voodoohop:
### How this quest works

Multiple contributors may submit PRs. We will select the clearest, smallest completed approach. The author of the selected merged PR receives the reward.

### Goal

Build a real GIMP 3 plug-in that makes Pollinations image generation and editing available inside GIMP, with each user paying through their own Pollinations account.

### Outcome

- Users can connect and disconnect Pollinations inside GIMP through BYOP device authorization. The plug-in opens the approval page; users never paste an API key into GIMP.
- An App Key identifies the integration for attribution, while the resulting user authorization stays private and works across GIMP restarts.
- The model picker loads `/image/models` at runtime and exposes every image model available to the connected account, including community models. Model IDs are not hardcoded.
- A prompt can generate an image and add the result to GIMP as a new image or layer.
- For models that advertise image input, the active layer or selection can be sent for editing and the result returns as a new layer without altering the source.
- Model capabilities drive the available controls so unsupported options are not presented.
- Expired authorization, insufficient Pollen, network failures, and API errors have clear recovery messages.
- The PR includes reusable open-source code, installation instructions for supported desktop platforms, and focused tests or a reproducible end-to-end demonstration.

Keep it focused: masks, batch queues, history, video generation, and a separate web dashboard are not required.

### Resources

- [Existing GimpToPolli prototype](https://github.com/GermanIllan/pollinationsai)
- [Orange City Pollinations-powered GIMP plug-in](https://github.com/GermanIllan/orange_city)
- [Pollinations image API documentation](https://gen.pollinations.ai/docs#tag/image)
- [Live image and video model catalog](https://gen.pollinations.ai/image/models)
- [Pollinations BYOP device flow](https://github.com/pollinations/pollinations/blob/main/BRING_YOUR_OWN_POLLEN.md#%EF%B8%8F-clis--headless-apps-device-flow)
- [OAuth device authorization standard](https://datatracker.ietf.org/doc/html/rfc8628)
- [GIMP 3 Python plug-in tutorial](https://developer.gimp.org/resource/writing-a-plug-in/tutorial-python/)
- [GIMP 3 API reference](https://developer.gimp.org/api/3.0/)
- [GIMP plug-in distribution guidance](https://developer.gimp.org/resource/distributing-plug-ins/)

### Reward

10 Pollen

voodoohop:
[...]

PR #14261, opened 2026-09-01
feat: add Pollinations image generation to GIMP

tomdacatto:
Fixes #14232

- Adds explicit Connect Account, Disconnect Account, and Generate or Edit actions in GIMP.
- Opens device authorization in the browser, keeps polling cancellable, and moves login, model loading, and generation off GTK's thread.
- Supports new-image and new-layer generation plus selection-aware editing into a new layer.
- Uses platform-native token paths and documents the exact Linux, macOS, and Windows plug-in layout.
- Adds 20 tests covering cancellation, expiry, URL/base64 images, malformed models, network failures, token lifecycle, and a real local HTTP path.

Automated: py_compile, 20/20 tests, and diff checks pass. The host running these changes has no GIMP 3 runtime, so the real GTK/PDB smoke remains an explicit checklist in the README rather than a mocked claim.


voodoohop:
- Thanks for the substantial GIMP submission.\n- #14569 was selected as the smallest complete implementation and passed native GIMP 3 and live API testing.\n- Eligible contributors received one 3-Pollen participation reward on the Quests page.

Issue #14162, opened 2026-08-31
Agent rows: don't show a price or 'free', show capabilities instead


voodoohop:
Agents show "free" / infinite req-per-pollen until the 7-day price average has enough data, so every new agent (e.g. `pollinations-router/floret`) looks free. Prices for agents can't be estimated reliably, so don't show one at all.

- Never render the per-pollen estimate or free chip for agent rows (`model-row.tsx`, `model-table.tsx`, `calculatePerPollen()` in `calculations.ts`).
- Show the agent's input **and** output modalities and capabilities instead (`agent-catalog.ts` currently copies text-to-text from the base model — derive the real ones).
- Make the description readable without hover, desktop and mobile.

Related: #13445, #11910, #12777, #14043

PR #14294, opened 2026-09-01
fix: derive prompt agent catalog metadata

Circuit-Overtime:
- Inherits base-model input modalities for managed prompt agents while reporting chat text output.
- Derives `pollinations_models` and `web_search` capabilities from enabled MCP servers.
- Deduplicates inherited and agent-specific capabilities.
- Adds the Pollinations tools capability and icon to the Models UI.

Floret content-array compatibility was split into #14308.

Addresses #14162.


Circuit-Overtime:
### Agent cost display

- Agent rows intentionally omit both the **Free** chip and the rolling **requests per Pollen** estimate, as requested in #14162.
- An agent run can fan out across different models and tools, so the production value (currently around **250 requests/Pollen** for Floret) is only a blended seven-day observation—not a stable per-request price.
- Users are still charged normally for the downstream work performed by the agent; this change only removes a misleading fixed-cost presentation from the Models UI.

Circuit-Overtime:
### Agent modality and tool icons

The agent cards can present output modalities as compact icons alongside capabilities. For Floret, the intended display after the catalog change is:

- **Outputs:** text, image, video, and audio
- **Capabilities:** web search, sandboxed code/media execution, and built-in Pollinations model tools

Floret does run an internal tool-calling loop, so `pollinations_models` already gives it the Tool icon and `web_search` gives it the Search icon. I would avoid advertising the standard `tool_calling` capability for now: Floret’s public request schema accepts `model`, `messages`, `stream`, and `routing`, but not caller-provided `tools` or `tool_choice`. We should add that capability only if the public endpoint starts accepting client tool definitions.

voodoohop (changes requested):
This PR mixes several separate changes and should be split:

- **Agent pricing:** keep the measured 7-day requests-per-Pollen estimate. It reflects the real total cost of agent runs. Only show `—` when an agent has no measured usage yet. Do not hide the price ledger, variants, balance-access chip, or “Tools — Usage-based” row.
- **Prompt-agent metadata:** keep this as a focused PR. Inheriting base-model input modalities, reporting text output, deriving `pollinations_models`/`web_search` from enabled MCPs, deduplicating capabilities, and adding the tool icon all make sense.
- **Floret compatibility:** put content-array parsing in a separate small PR.

Remove the endpoint-agent special case from this PR. Hardcoding `pollinations-router/floret` as text/image/video/audio makes Play classify it like a direct media model, then adds agent exceptions in three frontend files to undo that.

Endpoint agents should declare `inputModalities` and `outputModalities` in the create/update API, like community models. Default both to text. Floret can then declare its real outputs in its listing instead of being hardcoded in the catalog. Do that in a separate PR, and keep agents routed through chat even when they advertise media output.

Also remove the unused `getModelOutputModalities()`. Output modalities are not rendered; the test explicitly expects no `Output:` label.

Circuit-Overtime:
Refocused this PR following review:

- Restored measured seven-day requests-per-Pollen, the pricing ledger/variants, balance-access chip, and **Tools — Usage-based** row for agents.
- Removed the endpoint-agent/Floret model-ID special case and the unused output-modality helper.
- Removed all Play compatibility changes; content-array parsing now lives independently in #14308.
- The remaining diff is limited to managed prompt-agent metadata and its capability icon support.

Circuit-Overtime:
Thanks for the careful review — the separation makes the scope much clearer. I’ve kept this PR focused on prompt-agent metadata and restored the existing measured pricing UI.

The Floret content-array handling is now isolated in #14308, and the endpoint-agent hardcode and related Play exceptions have been removed from this branch.

voodoohop (approved):
- Re-reviewed the current head; the earlier scope concerns are resolved.
- Keeps measured pricing and the existing model UI unchanged.
- Removes the Floret/endpoint-agent special cases and Play compatibility changes.
- The remaining diff is limited to prompt-agent catalog metadata and the shared Pollinations-tools capability display.
- No blocking findings.

PR #14309, opened 2026-09-01
fix: show zero-Pollen notice on authorization screen

### LINKED ISSUE #14228: [QUEST] Show zero-balance help during BYOP authorization
## Problem

A user can arrive from a BYOP app, sign in, and authorize a new key with zero Pollen. The consent screen shows `0 pollen` and a generic **Top up** link, but does not explain that the app needs Pollen or point to free Quests. The user returns to the app and only discovers the problem when their first request fails.

## Outcome

When the signed-in user's loaded balance is zero, show a small friendly notice on the BYOP consent screen:

> You don't have any Pollen yet. Earn some through Quests or top up before using this app.

- Link **Quests** to `/quests` and **top up** to `/pollen#buy-pollen`.
- Open the links in a new tab so the pending authorization is preserved.
- Keep authorization available.
- Cover both web/PKCE and device authorization, which share this screen.
- Hide the notice for positive or unavailable balances.

No backend or API changes are needed.

### Reward

7

gggff123:
- Shows a friendly notice when the loaded Pollen balance is zero or negative.
- Places the notice before the app details on the shared consent screen.
- Uses shared UI links for Quest and Top up, opening both in new tabs so authorization stays open.
- Keeps authorization available for both PKCE and device flows.

Fixes #14228


Circuit-Overtime:
Thanks for the focused improvement. The notice is correctly gated on a loaded balance of exactly zero, remains hidden for positive or unavailable balances, and keeps authorization available. Using the shared consent screen covers both PKCE and device flows, while opening Quests and Top up in new tabs preserves the pending authorization. This looks ready to merge.

gggff123:
thanks @ElliotEtag

PR #14196, opened 2026-08-31
feat: add polli harness openclaw on|off|status

Marcus-Mok-GH:
Adds an `openclaw` adapter to `polli harness`, mirroring `dsh`.

**`on`** requires the `openclaw` binary (offers the official install page if missing), logs Polli in if needed, mints a dedicated `polli-harness-openclaw` key, and writes:
- `env.vars.POLLI_OPENCLAW_API_KEY`
- `models.providers.pollinations` with the **live model catalog** (`pollinations/*` wildcard — no second hardcoded list), `mode: "merge"`
- `agents.defaults.model.primary` (default `kimi`; override with `--model <id>`), preserving existing fallbacks
- the Polli skill under `~/.openclaw/skills/polli` via `skills.load.extraDirs`

**`status`** reports whether the integration is ready and which model is default.

**`off`** restores the byte-for-byte backup taken before `on`, or strips only the Pollinations-owned provider, model, key, and skill when the config changed since. Existing agents and unrelated config are preserved.

Also retires the competing hand-maintained path: `apps/openclaw/setup-pollinations.sh` becomes a shim over the harness, and `apps/openclaw/index.html` + `README.md` now use the current `env.vars` schema and the `pollinations/*` wildcard instead of a hardcoded model catalog.

- Adds `openclaw` harness adapter + tests (7) in `packages/polli-cli/src/harnesses/`
- Updates `CODING_HARNESSES.md` (Scala Coding Harnesses docs), polli-cli `SKILL.md`/`README.md`
- Converts `apps/openclaw` setup/manual config to the shared `polli harness` path

Fixes #14121

PR #14343, opened 2026-09-02
feat(cli): add OpenClaw harness to polli CLI

Marcus-Mok-GH:
Addresses #14121. Adds the OpenClaw harness to the shared `polli harness` workflow built on the foundation from #14107, using the existing `apps/openclaw` integration as reference.

## What this adds

`polli harness openclaw on|off|status`.

### Quest requirements, met minimally

- **New or existing installs** — `on` merges a `pollinations` provider into `~/.openclaw/openclaw.json` (honors `OPENCLAW_CONFIG_PATH` / `OPENCLAW_STATE_DIR`, tilde included). It works from scratch or on an existing config the same way, with no subprocess onboarding to simulate.
- **Missing OpenClaw → official install offered** — `on` refuses before any login/key/config work with `https://openclaw.ai/install`.
- **Dedicated key** — shared `resolveHarnessKey` mints `polli-harness-openclaw`; a still-valid stored key is reused. Stored once in `env.vars.POLLI_OPENCLAW_API_KEY`, referenced from the provider as `${POLLI_OPENCLAW_API_KEY}` (OpenClaw's documented substitution), so the secret never appears twice.
- **No second hardcoded catalog** — models come from `fetchHarnessModels()` (live `gen.pollinations.ai/v1/models`, tool-calling text only), the same helper as the other adapters.
- **Selectable default** — `--model <id>`, validated against the live catalog, default `kimi`; written to `agents.defaults.model.primary`.
- **Polli skill** — installed as a managed skill at `<state-dir>/skills/polli/SKILL.md` for image/audio/video generation.
- **Works with current OpenClaw** — schema matches `docs/concepts/model-providers.md` (`models.providers.<id>`, `models.mode` only set if unset, `env.vars` + `${VAR}`).
- **Unrelated config preserved** — only the config file and skill are ever written; an explicit existing `models.mode` (e.g. `"replace"`) is never overridden; atomic writes with 0600 permissions.
- **`status`** — reports configured/not and the active default model.
- **`off` removes only Pollinations-owned changes** — byte-for-byte snapshot restore when the files are untouched; otherwise surgical strip of just the pollinations provider, the key, a `pollinations/*` default (a user-chosen one like `openai/gpt-5.6` survives), and the skill.
- **No competing/outdated path** — `apps/openclaw/setup-pollinations.sh` (which hardcoded 9 models and had no revert) becomes a thin shim forwarding to `polli harness openclaw on`; the old `curl | bash -s -- KEY` one-liner keeps working.
- **Scalar docs** — `CODING_HARNESSES.md`, `packages/polli-cli/README.md`, `SKILL.md`, and `apps/openclaw/README.md` document setup, status, model selection, and removal.

## Verification

- `vitest run`: 78/78 pass across the package, including 10 new `openclaw.test.ts` tests (from-scratch write, config preservation, mode-preserve, byte-for-byte restore, strip-with-outside-edits, unchanged-before-on, config/state-dir env overrides, missing credential, missing binary).
- `biome check` clean; `tsup` build succeeds.
- **End-to-end smoke test** with the built CLI against a fake `openclaw` binary and a local mock of the Pollinations API server: `on` minted the key, fetched the live catalog (dropped a non-text model), wrote provider + default + skill, and `status` reported `configured: true, model: kimi`; `off` restored byte-for-byte (`outcome: restored`).
- Not verified against a real OpenClaw install (none available in this environment). Config schema and flags come from the current OpenClaw docs and the previously shipped `setup-pollinations.sh`; `openclaw doctor` will surface any OpenClaw-side drift.

Changed files:
- `packages/polli-cli/src/harnesses/openclaw.ts` (new)
- `packages/polli-cli/src/harnesses/openclaw.test.ts` (new)
- `packages/polli-cli/src/harnesses/index.ts` (register adapter)
- `packages/polli-cli/SKILL.md`, `packages/polli-cli/README.md` (mention the harness)
- `apps/openclaw/setup-pollinations.sh` (thin shim)
- `apps/openclaw/README.md` (rewritten around `polli harness`)
- `CODING_HARNESSES.md` (Scalar docs: OpenClaw section)


voodoohop (changes requested):
- Parse `openclaw.json` as JSON5, carrying over the valid-config handling from #14196; `JSON.parse` rejects comments and trailing commas.
- Honor `OPENCLAW_HOME` using OpenClaw’s path semantics: `OPENCLAW_CONFIG_PATH`, then `OPENCLAW_STATE_DIR`, then `$OPENCLAW_HOME/.openclaw`. #14262 has the useful path coverage here.
- Do not write `models.mode`; omitted mode already defaults to `merge` and avoids leaving a Pollinations-created field behind.
- Keep the fresh-install insight from #14262, but simplify it: for a missing config, run `openclaw setup --baseline` before taking the snapshot, then apply Pollinations. This initializes OpenClaw without making Pollinations part of the restore baseline.
- Run a smoke test against the current real OpenClaw release, following the useful verification demonstrated in #14253.
- Fix the README reference from `models.json` to `openclaw.json`.
- Keep the follow-up focused; the current 78 tests and build pass locally.

voodoohop (approved):
- Required fixes are complete.
- All CI and CodeQL checks pass.
- Verified setup, model discovery, status, and removal with OpenClaw 2026.8.2.

PR #14348, opened 2026-09-02
feat(transforms): add kimiK3Transform and optimizeRequestTransform (Add VSCODE support)

hlohaus:
- Implemented kimiK3Transform to ensure tool messages have resolvable names and echo reasoning content for Kimi K3 compatibility.
- Introduced optimizeRequestTransform to condense system prompts, remove duplicates, strip reasoning echoes, collapse whitespace, trim old turns, and condense tool descriptions to optimize request payloads.
- Updated TransformResult interface to include a saved property for tracking bytes saved during optimization.
- Enhanced type definitions for OpenAI-style prompt token details and chat completion chunks.


hlohaus:
@voodoohop the idea here is using Pollinations for VS Code. I’m already using Pollinations, but it works better with g4f as a proxy first. Thanks to my improvements, I want you to have this too.

PR #14421, opened 2026-09-04
[Fix] slow/failing model catalog load on enter.pollinations.ai

### LINKED ISSUE #14301: enter.pollinations.ai slow load and model list request failures
## Summary
Multiple users report that [enter.pollinations.ai](https://enter.pollinations.ai) takes a very long time to load, and the model list request is failing.

## Details
The dashboard frontend fetches the model catalog from `gen.pollinations.ai/models` via `fetchModelCatalog()` in `enter.pollinations.ai/frontend/src/components/models/model-catalog.ts`. That fetch has a 15-second abort timeout (`AbortSignal.timeout(15_000)`) and uses `cache: "no-store"`.

The `/models` endpoint currently returns a **large payload** (~250+ entries including all community models, with full pricing metadata, aliases, modality arrays, etc.). Combined with `no-store`, every dashboard visit forces a fresh, large fetch that can approach or exceed the 15s timeout on slower connections, causing the models tab/permission picker to fail or hang.

Model status (`/v1/models/status`) over the last 60 min confirms the gateway itself is healthy — latency issues appear to be payload-size + `no-store` related on the dashboard side, not an upstream outage.

## Reported by
`r3ap3redit` on Discord

## Suggested investigation
- Payload size of `/models` (community models have grown significantly)
- Whether `cache: "no-store"` is necessary for the catalog fetch, or if a short cache TTL would help
- Whether the catalog could be split/paginated or lazy-loaded (e.g. community models fetched separately/on-demand)
- 15s timeout adequacy given current payload size

## Repro
1. Visit https://enter.pollinations.ai
2. Navigate to the Models tab (or any flow that loads the model picker)
3. Observe long load time / request failure for the model list

---
**Author:** `R3AP3R editz (@r3ap3redit)` (UID: `1012739996773523527`)
**Source:** [View on Discord](https://discord.com/channels/885844321461485618/1544389803309797396/1544390020469756055)

gggff123:
Splits the /models fetch into official (?community=false) and community (?community=true) requests, removes cache: "no-store" in favor of default caching (with an explicit reload on refresh), and gives the larger community response its own longer timeout with graceful degradation on failure.

Fixes the dashboard hang/failure reported by r3ap3redit on Discord: enter.pollinations.ai was fetching the full ~250+ entry catalog with no-store on every visit, which could approach or exceed the 15s abort timeout on slower connections.
Closes #14301

### DIFF
--- a/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
+++ b/enter.pollinations.ai/frontend/src/components/models/model-catalog.ts
@@ -90,10 +90,23 @@ export function mergeModelCatalogs(
     return [...modelsById.values()];
 }
 
-async function fetchCatalog(url: string): Promise<ApiModelInfo[]> {
+// The official catalog is small and required — if it fails, the whole
+// catalog fetch should fail loudly. Its own request should complete quickly.
+const OFFICIAL_CATALOG_TIMEOUT_MS = 10_000;
+
+// Community models make up the bulk of the payload (~250+ entries with full
+// pricing metadata). They're fetched as a separate request so a slow/large
+// community response can't hold up rendering official models, and so it can
+// be given a more generous timeout without penalizing the common case.
+const COMMUNITY_CATALOG_TIMEOUT_MS = 30_000;
+
+async function fetchCatalog(
+    url: string,
+    { timeoutMs, cache }: { timeoutMs: number; cache: RequestCache },
+): Promise<ApiModelInfo[]> {
     const response = await fetch(url, {
-        cache: "no-store",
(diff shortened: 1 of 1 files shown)


gggff123:
@voodoohop please review when you are free.

voodoohop:
- Thanks @gggff123 for investigating! We're closing this in favor of #14515, which addresses dashboard loading by removing billing and earnings waits.
- Added a **2 Pollen bonus quest** for your effort 🌸 Claim it at https://enter.pollinations.ai/quests.

PR #14470, opened 2026-09-05
feat: unify account menus and separate dashboard login

ElliotEtag:
- Moves AccountMenu from a domain module into an SDK-free composition and reuses it in Enter, AppUserMenu, Economics and KPI. Preserves caller-owned account links and sign-out actions; distinguishes app allowance from wallet balance.
- Adds identity-only operations login with PKCE, trusted dashboard clients, Better Auth administrator checks at token exchange, and independent HttpOnly sessions. Fixes the immediate re-login loop after dashboard sign-out. Addresses #14089 and builds on #14452 without changing those PRs.
- The react.pollinations.ai showcase changes moved to #14473, which is stacked on this branch.
- Validation: 74 tests passed across OAuth, auth package, UI, Economics and deployment discovery; relevant type checks, Economics/KPI builds and desktop/mobile browser checks passed. Local commit hook reported missing lefthook; Biome was run separately and passed.
- Draft rollout gate: no secret values or secret-copy scripts changed; no deployment performed. Dashboard session-secret provisioning requires separate scoped approval. Live OAuth and Grafana runtime verification remain pending; the Grafana dry run bundled successfully but stalled afterward and was stopped.

PR #14255, opened 2026-09-01
refactor: authorize admin dashboards at backend

voodoohop:
- Adds independent admin-only Pollinations OAuth sessions to KPI, Economics and Observability. Sign-out affects only the current app; requests identity only, with no API-key delegation.
- Reuses the shared sign-in page, official provider button, Fraunces headings and account menu. Completes fresh GitHub sign-in and shows recoverable login errors. Registers both dashboard domain families.
- Rechecks admin status and token validity within a 60-second signed-cookie window, coalesces concurrent checks, and treats Enter outages as retryable 503s without clearing sessions. Encrypts the identity token in the HttpOnly cookie. Protects Economics registry assets and Grafana through the app gateway without a separate Grafana login session.
- Restores existing KPI GitHub credentials, Economics staging setup, deployment contract validation and secret synchronization paths. Verified 28 shared auth tests, 19 Enter integration tests, KPI/staging regressions, type checks, all dashboard builds, private-bundle isolation and Economics deployment dry run.
- Before rollout: separately approve/provision production session secrets, verify data readers, deploy Enter first, then verify all apps. Existing app cookies require sign-in again. No production deployment performed; Grafana container image validation still requires Docker.

Replaces #14089 and #14252. Consolidates #14565.

PR #14473, opened 2026-09-05
feat: add account and auth examples to react.pollinations.ai showcase

ElliotEtag:
Split out of #14470 so the dashboard login change can be reviewed on its own. Depends on the AccountMenu composition from #14470, so this branch includes its commits; review only the last commit here. Targets `main` because the PR checks workflow only runs for PRs against `main`. Once #14470 merges, this gets rebased onto `main` and the diff collapses to the showcase files.

- Adds account-context examples to Modules: Enter owner session, delegated app connection and identity-only operations session, all with fictional data and no live actions.
- Adds wallet display and auth presentation examples, and a bare `AccountMenu` example under Compositions.
- Adds missing primitives to the reference (Drawer, TableDisclosureButton, AppHeader, FieldStack, EditableCombobox, remaining icons) and wraps wide module tables in a horizontal ScrollArea.
- Removes the hardcoded App Earnings row and the request-count copy from the live AppUserMenu section; missing key data now shows "Unavailable".
- Points the `@pollinations/ui` README at the showcase examples.

Validation: `tsc --noEmit` and `vite build` for `apps/react` pass on this tree.


voodoohop:
- #14470 is closed; the shared AccountMenu changes are preserved in #14478.
- This showcase PR can stay open, but needs its showcase commit reapplied on top of #14478/main without the custom OAuth commits from #14470.
- Dashboard authentication continues in #14255 using the official Better Auth provider.

ElliotEtag:
- Closing: the shared AccountMenu lands via #14478 instead. The showcase changes will be redone on top of it as a full react.pollinations.ai refactor.

PR #14515, opened 2026-09-06
fix: unblock dashboard loading from billing and earnings

voodoohop:
- Removes Stripe billing from the shared dashboard loader; only the Pollen page fetches it. Previously every signed-in dashboard page waited for Stripe before rendering.
- Lets weekly earnings load after the dashboard renders using the router's existing `Await`. Wallet balances remain visible while earnings are pending or unavailable.
- Shows a loading message while Pollen waits for billing. Keeps the change to two route files, with no new dependencies or caching logic.
- Addresses #14000; related to #14421, which changes catalog fetching. This PR removes dashboard blockers before Models renders; it does not replace #14421. The reported long delay was not reproduced.
- Validation: production frontend build, TypeScript, and Biome pass. Actual loader, router, and rendering checks cover deferred/failed earnings, signed-out loading, Pollen-only billing and its pending state, and pack navigation preserving billing data.

### DIFF
--- a/enter.pollinations.ai/frontend/src/routes/_dashboard.pollen.tsx
+++ b/enter.pollinations.ai/frontend/src/routes/_dashboard.pollen.tsx
@@ -36,9 +36,9 @@ export const Route = createFileRoute("/_dashboard/pollen")({
     loader: () =>
         apiClient.stripe.billing.$get().then((r) => (r.ok ? r.json() : null)),
     pendingComponent: () => (
-        <p role="status" className="text-theme-text-muted">
+        <output className="text-theme-text-muted">
             Loading billing details…
-        </p>
+        </output>
     ),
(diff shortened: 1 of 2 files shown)


voodoohop:
- Reviewed routing, authentication, deferred rendering, failure handling, and wallet/billing state. No blocking defects found.
- Claude Fable independently reviewed the code. Addressed its useful suggestions: Pollen now shows a loading message while Stripe is pending, and wallet components receive only balance fields plus earnings.
- Kept the change to two files; no new helpers, dependencies, or caching layer. Left unrelated slider-history behavior out of scope.
- Rechecked the production build, TypeScript, Biome, actual router navigation with stalled billing/earnings, and React's immediate balance fallback. Pack changes preserve billing data and do not repeat Stripe requests.

Issue #14535, opened 2026-09-07
[QUEST] Filter model catalogs by source and reliability


voodoohop:
### Goal

Filter model listings (official-only or reliable models) and expose health information.

### Research so far

- Replace `community` filtering if research supports a better interface; compatibility isn't required. Reuse Enter's terminology and health data.
- [Open WebUI](https://github.com/open-webui/open-webui/blob/0a7c15832fb30b1903753e83f81dc7d27e5b0944/backend/open_webui/routers/openai.py#L88-L130), [LibreChat](https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/custom_endpoint#headers), and [Cline](https://github.com/cline/cline/blob/c21b17255b228e88a1518c18a73a473ee5876362/apps/vscode/src/core/controller/models/refreshOpenAiModels.ts#L16-L45) append `/models`: queries embedded in base URLs break. Custom headers are supported; LibreChat requires administrator-configured connections. Source-checked, not end-to-end tested.
- [OpenRouter](https://openrouter.ai/docs/api/api-reference/models/list-all-models-and-their-properties) provides simple query filters; [AI Horde](https://aihorde.net/api/swagger.json) exposes live serving capacity. Distinguish [Fireworks](https://docs.fireworks.ai/api-reference/list-models), [fal](https://fal.ai/docs/platform-apis/v1/models), and [Chutes](https://chutes.ai/docs/api-reference/chutes) management catalogs from OpenAI-compatible discovery.

### Deliver

- Validate discovery and generation in 2–3 clients. Start with query parameters and an optional header; add base-path presets only if needed. No generic filtering framework.
- Support `/v1/models` and relevant model lists; reuse dashboard semantics.
- Expose minimal health metadata: success rate, sample count, window, freshness. Count fallback rescues as successes; exclude caller-side failures. Missing data means unknown. Experimental status is separate from reliability.
- Preserve access rules and default listings; filtering discovery must not change generation permissions. Include client examples, source API docs, and focused tests.

### How this quest works

Submit a focused PR starting with `Fixes #14535`. The selected solution earns the reward when merged.

### Reward

10

davealan74:
Opened #14536.

[...]

PR #14536, opened 2026-09-07
feat: expose model reliability and availability in catalogs

davealan74:
Fixes #14535

Adds measured reliability/availability metadata to model catalog responses.

- `shared/registry/model-health.ts` (new): computes `status` (`healthy`/`degraded`/`unavailable`/`unknown`), `success_rate`, `sample_size`, `window_minutes`, `checked_at`, `stale` from the existing `model_health` Tinybird pipe rows. Sample is `2xx / (2xx + 5xx)`; 4xx (bad auth, no balance, invalid request, rate limits) is excluded from the denominator entirely rather than counted against the model. Below a sample-size threshold or with no data, status is `"unknown"`, never `"healthy"`.
- `gen.pollinations.ai/src/routes/model-status.ts`: extracted the existing fetch/cache logic into `getModelHealthSnapshot()` so catalog enrichment reuses the same 60s-cached Tinybird call the `/v1/models/status` route already makes — no per-model fetches, no new endpoint. Also validates the Tinybird response against the existing zod schema instead of an unchecked cast.
- `gen.pollinations.ai/src/model-registry.ts`: attaches health to every registry entry (official + community) from one shared snapshot per rebuild (24h window, so lower-traffic models stay populated). A Tinybird outage never blocks model listing — it falls back to `"unknown"`.
- `shared/registry/model-info.ts`, `shared/schemas/openai.ts`: add the optional `health` field to `ModelInfoSchema`/`OpenAIModelSchema`, so it's present on `/v1/models`, `/v1/models/:model`, `/models`, and every category list (`/image/models`, `/text/models`, etc.) via the shared mapper — no per-endpoint changes needed.
- `gen.pollinations.ai/src/routes/proxy.ts`: passes `health` through in the OpenAI-compatible mapper.

No changes to response shape/ordering/filtering/visibility beyond the new optional field; no new endpoints or Tinybird pipe changes.

Verified: `tsc --noEmit` clean for both `gen.pollinations.ai` and `enter.pollinations.ai` workspaces (shared schema change), `biome check` clean. Added focused tests in `test/model-registry.test.ts` (status thresholds, low-sample handling, unavailable-source fallback) and `test/models-retrieve.test.ts` (health present and consistent across list/retrieve/`/models`). Core classification logic double-checked directly with `tsx`. The Cloudflare Workers vitest pool (`workerd`) could not run in this sandbox (glibc too old for the pinned binary) — untouched pre-existing test files fail the same way, confirming it's an environment limitation, not a regression; CI should run the full suite.

### DIFF
--- a/gen.pollinations.ai/src/model-registry.ts
+++ b/gen.pollinations.ai/src/model-registry.ts
@@ -6,6 +6,12 @@ import { DEFAULT_AUDIO_MODEL } from "@shared/registry/audio.ts";
 import { DEFAULT_EMBEDDING_MODEL } from "@shared/registry/embeddings.ts";
 import { DEFAULT_IMAGE_MODEL } from "@shared/registry/image.ts";
 import {
+    buildModelHealthIndex,
+    type ModelHealth,
+    type ModelHealthRow,
+    unknownModelHealth,
+} from "@shared/registry/model-health.ts";
+import {
     type ModelInfo,
     modelInfoFromDefinition,
 } from "@shared/registry/model-info.ts";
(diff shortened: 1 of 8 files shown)


voodoohop:
Thanks for the health-metadata work. That part has shipped, while this PR does not add the requested source and reliability filters. Closing as superseded.

PR #14541, opened 2026-09-07
fix: preserve Floret routing with canonical model IDs

voodoohop:
- Accepts catalog aliases when validating explicit model choices; valid old IDs otherwise become "unknown model" after the rename.
- Keeps Veo reference-image duration handling for publisher-prefixed IDs, avoiding invalid 5/7-second requests.
- Leaves requested IDs, model priorities, and default choices unchanged; existing capability validation still applies.
- Adds four regression tests covering old/new IDs, invalid choices, exact-ID precedence, and video duration behavior.

Validation: focused `unittest` suite passes; Ruff formatting and routing/test lint pass. Existing `gen.py` lint warnings (BLE001, PIE810) are unchanged.

Addresses #13076

### DIFF
--- a/apps/floret/src/floret/routing.py
+++ b/apps/floret/src/floret/routing.py
@@ -216,6 +216,15 @@ async def validate_routing(value: RoutingInput | None) -> RoutingPreferences:
     for field, model in explicit.items():
         meta = catalog.get(model)
         if meta is None:
+            meta = next(
+                (
+                    entry
+                    for entry in catalog.values()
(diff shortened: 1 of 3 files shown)


ElliotEtag:
Changes incorporated into #13076, preserving the original commits. Follow-up compatibility fixes and verification are documented at https://github.com/pollinations/pollinations/pull/13076#issuecomment-5570762069. Closing this standalone PR; no separate merge into main.

PR #14560, opened 2026-09-08
docs: reduce repository transfer size without changing runtime

murderszn:
## Summary

- document shallow and partial clone options for consumers importing the repository
- add a copy/paste clone helper using depth and blob filtering
- link the guidance from the README

## Why this approach

The repository currently has a large commit history, while application behavior depends on the checked-out tree rather than historical commits. A history rewrite would invalidate existing clones, forks, commit links, and external automation, so this PR keeps history intact and gives importers a safe way to avoid downloading it.

The documented commands preserve a normal working tree and can be expanded with `git fetch --unshallow` when history is later needed. No runtime source, assets, fixtures, or lockfiles are removed.

## Validation

- `git diff --check`
- `bash -n scripts/clone-shallow.sh`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@
 
 </div>
 
+> [!TIP]
+> Need the current source without downloading repository history? See [Keeping local clones small](docs/REPOSITORY_SIZE.md) for shallow and partial-clone commands.
+
 > [!IMPORTANT]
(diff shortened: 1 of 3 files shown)


murderszn:
@voodoohop seems like it would work as a short term solution.

PR #14567, opened 2026-09-08
fix: repair Polli agent registration

ElliotEtag:
- Addresses the [Polli workflow failure](https://github.com/pollinations/pollinations/actions/runs/34139320216): its API requires delegated `ag_` credentials, while its production `proxy` listing sends the saved upstream bearer. Adds a guarded manual SQL repair to convert that exact listing to `endpoint_agent`.
- Preserves `pollinations-router/polli`, alias `community/pollinations-router/polli`, provider `community`, and `https://polli.pollinations.ai/v1/chat/completions` with upstream model `polli`. No Pollinations GPU; no equivalent fallback identified in the registry, so the existing absence of a gateway fallback is preserved.
- Changes billing from wrapper rates of 1.40/4.40 Pollen per million input/output tokens to downstream usage only, with wrapper multiplier `1` and `paidOnly: false`. Applying the repair removes the obsolete encrypted upstream bearer from the listing.
- Keeps production application behind explicit billing/credential-removal approval. Includes backup, rollback, and live verification instructions; preserves hidden state until recovery is verified. Merging this PR does not run the repair.
- Validation: SQLite conversion, idempotence, timestamp units, rate/hidden-state preservation, and 11 mismatch/pending-state guards passed. Biome skips SQL; whitespace check passed. Existing authentication tests could not start because the installed Workers runtime rejects a compatibility flag. Production remains unchanged.


Itachi-1824 (approved):
Reviewed the repair and runtime paths. Looks good.

PR #14570, opened 2026-09-08
feat: filter model catalogs by source and reliability

harisrehman155:
Fixes #14535

- Adds source/reliability query filters and equivalent headers to `/v1/models` and category catalogs, after existing visibility/key/balance checks. Default discovery and generation access are unchanged.
- Reuses cached health data, exposes rates/counts/window/freshness, and counts final fallback successes. Image-provider 4xx failures are included; caller 4xx and intermediate attempts are excluded. Experimental status stays independent.
- Preserves raw `/v1/models/status` passthrough and validates counts only for catalog calculations. Missing image-provider attribution is unknown and cannot qualify as reliable.
- Validation: 95 gateway/catalog/status/permission tests pass with `--testTimeout=20000` (an existing audio test exceeded the default 5s). The OpenAI JS + AI SDK discovery/completion test, TypeScript and Biome also pass. Desktop clients and live billed providers were not end-to-end tested.
- Rollout: deploy the updated `model_health` Tinybird pipe to staging, verify, then production before expecting image reliability classifications. This PR changes the pipe source; no Tinybird deployment was performed. Image reliability fails closed until the new counter is available. Source API docs and client examples are included.


Circuit-Overtime:
Thank you for the careful work here, especially the permission-first filtering and raw status preservation.

> Superseded by #14579: it exposes health on default listings and supports both query/header filters without requiring a Tinybird schema rollout or legacy-filter compatibility.

Two technical differences make that implementation safer to merge: this branch only attaches health when `reliability` is present, while the quest asks model listings to expose health; and `provider_errors_4xx` relies on an `Image provider error:` prefix that the current generation path does not emit, so the added counter would not classify those failures as intended. Keeping `community` compatibility and `source=all` also adds public surface the quest explicitly says is unnecessary.

I rebased this branch onto current `main`, but I would not merge it over #14579.

voodoohop:
- Thanks @harisrehman155! We adapted your OpenAI/AI SDK integration test and bounded health-fetch approach into #14553, with co-author credit.
- Recorded a 3-Pollen contribution bonus, available to claim on your Quests page.

voodoohop:
Thanks again for the careful tests and review. We incorporated your client-coverage idea and recorded the 3-Pollen bonus. Closing this superseded branch.

PR #14588, opened 2026-09-08
feat: filter model catalogs by source and reliability

PSFREQUENCY:
Fixes #14535

Adds `source`, `reliability` and `health` to every model catalog, plus header equivalents, and exposes per-model health derived from the same Tinybird pipe that already backs `/v1/models/status`.

## Filters

| Parameter | Values | Behaviour |
|---|---|---|
| `source` | `all` (default), `official`, `community` | Who publishes the model |
| `reliability` | `all` (default), `reliable` | Only models whose recent health clears both thresholds |
| `health` | `true`/`false`/`1`/`0` | Attach the summary. Implied by `reliability=reliable` |

`community=` still works and maps onto `source`; an explicit `source` wins. An unknown value is a **400** rather than a silent full listing.

```bash
curl "https://gen.pollinations.ai/models?source=official"
curl "https://gen.pollinations.ai/v1/models?reliability=reliable"
curl "https://gen.pollinations.ai/image/models?health=true"
```

## Header equivalents

Open WebUI, LibreChat and Cline all build the catalog URL by appending `/models` to a configured base URL, which discards any query string. All three support custom headers, so every filter is also readable from `X-Pollinations-Model-Source`, `X-Pollinations-Model-Reliability` and `X-Pollinations-Model-Health`. A query parameter wins over a header, so an explicit URL beats a client-wide default.

- **Open WebUI** — *Settings → Connections → OpenAI API*, base URL `https://gen.pollinations.ai/v1`, add `X-Pollinations-Model-Source: official` to that connection's custom headers.
- **LibreChat** — `headers: { X-Pollinations-Model-Source: "official" }` under the custom endpoint in `librechat.yaml`.
- **Cline** — *OpenAI Compatible* provider, same header in the custom-headers section.

## Health

`status` · `success_rate` · `sample_size` · `window_minutes` · `checked_at` · `stale`

- A request the fallback layer rescued still returned a usable result, so it **counts as a success**.
- Caller-side `4xx` failures are **excluded** from both the rate and the sample — they say nothing about the model.
- Rows are summed per model **before** any rate is taken, so a quiet provider cannot outvote a busy one.
- A model with no traffic in the window is `unknown`, and `reliable` excludes it. `reliable` means "we have recently seen this work", not "nothing has gone wrong yet".
- `reliable` needs a high success rate **and** a minimum sample, so a model that served three requests successfully is not promoted over one with thousands.
- Experimental/preview status is separate from health and unaffected by these filters.

## Scope

Filtering is **discovery only**. Permission and balance filtering still runs first and is untouched, and a test asserts every filtered listing is a subset of the unfiltered one. No generic filtering framework — three named parameters and their header aliases.

## Refactor included

`model-status.ts` and the catalogs now share `src/model-health.ts`, so the raw rows and the summary can never disagree about whether a model is healthy. That route drops 134 lines.

The shared fetch gained a **2.5s timeout**: a slow monitoring backend must not stall a model listing. This also fixed real flakiness — the docs suite was intermittently failing under a shared time budget before the timeout was added.

## Tests

20 new cases in `test/model-catalog-filters.test.ts`:

- health arithmetic — fallback rescues, 4xx exclusion, cross-provider summing, both thresholds, empty windows
- filter resolution — query/header/deprecated-`community` precedence, `reliable` implying `health`
- catalogs — `official` and `community` partition the list, header matches query, `health` shape on every model, `reliable` never returns unknown, filtered ⊆ unfiltered, `/v1/models` and category lists, 400 on bad values

**185 pass** across `model-catalog-filters`, `docs`, `model-permissions`, `models-retrieve` and `community-endpoints`. Typecheck clean, biome clean.

Docs updated in `APIDOCS.md` and `gen.pollinations.ai/src/docs/models.md` (the served page is generated from the same section, so the two cannot drift).

### DIFF
--- a/APIDOCS.md
+++ b/APIDOCS.md
@@ -1182,7 +1182,7 @@ Owners can publish their own embedding backend as a community endpoint. A commun
 
 #### `GET` `/embeddings/models` — List Embedding Models
 
-Returns available embedding models with pricing, capabilities, and supported input modalities. When authenticated: models are filtered by API key permissions, and `paid_only` models are hidden if the account has no paid balance. Pass `?community=false` to exclude community models or `?community=true` to return only community models.
+Returns available embedding models with pricing, capabilities, and supported input modalities. When authenticated: models are filtered by API key permissions, and `paid_only` models are hidden if the account has no paid balance. Pass `?source=official` to exclude community models or `?source=community` to return only community models.
 
 ⚙️ **Parameters**
 
@@ -1197,7 +1197,7 @@ Returns available embedding models with pricing, capabilities, and supported inp
 💻 **Example**
 
 ```bash
-curl "https://gen.pollinations.ai/embeddings/models?community=0" \
+curl "https://gen.pollinations.ai/embeddings/models?source=official" \
   -H "Authorization: Bearer $POLLINATIONS_KEY"
 ```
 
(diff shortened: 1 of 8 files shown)


voodoohop:
Thanks for the comprehensive submission. We selected the smaller #14553 approach; this branch adds extra options and changes generated APIDOCS.md. Closing the superseded PR.

PR #14608, opened 2026-09-08
fix: show image edit test failures in the model dashboard

voodoohop:
- Shows the actual edit-test failure instead of claiming editing is unsupported.
- Keeps successful generation and pricing detection usable when editing fails; preserves existing modality declarations.
- Reuses the current dashboard message UI and documents the optional imageEditError response field.
- Verified: 38 endpoint service tests, Enter typecheck, Biome, and regression cases for HTTP errors, timeouts, and invalid image responses.


sharktide:
Good idea

PR #14617, opened 2026-09-09
fix: bill community videos from reported usage

voodoohop:
## Changes
- Allow omitted video duration when the provider reports generated seconds in `usage.duration`.
- Bill reported duration first; fall back to requested duration only when usage is missing. Invalid reported usage or neither duration returns an error.
- Share the reader between registration and generation; update publisher and Scalar docs.

## Verification
187 focused tests pass; Gen and Enter typechecks pass. Coverage includes reported-duration priority, requested-duration fallback, all three media routes, failures without charges, and disconnect/rejoin/cache with one billed execution.

## Status
Directly tested `NamanSoni78/Seedance-2.5` without duration: HTTP 200 with a video URL but no usage. Existing calls with duration remain supported. The publisher still needs to report usage for duration-less calls from the media-to-Responses/Chat adapter. Draft pending that end-to-end check. No catalog pricing or routing changes.


NamanSoni78:
Hey @voodoohop ,

I have updated the endpoint to always report generated seconds in `usage.duration`. 

The endpoint now returns `usage: { "duration": <seconds> }` alongside `data` for all calls:
- When `duration` is omitted, it generates and bills the default 6 seconds.
- When `duration` is explicitly requested, it reflects the generated duration.

Here is the response from a direct test without passing a duration:

```json
{
  "created": 1788945734,
  "model": "Seedance-2.5",
  "data": [
    {
      "url": "https://fragrant-morning-b0a6.07mrbeastofficial.workers.dev/view/eyJwIjoiYSBjaW5lbWF........mp4"
    }
  ],
  "usage": {
    "duration": 6
  }
}
```

The endpoint is ready for re-testing. Let me know if everything looks good on your end!

Thanks!
EDIT: Please test it tomorrow, i have updated api token (form my model interface), so that change is in queue...

Issue #14599, opened 2026-09-08
[QUEST] Expose supported parameters and defaults in model listings


voodoohop:
### Goal

- Let apps discover which controls a model supports and their defaults, so they can show useful settings instead of guessing.

### Expected outcome

- Model listings expose `supported_parameters` and `default_parameters`, backed by the model registry, including `/models`, `/text/models`, `/v1/models`, and individual model lookup.
- Start with at least three verified official Chat models with different capabilities. Describe what works through Pollinations, not just what an upstream provider advertises. Don't list controls we silently ignore or invent unknown defaults; document conditional support.
- Keep metadata consistent across listings and aliases. Don't imply Chat metadata describes the native Responses API.
- This is information only: don't change generation behavior, billing, or access rules. Include focused tests and API documentation. Prefer the simplest implementation using existing patterns.

### References

- [OpenRouter model fields](https://openrouter.ai/docs/guides/overview/models#model-object-schema)
- [Per-provider parameter support](https://openrouter.ai/docs/api/api-reference/endpoints/list-endpoints)
- [Live model listing](https://openrouter.ai/api/v1/models)

### How this quest works

- No assignment needed; multiple solutions welcome. Submit a PR against `main` starting with `Fixes #14599`. We select for correctness, clarity, and minimal code. The selected merged PR earns the reward.

### Reward

10

afanasevmylife:
Quest completed. PR #14630 adds `supported_parameters` and `default_parameters` to the model registry and exposes them in all model listing endpoints.

Investigated per-model through the generation pipeline (requestUtils.ts, generateTextPortkey.ts, proxy.ts) - not copied from upstream docs.

3 Chat models documented with different capability sets:
- `gpt-5.4-nano`: 18 standard params
- `gpt-5.4`: 19 params including `reasoning_effort`
- `deepseek-v4-flash`: 15 params, no `logprobs`

Ready for review.

NamanSoni78:
OPENED PR #14636 
Fixes #14599 

- Added `supportedParameters` and `defaultParameters` fields to `ModelDefinition` in the model registry (`shared/registry/registry.ts`).
- Updated `ModelInfoSchema` (`shared/registry/model-info.ts`) and `OpenAIModelSchema` (`shared/schemas/openai.ts`) to expose `supported_parameters` and `default_parameters`.
- Implemented parameter derivation in `modelInfoFromDefinition` to provide sensible category-based defaults when omitted.
[...]

PR #14625, opened 2026-09-09
feat(models): expose supported and default chat parameters in model listings

xiaotian1171:
## Summary

Model discovery now exposes **`supported_parameters`** and **`default_parameters`** for official Chat models whose support is verified against the Pollinations route — so apps can show real settings instead of guessing.

- New `shared/registry/parameter-support.ts`: a static, route-level truth table. Entries record which `/v1/chat/completions` controls survive the model's transform chain (cross-checked against `availableModels.ts` and `configs/modelConfigs.ts`), plus gateway-applied defaults.
- Wired into `/models` + `/text/models` via `ModelInfoSchema` and into `/v1/models` (list + retrieve) via the shared OpenAI mapper, so metadata is consistent across listings and aliases resolve identically.
- Three verified official Chat models with genuinely different routes:
  - `openai/gpt-5.4` (Azure): sampling controls (`temperature`, `top_p`, penalties, `seed`) are accepted by the request schema but silently dropped by the route transform — **not listed**, documenting conditional/ignored support honestly.
  - `openai/gpt-oss-20b` (OVHcloud): full standard sampling set passes through; defaults `temperature=1.0` / `top_p=1.0` per the OpenAI model card.
  - `anthropic/claude-sonnet-4.6` (Bedrock): `temperature` and `top_p` are conditionally supported (mutually exclusive, temperature wins); no seed/penalty equivalents exist, so they stay unlisted.
- Both fields describe the Pollinations Chat route only, never the native `/v1/responses` API. Models without a verified entry (including all community models) omit both fields — unknown support is never invented.
- Documentation section in `models.md` covering scope, verified models, and the mutual-exclusion caveat; 4 new focused tests (rich list, text list, OpenAI list + alias equivalence, omission for unverified models).

Information-only change: no generation behavior, billing, or access rules touched.

## Testing

- `tsc --noEmit` clean; `docs:generate` + `docs:validate` pass.
- Full `vitest run`: 1984 passed / 1 flaky pre-existing timeout (`community-endpoints.test.ts` account-API fixture, re-run in isolation → 142/142 pass).
- New `test/model-parameters.test.ts` covers: Azure model omits silently-dropped sampling controls, gpt-oss passthrough + model-card defaults, Claude conditional sampling, identical metadata across `/v1/models`, alias retrieve, and full omission for unverified models.

Fixes #14599

### DIFF
--- a/gen.pollinations.ai/src/docs/models.md
+++ b/gen.pollinations.ai/src/docs/models.md
@@ -42,6 +42,36 @@ Supported media models also advertise both endpoints and return generated-file
 links as assistant text. Reference-required models return their normal missing-input
 error; use their native endpoint until attachments are supported here.
 
+### Chat parameter support
+
+`/models`, `/text/models`, and the OpenAI-compatible `/v1/models` endpoints
+expose two optional fields on official Chat models whose support is verified
+against the Pollinations Chat route:
+
+| Field | Meaning |
+|-------|---------|
+| `supported_parameters` | Controls accepted and forwarded by `/v1/chat/completions` for this model |
+| `default_parameters` | Values Pollinations applies when a Chat caller omits the control |
+
+Verified models today: `openai/gpt-5.4`, `openai/gpt-oss-20b`, and
+`anthropic/claude-sonnet-4.6`. Their routes differ (Azure OpenAI, OVHcloud,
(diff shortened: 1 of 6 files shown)


voodoohop:

PR #14623, opened 2026-09-09
Fixes #14599 — expose supported/default parameters in model listings

aikhusus2025-ctrl:
Expose `supported_parameters` and `default_parameters` in model listings (registry-backed).

- Add `ChatParameter`/`ChatDefaultParameters` types + schema in the shared registry, and carry both fields through to `/models`, `/text/models`, `/v1/models` and `/v1/models/:model`.
- Document verified parameter metadata for 4 official Chat models with differing capabilities: `openai/gpt-5.4-nano`, `anthropic/claude-sonnet-4.6`, `google/gemini-3-flash-preview`, `perplexity/sonar`. Only parameters that actually work through Pollinations are listed (gateway-stripped ones are omitted; `condition` documents provider restrictions). Chat fields describe Chat-completion behavior, not the native Responses API.
- Update the OpenAI model schema, SDK types, API docs (`models.md`) and the docs-samples response.
- Add focused tests (13) covering registry fields, endpoint exposure, and consistency between list and retrieve.

Fixes #14599, /v1/models/:model, + tes (13 test) dan update docs.

PR #14629, opened 2026-09-09
feat: expose supported/default Chat parameters in model listings

Marcus-Mok-GH:
- Adds `supported_parameters` and `default_parameters` to text model listings (`/v1/models/*`, registry-driven) so apps can discover which Chat-completions controls a model accepts and their gateway-applied defaults.
- Exposes on the shared registry `ModelDefinition`, the `ModelInfoSchema`/`OpenAIModelSchema`, the `/models` proxy route, and the SDK `ModelInfo` type. Only text-category models surface these fields.
- Documents accuracy rule: a parameter is listed only if it actually survives the model's gateway transform pipeline (`availableModels.ts`) and reaches the provider.
- Lists `openai/gpt-5.4-nano`, `x-ai/grok-4.6`, and `anthropic/claude-sonnet-5` (with `defaultParameters: { max_tokens: 64000 }`).
- Corrects competing PR #14623: `temperature` and other sampling controls are stripped by `omitOpenAISampling`/`omitClaudeSampling`, not "supported" or "locked to 1" — regression-guarded in tests.
- Adds `chatParameters.test.ts` covering registry round-trips, non-text absence, and the transform-pipeline regression.

Fixes #14599

### DIFF
--- a/gen.pollinations.ai/src/docs/models.md
+++ b/gen.pollinations.ai/src/docs/models.md
@@ -42,6 +42,17 @@ Supported media models also advertise both endpoints and return generated-file
 links as assistant text. Reference-required models return their normal missing-input
 error; use their native endpoint until attachments are supported here.
 
+`supported_parameters` and `default_parameters` let apps discover which
+Chat-completions controls a text model accepts and their defaults, backed by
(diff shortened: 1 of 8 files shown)


voodoohop:

PR #14638, opened 2026-09-09
feat: expose supported_parameters and default_parameters in model listings

afanasevmylife:
Closes #14599.

Adds `supported_parameters` and `default_parameters` to model registry and exposes them in `/v1/models`, `/text/models`, `/models`, and individual model lookup.

- `supportedParameters` / `defaultParameters` added to `ModelDefinition`
- Propagated through `ModelInfo`, OpenAI schema, and proxy endpoints
- Documented 3 Chat models with different capabilities:
  * `gpt-5.4-nano`: standard chat + tools (18 params)
  * `gpt-5.4`: reasoning + tools (19 params, includes `reasoning_effort`)
  * `deepseek-v4-flash`: reasoning, limited params (15 params, no `logprobs`)

Investigated per-model through the generation pipeline, not copied from upstream docs.


voodoohop:

PR #14650, opened 2026-09-09
feat: make quest badge actionable

tomdacatto:
- Replaces the static quest badge with live account counts.
- Shows unclaimed rewards as `X claimable!`, otherwise incomplete available quests as `X new!`, and hides the badge when nothing needs attention.
- Loads quest state after render and refreshes after automatic quest checks and claims; adds focused status-priority tests.

Tests: 3 focused tests, Biome, diff check, and Enter frontend production build.


Circuit-Overtime (approved):
Thanks for keeping this focused. The badge now prioritizes earned, unclaimed rewards, otherwise counts only available quests the account has not completed, and disappears when there is nothing actionable.

The refresh path after automatic checks and claims is appropriately scoped, and the request ID plus cleanup guard prevents stale responses from overwriting newer state. The change stays within navigation status and its two mutation points. LGTM.

PR #14651, opened 2026-09-09
feat: prepare Reddit news publishing in Actions

Itachi-1824:
- Adds a manual Devvit workflow for dated Reddit news posts, defaulting to validation only.
- Records post receipts to avoid duplicate submissions and fails on unconfirmed delivery.
- Leaves existing publishing unchanged; activation awaits credentials, dependency updates, and a verified test post.

### DIFF
--- a/.github/workflows/reddit-news-devvit.yml
+++ b/.github/workflows/reddit-news-devvit.yml
@@ -1,0 +1,99 @@
+name: Reddit news Devvit preparation
+
+on:
+  workflow_dispatch:
+    inputs:
+      mode:
+        description: "Validation defaults to no external writes"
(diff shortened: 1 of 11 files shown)


ElliotEtag (commented):
@Itachi-1824, keeping the existing Devvit approach makes sense. A few things look worth addressing before merging:

- The entry point and trigger configuration appear to prevent `AppUpgrade` from registering.
- Empty payloads and failed image uploads leave permanent `pending` locks, blocking retries.
- Formatting checks are failing.

Before activation, a verified post and duplicate retry would help confirm delivery. The generated body text is also currently omitted.

Issue #14660, opened 2026-09-09
[QUEST] Add Pollinations to Simon Willison’s LLM


voodoohop:
### How this quest works

Multiple contributors may submit PRs. We will select the clearest, smallest completed approach. The author of the selected merged PR receives the reward.

### Goal

Add a native Pollinations provider plugin for Simon Willison’s `llm` CLI and Python library.

### Outcome

- `llm-pollinations` is an installable, focused Python package.
- It reuses LLM’s OpenAI-compatible `Chat` and `AsyncChat` classes against `https://gen.pollinations.ai/v1`; it does not recreate streaming, conversations, tools, or attachment handling.
- Authentication uses `llm keys set pollinations` or `POLLINATIONS_API_KEY`.
- Compatible text models are loaded from the authenticated `/v1/models` catalog and registered as `pollinations/<model-id>` without a hardcoded model list or provider-name collisions.
- Image input, tool calling, and reasoning are enabled only when advertised by the catalog.
- Catalog loading uses a small time-limited cache with stale-cache fallback.
- Normal prompts, streaming, chat, an image attachment, tool calling, and the LLM Python API are verified against a current LLM release.
- Documentation includes direct key setup and creating a dedicated key through Polli CLI.
- The package is ready for PyPI publication and a concise upstream plugin-directory submission after merge.

Keep it focused: image generation, embeddings, a custom device-auth flow, a configuration UI, and a new Polli CLI integration command are outside this first version.

### Resources

- [LLM repository](https://github.com/simonw/llm)
- [Developing an LLM model plugin](https://llm.datasette.io/en/stable/plugins/tutorial-model-plugin.html)
- [LLM OpenAI-compatible configuration](https://llm.datasette.io/en/stable/other-models.html#openai-compatible-models)
- [llm-openrouter reference](https://github.com/simonw/llm-openrouter)
- [LLM plugin directory](https://llm.datasette.io/en/stable/plugins/directory.html)
- [Pollinations model catalog](https://gen.pollinations.ai/v1/models)
- [Pollinations API documentation](https://gen.pollinations.ai/docs)
- [Polli CLI](https://github.com/pollinations/pollinations/tree/main/packages/polli-cli)

### Reward

10 Pollen

davealan74:
Opened a PR: #14662

- `llm-pollinations` package registers Pollinations text models as `pollinations/<model-id>` inside Simon Willison's `llm` CLI/library.
[...]

PR #14662, opened 2026-09-09

davealan74:
Fixes #14660

Adds `packages/llm-pollinations`, a native provider plugin for Simon Willison's `llm` CLI/library.

- Registers Pollinations' text models as `pollinations/<model-id>`, loaded live from `gen.pollinations.ai/v1/models` (no hardcoded list).
- Reuses `llm`'s built-in `Chat`/`AsyncChat` OpenAI-compatible base classes — streaming, conversations, tool calling, and attachments are handled by `llm`, not reimplemented.
- Vision, tool calling, and reasoning are only enabled per-model when the catalog advertises them (`input_modalities`, `tools`, `reasoning`).
- Catalog fetch is gated on a configured key (`llm keys set pollinations` or `POLLINATIONS_API_KEY`) and cached for an hour with stale-cache fallback on network errors.
- README covers install, key setup (direct + via `polli keys create`), CLI and Python usage.

**Testing:**
- `python -m pytest` — 8 unit tests covering model filtering, capability detection, cache read/write/stale-fallback, and `register_models` behavior (no key → no-op, dedup, correct kwargs).
- Verified live against `gen.pollinations.ai/v1/models` (public, no key required to list): confirms real catalog shape and that filtering to `category == "text"` + `/v1/chat/completions` support yields 282 unique, non-colliding model IDs with correct vision/tools/reasoning flags.
- No live Pollinations API key with balance was available in this environment to run real chat completions, so request/response wiring (non-streaming, streaming, multi-turn conversation, image attachment, tool-calling loop, and the async Python API) was verified end-to-end against a local OpenAI-compatible mock server through the actual `PollinationsChat`/`PollinationsAsyncChat` classes — all passed.

### DIFF
--- a/packages/README.md
+++ b/packages/README.md
@@ -24,6 +24,12 @@ A terminal app for using Pollinations without writing any code. Type a prompt, g
 
 **Who it's for:** people comfortable in a terminal — humans testing things quickly, and AI agents driving workflows.
 
+### [llm-pollinations/](./llm-pollinations) — plugin for Simon Willison's `llm`
+
+A native provider plugin that registers Pollinations' text models inside the [`llm`](https://llm.datasette.io/) CLI and Python library, so you can run `llm -m pollinations/<model>` alongside any other provider.
+
+**Who it's for:** developers already using the `llm` CLI/library who want Pollinations models available without leaving that tool.
+
 ### [n8n/](./n8n) — n8n tunnel setup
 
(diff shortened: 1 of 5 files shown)


voodoohop:
- For future plugins, handle discovery/file errors at registration so a failed catalog load does not affect other LLM providers.

PR #14665, opened 2026-09-09
feat(models): filter listings by reliability and expose health metadata

TRDEFI:
﻿Fixes #14535

Adds reliability filtering and minimal health metadata to model discovery, building on the existing `?community=` filter instead of a parallel system.

- `GET /v1/models`, `GET /v1/models/:model` and the rich lists (`/models` and the category lists share one factory) accept `?reliability=reliable`. Clients that append `/models` to a base URL and cannot add query strings (Open WebUI, LibreChat, Cline) can send `X-Model-Source: official|community|all` and `X-Model-Reliability: reliable|all` headers instead; query params win when both are present.
- Every entry carries `reliability` (`reliable` / `unreliable` / `unknown`) and a minimal `health` object (success rate, sample count, window, freshness). A model is reliable at 95%+ success over 20+ recent requests; fallback rescues count as successes, caller-side 4xx is excluded, and missing data means `unknown`. Experimental status is untouched.
- Filtering runs after permission and paid-balance filtering, so access rules and default listings are unchanged. Health comes from the existing cached `model_health` feed (60s TTL, stale fallback); when it is unreachable everything reports unknown rather than failing discovery.
- Docs in `gen.pollinations.ai/src/docs/models.md` with curl, Python, and header-based client examples.
- 13 focused unit tests cover the rate math (including the rescue overlap clamp), thresholds, param/header parsing and precedence, and the filter itself. `vitest run` and `tsc --noEmit` pass in `gen.pollinations.ai`.

### DIFF
--- a/gen.pollinations.ai/src/docs/models.md
+++ b/gen.pollinations.ai/src/docs/models.md
@@ -27,6 +27,56 @@ Any other value (e.g. `tru`, `yes`, `2`) returns **400 Bad Request**.
 
 Example: `GET /models?community=false`
 
+### Reliability filtering and health metadata
+
+All model discovery endpoints also accept an optional `reliability` query
+parameter:
+
+| Parameter | Values | Behaviour |
+|-----------|--------|-----------|
+| *(omitted)* | | Returns all models (default, backward-compatible) |
+| `reliability=reliable` | `reliable` | Keeps only models with a high recent success rate |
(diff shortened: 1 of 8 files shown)


voodoohop:
Closing because #14553 was selected for this quest. Thank you for the thorough submission. I replied with the brief comparison and 3-Pollen bonus on #14535.

PR #14667, opened 2026-09-10
feat(models): expose verified chat parameters in listings

TRDEFI:
﻿Fixes #14599

Exposes which Chat controls each model honors, verified against the generation pipeline rather than upstream docs.

- New `shared/registry/model-parameters.ts` holds the gateway's base control surface plus per-model profiles derived from the request transforms in `availableModels.ts`: sampling knobs stripped upstream never appear, `reasoning_effort` shows as supported where it is honored, mapped (Claude thinking budget), or toggled (Fireworks), and gateway defaults only list supported controls.
- Three official Chat models with different capabilities to start: `openai/gpt-5.4-nano` (no sampling knobs), `deepseek/deepseek-v4-flash` (full surface), `anthropic/claude-sonnet-4.6` (no temperature/top_p, reasoning mapped). Anything unverified omits the fields instead of guessing; aliases resolve to the same definition so all listings agree.
- Threaded through `ModelDefinition` → `ModelInfo` → the OpenAI mapper, so `/models`, `/text/models`, `/v1/models` and the single-model lookup all expose `supported_parameters` / `default_parameters`. Generation, billing, and access rules untouched. The fields describe Chat Completions (and the text GET surface), not the native Responses API.
- Docs in `gen.pollinations.ai/src/docs/models.md` with a lookup example.
- 8 focused tests cover the three profiles, alias consistency, defaults ⊆ supported, and the listing threading. `vitest run` and `tsc --noEmit` pass in `gen.pollinations.ai`.

### DIFF
--- a/gen.pollinations.ai/src/docs/models.md
+++ b/gen.pollinations.ai/src/docs/models.md
@@ -32,6 +32,31 @@ Rich model endpoints include `capabilities` for agentic/model traits:
 Modalities, video frame controls, voices, and context length remain separate
 structured fields.
 
+### Supported parameters
+
+Models verified against the generation pipeline expose
+`supported_parameters` (the Chat controls they honor) and
+`default_parameters` (gateway defaults applied when omitted):
+
+```bash
+curl "https://gen.pollinations.ai/v1/models/openai" | jq '{id, supported_parameters, default_parameters}'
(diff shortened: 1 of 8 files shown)


voodoohop:

PR #14668, opened 2026-09-10
feat(llm): add Pollinations provider plugin for Simon Willison's llm

TRDEFI:
﻿Fixes #14660

Adds `llm-pollinations`: Pollinations text models inside Simon Willison's `llm` as `pollinations/<model-id>` (plus catalog aliases like `pollinations/openai`).

- Built on llm's own OpenAI-compatible `Chat`/`AsyncChat` against `https://gen.pollinations.ai/v1` — no reimplementation of streaming, conversations, tools, or attachments.
- Auth via `llm keys set pollinations` or `POLLINATIONS_API_KEY`; README documents minting a dedicated key with `polli keys create`.
- Models load live from the authenticated `/v1/models` catalog (community models included), cached for an hour with stale-cache fallback. No hardcoded list, `pollinations/` prefix avoids collisions.
- Vision, tool-calling, and reasoning flags follow whatever the catalog advertises per model.
- Verified end to end against the real API with llm 0.35 on a funded key: plain prompt, streaming, multi-turn chat (remembers context), image attachment, tool-calling loop (21+21 via tool = 42), and the async Python API.
- 8 unit tests cover filtering, capability flags, aliases, cache/fallback behavior, and registration. Docs cover install, key setup, usage, and PyPI + upstream directory submission.

### DIFF
--- a/packages/README.md
+++ b/packages/README.md
@@ -24,6 +24,12 @@ A terminal app for using Pollinations without writing any code. Type a prompt, g
 
 **Who it's for:** people comfortable in a terminal — humans testing things quickly, and AI agents driving workflows.
 
+### [llm-pollinations/](./llm-pollinations) — the `llm` CLI plugin
+
+Pollinations text models inside [Simon Willison's `llm`](https://llm.datasette.io/) as `pollinations/<model-id>`, read live from your `/v1/models` catalog. No hardcoded model list.
+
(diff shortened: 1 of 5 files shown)


voodoohop:
- One improvement: check the advertised chat-completions endpoint; text output alone also admits Responses-only agents.

PR #14671, opened 2026-09-10
[feat]   Add private media library for users to store their creation

### LINKED ISSUE #14582: 🗂️ Add a private list of uploaded and generated media
- Let users list their uploads and generations, including untagged media. Record user association independently of publishing: tags still control public galleries, and the personal list requires authentication. This does not make existing file URLs private.
- Decide how shared cached results appear in personal lists and what deletion means without removing a shared file for everyone. Historical files with unknown creators should not be attributed to whoever triggers their migration.
- Private library listing was explicitly deferred in #12277. Keep this separate from the cache-storage migration and coordinate with the S3 storage work in #13515.

gggff123:
Closes #14582


Circuit-Overtime (approved):
Thank you for working through the shared-cache ownership edge case here. Separating single-owner uploads from per-user media associations is the right model: cache hits can appear in each requester's library, while unlinking never removes the shared object or another user's association.

I rebased the branch onto current `main` and added a focused follow-up commit:

- returns `source: upload | generation` so clients can choose the appropriate delete/unlink action;
- keeps the private endpoint authenticated in the generated OpenAPI contract;
- updates stale integration expectations now that untagged uploads are cataloged;
- covers private upload visibility, secret-key access, shared-generation association, per-user unlinking, and the required `source` fixture field.

The final scope stays within authenticated media listing and its necessary upload/generation association paths. LGTM.

PR #14674, opened 2026-09-10

xiaotian1171:
Fixes #14660

Adds `packages/llm-pollinations`, a native provider plugin for Simon Willison's `llm` CLI/library.

- Registers Pollinations text models as `pollinations/<model-id>`, loaded live from the authenticated `gen.pollinations.ai/v1/models` catalog (no hardcoded list). Only models exposing `/v1/chat/completions` are registered, so image/audio/transcription endpoints are excluded.
- Reuses `llm`'s built-in `Chat`/`AsyncChat` OpenAI-compatible base classes — streaming, conversations, tool calling and attachments are handled by `llm`, not reimplemented.
- Vision, tool calling and reasoning are enabled per model only when the catalog advertises them (`input_modalities`, `capabilities`, top-level flags).
- Catalog fetch uses the configured key, is cached for an hour on disk, and falls back to the stale cache on network errors. `llm pollinations models --skip-cache` bypasses the cache.
- README covers install, key setup (direct + via `polli keys create`), CLI and Python usage.

**Testing:**
- `python -m pytest` — 10 unit tests covering model filtering, capability detection (capabilities list + top-level flags), cache read/write/stale-fallback, and `register_models` behavior (no key → no-op, dedupe, sync+async pairs, reasoning → `reasoning_effort` option).
- Verified live against a funded Pollinations key with llm 0.35: normal prompts, streaming, `--no-stream`, interactive `llm chat`, image attachment (1×1 png → "pink"), tool calling (23×42 via a `multiply` tool → 966), multi-turn conversation memory, the async Python API, and catalog-driven registration (179 chat-capable models registered from the authenticated catalog, 95 of them with reasoning).

### DIFF
--- a/packages/README.md
+++ b/packages/README.md
@@ -24,6 +24,12 @@ A terminal app for using Pollinations without writing any code. Type a prompt, g
 
 **Who it's for:** people comfortable in a terminal — humans testing things quickly, and AI agents driving workflows.
 
+### [llm-pollinations/](./llm-pollinations) — plugin for Simon Willison's `llm`
+
+A native provider plugin that registers Pollinations text models inside the [`llm`](https://llm.datasette.io/) CLI and Python library, so you can run `llm -m pollinations/<model>` alongside any other provider.
+
+**Who it's for:** developers already using the `llm` CLI/library who want Pollinations models available without leaving that tool.
+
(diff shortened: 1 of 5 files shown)


voodoohop:
- For a thin integration, prefer LLM’s existing model-list commands over adding provider-specific commands.

PR #14678, opened 2026-09-10
feat: add llm-pollinations plugin

afanasevmylife:
Closes #14660.

Minimal LLM CLI plugin for Pollinations AI:
- `llm keys set pollinations` for API key
- `llm -m pollinations/gpt-5.4-nano "prompt"`
- Auto-discovered models from /v1/models
- Vision, streaming, async, tool calls
- Fallback models when API unreachable
- Tests with mocked API responses

235 lines total — thinner than competing PRs.


voodoohop:
- One key fix for future plugins: register model instances, not classes, using LLM’s actual registration signature.

PR #14687, opened 2026-09-10

ammarelshaf3y:
Fixes #14660

Adds `llm-pollinations`: a minimal, installable Python package exposing Pollinations text models as `pollinations/<model-id>` in Simon Willison's `llm` CLI and Python library.

- Reuses `llm`'s OpenAI-compatible `Chat`/`AsyncChat` against `https://gen.pollinations.ai/v1` (no custom streaming/conversation/tool plumbing)
- Auth via `llm keys set pollinations` or `POLLINATIONS_API_KEY`
- Models loaded from the authenticated `/v1/models` catalog (text + chat-completions only, no hardcoded list, no id collisions)
- Vision/tools/reasoning enabled only when the catalog advertises them
- Per-key 30-min catalog cache with stale-cache fallback (different keys can see different catalogs)
- 8 focused tests (registration, capability mapping, no-key skip, cache + fallback, per-key scoping, entry point)
- README with key setup (incl. dedicated key via Polli CLI) and usage examples

Verified: pytest suite green (8/8); wheel + sdist builds verified (entry point present); mapping validated against a live `/v1/models` snapshot (384 models → 257 registered, 0 non-text leaked). No live prompt run (needs a key in CI).

Out of scope per the quest: image generation, embeddings, device-auth flow, config UI, new Polli CLI commands.


ammarelshaf3y:
Friendly ping — CI workflows are still awaiting approval so checks haven't run yet. Since opening, I've added per-key cache scoping, an 8th test, and verified wheel/sdist builds (see updated description). Happy to adjust anything.

voodoohop:
- One test improvement: verify reasoning_effort in the model’s Options fields; current LLM does not expose a model.reasoning attribute.

PR #14689, opened 2026-09-10
feat: add Pollinations plugin for LLM

tomdacatto:
Fixes #14660

- Adds a four-file `llm-pollinations` package built on LLM 0.35's native `Chat` and `AsyncChat` classes.
- Registers authenticated chat models as `pollinations/<model-id>` with catalog-driven vision, tools, and reasoning; no aliases, heuristics, or hardcoded models.
- Uses a one-hour key-scoped catalog cache with stale fallback for transport, malformed-response, and server failures.
- Documents LLM key storage, environment auth, dedicated Polli CLI keys, CLI/Python usage, and PyPI publication.

Verification:
- 6 focused cache, auth, filtering, deduplication, and capability tests pass.
- LLM 0.35 native smoke passes prompt, streaming, multi-turn chat, image attachment, tool call, and sync/async Python paths through a local OpenAI-compatible server.
- Wheel and sdist build; Ruff passes; live catalog schema check found 259 unique chat-compatible models.


tomdacatto:
- Addressed the two blocking review items in commit 96ef1da11: added path-scoped Python CI and isolated malformed catalog records.
- Test llm-pollinations now installs .[test], runs the focused suite, and builds wheel plus sdist; the final run passed in 20 seconds.
- Invalid supported_endpoints or input_modalities values now skip only that record; the regression places both malformed shapes before a valid model.
- Added explicit contents: read workflow permissions in commit aed6bad42; the CodeQL thread is resolved and all checks pass.
- Kept atomic cache writes out of this focused PR because the review marked them non-blocking.

gggff123 (approved):

voodoohop:
- Your direct capability mapping informed the final simplification, and you are credited as co-author. Your 2-Pollen bonus is ready to claim on [Quests](https://enter.pollinations.ai/quests).

PR #14706, opened 2026-09-10
feat: add llm-pollinations plugin

SvirepyiBambr:
Adds a native Pollinations provider plugin for Simon Willison's [llm](https://llm.datasette.io/) CLI and Python library.

Fixes #14660

## What it does

- `llm-pollinations` is an installable, focused Python package at `packages/llm-pollinations/`
- Reuses LLM's OpenAI-compatible `Chat`/`AsyncChat` classes against `https://gen.pollinations.ai/v1` — no recreated streaming, conversations, tools, or attachment handling
- Authentication via `llm keys set pollinations` or the `POLLINATIONS_API_KEY` environment variable
- Compatible text models are loaded dynamically from the authenticated `/v1/models` catalog and registered as `pollinations/<model-id>` — no hardcoded model list, no provider-name collisions
- Image input, tool calling, reasoning, and structured outputs are enabled only when the catalog advertises them
- Catalog loading uses a small time-limited cache (1 hour) in the llm user directory with stale-cache fallback when Pollinations is unreachable
- Documentation includes direct key setup and creating a dedicated key through Polli CLI

## Verified against current llm (0.35)

- Normal prompts, streaming, `llm chat`, an image attachment, tool calling (`llm.Tool.function` + `response.reply()`), and the LLM Python API — 174 models registered dynamically from the live catalog
- 8 focused unit tests: catalog filtering, capability flags, sync+async registration, key gating, cache freshness / skip_cache / stale-fallback / fetch-and-cache

## Notes

- `httpx` is declared explicitly: current llm ships an internal `httpx2`, so plugins importing `httpx` need their own dependency (caught by tests)
- Out of scope per the issue: image generation, embeddings, a custom device-auth flow, a configuration UI, and a new Polli CLI integration command

### DIFF
--- a/packages/llm-pollinations/README.md
+++ b/packages/llm-pollinations/README.md
@@ -1,0 +1,55 @@
+# llm-pollinations
+
+[Pollinations](https://pollinations.ai) models for [Simon Willison's llm](https://llm.datasette.io/).
+
+Text models are loaded from the [Pollinations catalog](https://gen.pollinations.ai/v1/models) and registered as `pollinations/<model-id>` — no hardcoded model list. The catalog is cached for an hour in the llm user directory and falls back to a stale copy when Pollinations is unreachable.
+
+Image input, tool calling, reasoning and structured outputs are enabled only when the catalog advertises them for a model.
+
+## Installation
+
+```bash
+llm install llm-pollinations
+```
+
+## API keys
+
+```bash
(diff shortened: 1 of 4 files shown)


voodoohop:
- One improvement: catch catalog download failures during registration so unrelated LLM providers still work when Pollinations is unreachable.

PR #14708, opened 2026-09-10
Fixes #14535: Filter model catalogs by source and reliability

SvirepyiBambr:
Fixes #14535

Model list endpoints gain source/reliability discovery filters and minimal health metadata.

## What it does

- `?source=official|community` replaces the `?community=true|false|1|0` filter (the issue explicitly allows breaking compatibility); an optional `X-Pollinations-Model-Source` header does the same for clients whose fixed base URLs cannot carry query strings — the query parameter wins when both are present
- `?reliable=true` keeps only models whose health window shows a success rate of at least 0.9 (`RELIABLE_SUCCESS_RATE_THRESHOLD`); models without health data are excluded — unknown is not reliable
- Listings expose minimal `health` metadata — `success_rate` (0–1), `samples`, `window_minutes` (60), `as_of` freshness — computed from the same cached Tinybird `model_health` pipe that powers `/v1/models/status`, now extracted into a shared `fetchModelHealthData`
- Success rate counts fallback rescues as successes (they end in 2xx) and excludes caller-side 4xx failures from the denominator; missing data means unknown — entries without health rows stay untouched
- Experimental (`alpha`) status stays separate from reliability; access rules, permission filtering, and the default listing are unchanged

Applied to `/v1/models`, `/v1/models/:model`, `/models`, `/text/models`, and the other category lists through the shared `modelsListHandler`.

## Tests

New focused suite `gen.pollinations.ai/test/model-source-health.test.ts`:
- source filtering (official / community / omitted)
- health computation (fallback rescues as successes, caller-side failures excluded, unknown cases)
- the reliability threshold

## Docs + client examples

- `gen.pollinations.ai/src/docs/models.md` documents the filters, the health metadata, and client examples:
  ```bash
  curl "https://gen.pollinations.ai/v1/models?source=official"
  curl "https://gen.pollinations.ai/v1/models?reliable=true"
  curl -H "X-Pollinations-Model-Source: official" "https://gen.pollinations.ai/v1/models"
  ```

Information only: no generation behavior, billing, or access rules changed.

### DIFF
--- a/gen.pollinations.ai/src/docs/models.md
+++ b/gen.pollinations.ai/src/docs/models.md
@@ -42,6 +42,29 @@ Supported media models also advertise both endpoints and return generated-file
 links as assistant text. Reference-required models return their normal missing-input
 error; use their native endpoint until attachments are supported here.
 
+Chat model listings also expose `supported_parameters` — the OpenAI-compatible
+Chat Completions request parameters a model accepts through Pollinations — and
+`default_parameters`, which lists only the defaults the gateway itself applies
+(unset parameters use the upstream model's own defaults). These fields describe
+the Chat Completions and `/text` routes, not the native Responses API.
+
+Model lists also accept two discovery filters: `?source=official|community`
+picks built-in or community models, and `?reliable=true` keeps only models
+whose health window shows a success rate of at least 0.9. A matching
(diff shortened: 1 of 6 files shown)


voodoohop:
Thanks for the thorough submission. #14553 was selected for this quest, and the catalog health path has since changed. Closing this superseded branch.

PR #14716, opened 2026-09-11
Fixes #14660: add native llm-pollinations provider plugin for Simon Willison's llm

CryptoRepublic:
Fixes #14660.

Native Pollinations provider plugin (`llm-pollinations`) for Simon Willison's llm CLI + Python library. Focused package under `packages/llm-pollinations` (single module, pyproject, README, pytest suite).

What:
- Reuses LLM's OpenAI-compatible `Chat` / `AsyncChat` against `https://gen.pollinations.ai/v1` — no recreated streaming, conversations, tools, or attachment handling (subclasses add only key config).
- Auth via `llm keys set pollinations` or `POLLINATIONS_API_KEY`; models register only when a key is configured so the catalog reflects the key's permissions.
- Compatible text models load from the authenticated `/v1/models` catalog, registered as `pollinations/<model-id>` (namespaced, no hardcoded list, no provider-name collisions). Non-text categories excluded.
- Vision / tool-calling / reasoning enabled only when the catalog advertises them (`input_modalities` includes image; `capabilities`/`tools`/`reasoning`); reasoning wires LLM's standard `-o reasoning_effort` option.
- 5-minute cache with stale-cache fallback (`llm pollinations refresh`, `llm pollinations models [--json]`); discovery never breaks `llm` startup when offline/uncached.
- README: direct key setup + dedicated key via Polli CLI (`polli keys create --name llm --budget 100`), usage, verification commands, PyPI + upstream plugin-directory submission notes.
- Focused pytest suite (6 tests): base-class reuse, auth wiring, capability gating, namespacing, reasoning options, stale fallback.

Verified live against llm 0.35 (current): plugin loads (`llm plugins` shows both hooks), `llm pollinations models` works, and with a key set 262 `pollinations/*` text models register with correct vision/tools/reasoning flags (`llm models --options` checked). Prompt/streaming/chat/attachment/tool/Python-API paths are pure base-class behavior (same as the llm-openrouter reference); run the README verification block with a funded key for end-to-end generation.

Out of scope (per quest): image generation, embeddings, custom device-auth flow, config UI, new Polli CLI integration command.

### DIFF
--- a/packages/llm-pollinations/README.md
+++ b/packages/llm-pollinations/README.md
@@ -1,0 +1,96 @@
+# llm-pollinations
+
+Native [Pollinations](https://pollinations.ai) provider plugin for [Simon Willison's `llm`](https://github.com/simonw/llm) CLI and Python library.
+
+It reuses LLM's OpenAI-compatible `Chat` / `AsyncChat` against `https://gen.pollinations.ai/v1` — streaming, conversations, tools, and attachments all come from the base classes. Compatible text models load from your authenticated `/v1/models` catalog and register as `pollinations/<model-id>` with no hardcoded list.
+
+Out of scope for this first version: image generation, embeddings, custom device-auth flows, configuration UIs, and new Polli CLI integration commands.
+
+## Install
+
+```bash
+llm install llm-pollinations
+```
+
+Or from this repo:
+
+```bash
+llm install -e packages/llm-pollinations
+```
+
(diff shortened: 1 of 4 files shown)


voodoohop:
- One improvement: filter by the chat-completions endpoint and text output, not category alone, to exclude Responses-only agents.

PR #14717, opened 2026-09-11
Add native llm-pollinations provider plugin

Marcus-Mok-GH:
## Fixes #14660: add native llm-pollinations provider plugin

This PR adds a focused native Pollinations provider for Simon Willison's `llm` CLI and Python library. It reuses LLM's existing OpenAI-compatible `Chat` / `AsyncChat` classes rather than reimplementing streaming, conversations, tools, or attachment handling.

This is the smaller, four-file implementation for the quest: 383 added lines versus 488 in the related #14716 proposal. It keeps the scope limited to the provider package, model discovery, documentation, and focused tests.

The plugin:
- accepts the live `output_modalities: ["text"]` catalog schema while retaining the legacy `category: "text"` fallback
- excludes image-only and non-chat models, namespaces registrations as `pollinations/<model-id>`, and deduplicates model IDs
- enables vision, tools, and reasoning only when advertised by the catalog
- uses a five-minute, API-key-scoped cache with stale-cache fallback for normal discovery and explicit failure for forced refresh
- registers models only when a Pollinations key is configured

Verification: `python3 -m py_compile` passed for the plugin and test module. The focused tests cover authentication, key-scoped caching, stale-cache behavior, schema filtering, deduplication, capability mapping, malformed entries, and no-key registration.

### DIFF
--- a/packages/llm-pollinations/README.md
+++ b/packages/llm-pollinations/README.md
@@ -1,87 +1,89 @@
 # llm-pollinations
 
-Native [Pollinations](https://pollinations.ai) provider plugin for [Simon Willison's `llm`](https://github.com/simonw/llm) CLI and Python library.
+Native [Pollinations](https://pollinations.ai) provider plugin for [Simon Willison's \x60llm\x60](https://github.com/simonw/llm) CLI and Python library.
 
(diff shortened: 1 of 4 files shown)


voodoohop:
- Your submission passed our live smoke checks; a 2-Pollen runner-up bonus is ready on [Quests](https://enter.pollinations.ai/quests). One documentation fix: remove the model-list/refresh commands that are not implemented.

PR #14735, opened 2026-09-11
feat: OAuth refresh tokens via offline_access

voodoohop:
- Adds `offline_access` scope: with it the code exchange returns a one-hour `expires_in` plus a `refresh_token`; without it the response is unchanged (`expires_in` = key lifetime, no refresh token)
- Adds `grant_type=refresh_token` (`enter.pollinations.ai/src/routes/oauth-refresh.ts`): returns the approved key while it exists; if the user deleted it, re-mints it with the approved name, models, account permissions and remaining expiry, and the budget left at the last refresh (snapshotted hourly), so deleting the key never resets spend (`createdVia: redirect-auth`); a disabled key is left alone; `invalid_grant` once the approved expiry has passed or the client registration no longer covers the redirect
- Grant lives in KV under `oauth-refresh:` with TTL = approved key expiry; refresh token is not rotated
- Consent page forwards `offline_access` in the granted scope; RFC 8414 metadata advertises the scope and grant
- Open WebUI requests `profile offline_access`, so a consent key deleted from the dashboard heals at the next refresh (≤1 h) instead of failing with "A valid API key is required" until re-login
- Docs: BRING_YOUR_OWN_POLLEN.md, enter AGENTS.md, apps/openwebui/README.md
- Tests: `test/oauth-code.test.ts` (22 pass)

https://claude.ai/code/session_015mNrP6DkRQmm3YStNjfFv3


voodoohop:
Parked: Open WebUI keeps the user session when a refresh fails, so a refresh grant without re-minting changes nothing for it, and re-minting a deleted key is a consent question we don't want to answer implicitly. Branch kept for reference.

PR #14745, opened 2026-09-12

kreggscode:
Fixes #14660

## Summary

Native Pollinations provider plugin for Simon Willison's llm CLI and Python library.

## What it does

- Reuses LLM's OpenAI-compatible Chat/AsyncChat against https://gen.pollinations.ai/v1 — no recreated streaming, conversations, tools, or attachment handling
- Auth via llm keys set pollinations or POLLINATIONS_API_KEY
- Models loaded from the authenticated /v1/models catalog, registered as pollinations/<model-id> (no hardcoded model list, no provider-name collisions)
- Vision, tool calling, and reasoning enabled only when the catalog advertises them
- Per-key 30-minute catalog cache with stale-cache fallback (different keys see different catalogs)

## Key differentiator: zero extra dependencies

Uses Python stdlib urllib.request for HTTP — no httpx dependency. Only depends on llm itself. This means:
- Smaller install footprint
- No transitive dependency conflicts
- Works in environments where httpx is unavailable

## Files

- packages/llm-pollinations/llm_pollinations.py — 93 lines, single-module plugin
- packages/llm-pollinations/pyproject.toml — minimal metadata + entry point
- packages/llm-pollinations/README.md — install, setup, usage, dev instructions
- packages/llm-pollinations/tests/test_llm_pollinations.py — 11 focused tests
- packages/README.md — updated with plugin entry

## Tests (11 passing)

- Helper functions: _is_chat_model, _has_vision, _has_tools, _has_reasoning
- Registration: with key, without key (skipped), capability mapping
- Cache: path scoping, cache hit, stale fallback when offline
- Entry point: pyproject.toml validation

## Verified

- All 11 tests pass (pytest -v)
- Wheel + sdist build successfully
- Plugin registered via llm plugins
- No live API calls in tests (all mocked)

### DIFF
--- a/packages/README.md
+++ b/packages/README.md
@@ -30,6 +30,12 @@ A small set of helper scripts for exposing a self-hosted [n8n](https://n8n.io) a
 
 **Who it's for:** the Pollinations team and self-hosters running n8n alongside Pollinations.
 
+### [llm-pollinations/](./llm-pollinations) — the `llm` CLI plugin
+
+A native Pollinations provider for Simon Willison's [llm](https://llm.datasette.io/) tool. Install it, set your key once, and prompt any Pollinations text model from the terminal or Python as `pollinations/<model-id>`.
+
+**Who it's for:** terminal-first developers and scripters who already use `llm` and want Pollinations models inside it.
+
 ## How they relate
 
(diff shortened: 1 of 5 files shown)


voodoohop:
- One improvement: handle discovery failures at registration so a temporary outage does not prevent other LLM providers from loading.

PR #14756, opened 2026-09-12
refactor: pick the search context size per fallback attempt

voodoohop:
- Moves the `web_search_options` normalization in `gen.pollinations.ai/src/text/handler.ts` into the fallback attempt, so the candidate that actually runs decides the `search_context_size` it sends and the pricing input it bills.
- Keeps the 400 for an unsupported size before the loop, validated against the model the caller asked for, with the same body.
- No behaviour change for declared registry routes: they inherit `searchContextSizes` from their parent. It matters once a fallback chain mixes models with different (or no) search sizes.
- Verified: `tsc`, biome, `fallback.test.ts`, `static-provider-fallbacks.test.ts`, and the five Perplexity search-context tests in `generation-vcr.test.ts`.

https://claude.ai/code/session_0161ggPVjwNMTu6oaevEScZw

### DIFF
--- a/gen.pollinations.ai/src/text/handler.ts
+++ b/gen.pollinations.ai/src/text/handler.ts
@@ -338,11 +338,8 @@ async function generateTextResponse(
     syncTextEnvironment(c.env);
 
     try {
-        const normalization = normalizeSearchContext(c, requestData);
-        if ("errorResponse" in normalization) {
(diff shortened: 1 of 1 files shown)


voodoohop:
Superseded by the thin-client version: forward web_search_options untouched and bill the tier from the response.

PR #14791, opened 2026-09-12
feat: add openai/gpt-4o-mini via OpenRouter

tomdacatto:
## Summary
- Adds `openai/gpt-4o-mini` as a new paid-only text model (no aliases), routed through OpenRouter
- Pins OpenRouter provider order to `[azure, openai]` with `allow_fallbacks: false`, excluding the `azure/swedencentral` endpoint (bills 10% higher) so pricing stays fixed
- Pricing: $0.15/M input, $0.60/M output, $0.075/M cache read — verified live against OpenRouter's `/api/v1/models/openai/gpt-4o-mini/endpoints`
- Context 128K, max output 16,384 tokens (also verified live)
- `priceMultiplier: 1`, `paidOnly: true`

## Contract (confirmed)
| Field | Value |
|---|---|
| Canonical name | `openai/gpt-4o-mini` |
| Aliases | none |
| Registry provider | `openrouter` |
| Primary route | OpenRouter `azure` tag |
| Fallback | OpenRouter `openai` tag (same price) |
| Pollinations GPU | no |

## Verification
- Pricing/context/max-output/supported-params confirmed against OpenRouter's public endpoints API (unauthenticated, no request needed)
- Added a unit test (`modelResolver.test.ts`) mirroring existing pinned-OpenRouter-model tests, asserting the `provider.order`/`allow_fallbacks` shape
- **Known limitation**: could not run the Vitest/CF-Workers suite or `tsc` in this sandbox — both fail environment-wide due to a pre-existing, unrelated issue (missing `@better-auth/api-key` dependency and a workerd compatibility-flag mismatch), not caused by this change. Structure and field types were manually verified against working entries (e.g. `poolside/laguna-s-2.1`, `meituan/longcat-2.0`). Recommend running the full suite in CI before merge.


ElliotEtag:
Tested end to end on local gen and pushed a small follow-up commit: explicit `ignore` for `azure/swedencentral` (the base slug `azure` matches it, so `order` alone did not exclude it), `maxReferenceImages`, entry placement, a comment fix, and the resolver test. Verified billing, prompt and output caching, tools, vision, structured output, and paid-only enforcement. We could not host this on our Azure resource because Azure refuses new gpt-4o-mini deployments (deprecating, retiring April 2027), so OpenRouter it is. Ready to merge from our side.

PR #14793, opened 2026-09-12
feat: add sakana/fugu-max via OpenRouter

tomdacatto:
## Summary
Draft PR — **not ready to merge**, opened for evaluation per the concerns below.

**Blocked (found in review):** every request to `sakana/fugu-max` — stream and non-stream, through Gen and via a direct OpenRouter probe — returns 403: `sakana/fugu-max is not available in your region: Sakana AI blocks requests originating from your location.` Reproducible, not transient. OpenRouter proxies to Sakana from its own infrastructure, so this is the same path production traffic would hit. Since `sakana` is the model's only endpoint, there is no fallback to route around it — this is a hard outage, not a degradation. Not resolved in this PR; staying in draft until Sakana access is confirmed.

- Adds `sakana/fugu-max` as a paid-only text model (no aliases), routed through OpenRouter, pinned to the `sakana` tag (its only endpoint — no fallback exists)
- $2/M input, $6/M output, $0.25/M cache read, 1M context, 128K max output — verified live against OpenRouter's `/api/v1/models/sakana/fugu-max/endpoints`
- `web_search_options` is intentionally **not** exposed: OpenRouter bills web search as a flat $0.01/call, a non-token charge our registry cost model can't meter today (the only precedent is Perplexity/Sonar's bespoke `perplexity-billing.ts` module — building an equivalent for Fugu is out of scope here)
- No transform applied to `reasoning_effort` — following the `thinkingmachines/inkling-small` precedent (single-provider, OpenRouter-native reasoning passthrough)

## Why draft
Per review:
- **Region block with no fallback path** (see above) — the model cannot currently serve any Pollinations traffic
- Fugu Max launched yesterday; OpenRouter reports **87.97% successful availability over 24h** despite 100% provider reachability — too weak for a dependable public offering today
- Internal multi-agent orchestration tokens are also billed, so the attractive per-token rate doesn't necessarily mean cheap completed tasks — unverified against actual task cost
- Quality is unproven for our use cases — needs comparison against Terra, Sonnet, and DeepSeek on real tasks, cost, and completion time
- No existing Fugu integration or related issue/PR was found

## Recommendation (from review)
Prioritize a small evaluation through OpenRouter first — verify streaming usage, internal-agent costs, and search charges — before enabling this publicly. This PR gives the model a registry entry to eval against without exposing it to public traffic while in draft. The region block must be resolved (or a working route found) before that evaluation can even start.

## Verification
- Contract confirmed against live OpenRouter endpoint data (single `sakana` tag, no ambiguity)
- Added a unit test (`modelResolver.test.ts`) asserting the pinned `provider.only`/`allow_fallbacks` shape
- Updated the `effective-prices` snapshot in `enter.pollinations.ai` for the new registry entry, including the mandatory 5.5% OpenRouter credit fee (#14895)
- **Not run**: Vitest/CF-Workers suite and `tsc` — both fail environment-wide in this sandbox due to a pre-existing, unrelated dependency/workerd issue (see #14791 for the same caveat). Recommend running full CI and, before un-drafting, the live-inference evaluation described above.

### DIFF
--- a/enter.pollinations.ai/frontend/public/brand-logos/sakana-ai.svg
+++ b/enter.pollinations.ai/frontend/public/brand-logos/sakana-ai.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor">
+  <title>Sakana AI</title>
+  <circle cx="12" cy="12" r="9"/>
+</svg>
--- a/enter.pollinations.ai/frontend/src/components/models/model-info.ts
+++ b/enter.pollinations.ai/frontend/src/components/models/model-info.ts
@@ -33,6 +33,7 @@ const BRAND_LOGOS: Record<string, string> = {
     Pruna: "pruna",
     Qwen: "qwen",
     Recraft: "recraft",
+    "Sakana AI": "sakana-ai",
     Sesame: "sesame",
     "Stability AI": "stability",
     StepFun: "stepfun",
--- a/enter.pollinations.ai/test/__snapshots__/effective-prices.snapshot.test.ts.snap
+++ b/enter.pollinations.ai/test/__snapshots__/effective-prices.snapshot.test.ts.snap
@@ -2626,6 +2626,18 @@ exports[`effective cost and price per model — snapshot 1`] = `
       "completionImageTokens": 0.08,
     },
   },
+  "sakana/fugu-max": {
+    "cost": {
+      "completionTextTokens": 0.000006,
+      "promptCachedTokens": 0.00000025,
+      "promptTextTokens": 0.000002,
+    },
+    "price": {
+      "completionTextTokens": 0.000006,
+      "promptCachedTokens": 0.00000025,
+      "promptTextTokens": 0.000002,
+    },
+  },
   "sesame/csm-1b": {
     "cost": {
       "completionAudioTokens": 0.000007,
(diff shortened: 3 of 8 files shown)


ElliotEtag:
Tested this branch end to end (local Gen + a direct OpenRouter probe), on top of the availability/quality concerns already in the PR description:

- **Region block, no fallback path.** Every request to `sakana/fugu-max` — stream and non-stream, through Gen and directly against OpenRouter with the same pinned `{only:["sakana"], allow_fallbacks:false}` — returns 403: `sakana/fugu-max is not available in your region: Sakana AI blocks requests originating from your location.` Reproducible, not transient. OpenRouter proxies to Sakana from its own infrastructure, so this is the same path production traffic would hit — not a local-only artifact. Since `sakana` is the model's only endpoint, it's a hard outage with no degradation path.
- **Registry cost is missing the OpenRouter 5.5% credit fee.** [text.ts#L1588-L1592](https://github.com/pollinations/pollinations/blob/56dc011ae8e1aa52f06e44fd55a7d91fda8e999c/shared/registry/text.ts#L1588-L1592) uses plain `perMillion(2)/perMillion(6)/perMillion(0.25)`. Every sibling OpenRouter model in this file (`meituan/longcat-2.0`, `thinkingmachines/inkling-small`) applies `perMillion(X) * 1.055`. As written this undercharges ~5.5% vs. the route's actual cost.

Rest checks out: real logo (not a placeholder), clean catalog description, `paidOnly` gate and error billing behave correctly, `modelResolver.test.ts`/tsc/biome all pass.

Given the region block plus zero fallback, I'd leave this in draft until at least one of those is resolved.

tomdacatto:
Applied the fee fix (1bc068de3). Not attempting to fix the region block — there's no second endpoint to route around it, and it's an infrastructure-level block on OpenRouter's side, not something fixable in this repo. Updated the PR description to surface it prominently up top. Staying in draft as you recommended until Sakana access is confirmed working.

PR #14798, opened 2026-09-12
feat: add tencent/hy4-preview via OpenRouter

### LINKED ISSUE #15138: Add tencent/hy4-preview via OpenRouter
### How this quest works

Retroactive quest: this was submitted unsolicited as PR #14798 before a quest issue existed for it. Opening this after the fact to record the fixed Pollen reward once it merges.

### Goal

Add `tencent/hy4-preview` as a new paid-only text model, routed through OpenRouter.

### What to build

- Registry entry in `shared/registry/text.ts` (pricing, context length, modalities)
- OpenRouter provider pin in `gen.pollinations.ai/src/text/configs/modelConfigs.ts`
- Catalog logo + model-info entry
- Resolver test coverage

### Acceptance criteria

- [x] `model=tencent/hy4-preview` resolves and generates via OpenRouter
- [x] Pricing includes the mandatory 5.5% OpenRouter credit fee (#14895)
- [x] Tests passing locally (`npm run test`)
- [x] `npx biome check --write` run on modified files

### Resources

- [AGENTS.md](https://github.com/pollinations/pollinations/blob/main/AGENTS.md)
- `shared/registry/text.ts`

### Reward

3

tomdacatto:
Fixes #15138

## Summary
- Adds `tencent/hy4-preview` as a new paid-only text model (no aliases), routed through OpenRouter, pinned to its only endpoint (`tencent/fp8` tag — Tencent is the sole provider)
- Pricing: $0.834/M input, $2.501/M output, $0.042/M cache read — verified live against OpenRouter's `/api/v1/models/tencent/hy4-preview/endpoints`
- Context 1,048,576, max output 64,000 (also verified live)
- `priceMultiplier: 1`, `paidOnly: true`

Currently #1 on OpenRouter's weekly usage rankings (17.8T tokens processed, +31% week-over-week), and Tencent wasn't represented in the catalog at all before this.

## Contract (confirmed)
| Field | Value |
|---|---|
| Canonical name | `tencent/hy4-preview` |
| Aliases | none |
| Registry provider | `openrouter` |
| Primary route | OpenRouter `tencent/fp8` tag (only endpoint) |
| Fallback | none found — single endpoint, no existing Tencent integration elsewhere in the repo |
| Pollinations GPU | no |

## Notes
- Forced `tool_choice` isn't supported by this endpoint (only `auto`/`none` per `supports_tool_choice`), so `tools` is declared without `tool_choice`, matching the existing convention for similarly-limited models
- No `top_p`, penalties, `seed`, or logprobs in the endpoint's `supported_parameters` — declared parameter set reflects exactly what's live
- Text-only input (no vision) on this endpoint
- Adds a placeholder Tencent brand logo (`frontend/public/brand-logos/tencent.svg`) since no verified official vector was available — required by `pricing-data.test.ts`'s catalog logo-resolution test, same convention already used for Thinking Machines and Sakana AI

## Verification
- Pricing/context/max-output/supported-params confirmed against OpenRouter's public endpoints API
- Added a unit test (`modelResolver.test.ts`) asserting the pinned `provider.only`/`allow_fallbacks` shape
- Updated the `effective-prices` snapshot in `enter.pollinations.ai`
- **Not run locally**: Vitest/CF-Workers suite and `tsc` fail environment-wide in this sandbox due to a pre-existing, unrelated dependency/workerd issue (same caveat as #14791). CI should be authoritative here.

### DIFF
--- a/enter.pollinations.ai/frontend/public/brand-logos/tencent.svg
+++ b/enter.pollinations.ai/frontend/public/brand-logos/tencent.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor">
+  <title>Tencent</title>
+  <path d="M12 3 21 19H3Z"/>
+</svg>
--- a/enter.pollinations.ai/frontend/src/components/models/model-info.ts
+++ b/enter.pollinations.ai/frontend/src/components/models/model-info.ts
@@ -36,6 +36,7 @@ const BRAND_LOGOS: Record<string, string> = {
     Sesame: "sesame",
     "Stability AI": "stability",
     StepFun: "stepfun",
+    Tencent: "tencent",
     "Thinking Machines": "thinking-machines",
     Xiaomi: "xiaomi",
     "Z.ai": "zai",
--- a/enter.pollinations.ai/test/__snapshots__/effective-prices.snapshot.test.ts.snap
+++ b/enter.pollinations.ai/test/__snapshots__/effective-prices.snapshot.test.ts.snap
@@ -2674,6 +2674,18 @@ exports[`effective cost and price per model — snapshot 1`] = `
       "promptTextTokens": 0.0000002,
     },
   },
+  "tencent/hy4-preview": {
+    "cost": {
+      "completionTextTokens": 0.000002501,
+      "promptCachedTokens": 0.000000042,
+      "promptTextTokens": 0.000000834,
+    },
+    "price": {
+      "completionTextTokens": 0.000002501,
+      "promptCachedTokens": 0.000000042,
+      "promptTextTokens": 0.000000834,
+    },
+  },
   "thinkingmachines/inkling": {
     "cost": {
       "completionTextTokens": 0.00000405,
--- a/gen.pollinations.ai/src/text/availableModels.ts
+++ b/gen.pollinations.ai/src/text/availableModels.ts
@@ -581,6 +581,10 @@ const models: ModelDefinition[] = [
         transform: createReasoningEffortTransform("toggle"),
     },
     {
+        name: "tencent/hy4-preview",
+        config: portkeyConfig["tencent/hy4-preview"],
+    },
+    {
         name: "thinkingmachines/inkling-small",
         config: portkeyConfig["thinkingmachines/inkling-small"],
     },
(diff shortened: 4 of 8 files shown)


ElliotEtag:
Two issues found in local E2E review (full matrix otherwise clean — streaming, tool-call round-trip, reasoning_effort, multi-turn, paidOnly gating, error handling, tests, tsc all pass):

**1. Missing the OpenRouter credit fee.** #14895 (merged 2026-09-15, after this branched) made the 5.5% fee mandatory on every OpenRouter registry cost. Needs:
```js
cost: {
    promptTextTokens: perMillion(0.834) * 1.055,
    promptCachedTokens: perMillion(0.042) * 1.055,
    completionTextTokens: perMillion(2.501) * 1.055,
},
```

**2. Strict `response_format: json_schema` silently returns `null` content under a normal `max_tokens` budget.** Default reasoning effort burns ~1,300 tokens on even a trivial 3-field schema, so any caller with `max_tokens` ≤500–1000 gets `finish_reason:"length"`, `content:null` — no error, just a blank billed response. This file already guards the identical failure mode for `qwen/qwen3.7-plus`/`qwen/qwen3.7-max`. Recommend the same here:
```js
"tencent/hy4-preview": () =>
    createOpenRouterModelConfig({
        model: "tencent/hy4-preview",
        defaultOptions: {
            max_tokens: 64000, // matches this route's own max_completion_tokens cap
            provider: { only: ["tencent/fp8"], allow_fallbacks: false },
        },
    }),
```

Non-blocking: `promptCachedTokens` billing is wired correctly, but this route reports `supports_implicit_caching:false` on OpenRouter — don't expect it to ever actually bill a cache hit.

ElliotEtag:
Follow-up on the `max_tokens` fix suggested above, verified empirically: it only protects callers who **omit** `max_tokens`. `modelResolver.ts` deletes the model's default the moment a caller supplies any explicit `max_tokens` (even a small one), so a caller passing e.g. `max_tokens: 500` still hits the same null-content-under-budget bug after this fix ships. Not a new gap — it's inherent to the identical `qwen3.7-plus`/`qwen3.7-max` guard already in this file — but worth knowing it's a partial mitigation (covers the "no max_tokens set" case) rather than a complete fix for every caller.

tomdacatto:
Fixed in d5a016d4e:
- Applied the mandatory 5.5% OpenRouter credit fee (#14895)
- Added `max_tokens: 64000` default (matching the existing `qwen3.7-plus`/`qwen3.7-max` guard) to stop the default-high reasoning trace from exhausting a normal budget. Noted in the config comment that this only covers callers who omit `max_tokens` — a caller-supplied value still hits the same gap, since `resolveModelConfig` drops the default the moment one is provided.

PR #14801, opened 2026-09-13
feat: add llm-pollinations provider plugin

jacobcreation:
Closes #14660

## What this adds

- Adds a standalone `llm-pollinations` Python package under `packages/llm-pollinations`.
- Reuses LLM's native OpenAI-compatible `Chat` and `AsyncChat` classes against `https://gen.pollinations.ai/v1`.
- Supports `llm keys set pollinations` and `POLLINATIONS_API_KEY`.
- Loads compatible text models from the authenticated `/v1/models` catalog as `pollinations/<model-id>`.
- Derives vision, reasoning, tool, and structured-output flags from catalog metadata.
- Uses a 15-minute cache with stale-cache fallback.
- Includes focused tests for catalog filtering, auth, caching, fallback, registration, and collision avoidance.

## Validation

`4 passed` with the package installed against LLM 0.35.


voodoohop:
- One improvement: handle discovery failures at registration so a temporary outage does not prevent other LLM providers from loading.

jacobcreation:
Thanks for reviewing and for selecting the winning implementation. Since #14801 was submitted for this quest before the selection and covered the core provider integration, could it still be considered for any runner-up bonus if one is available? No worries if the bonus list is final—just wanted to ask respectfully.

PR #14804, opened 2026-09-13
feat(floret): add MCP tools, 3D generation, and Quest routing

Itachi-1824:
- Depends on #14815; merge the agent-model header PR first.
- Adds Quest-only model selection through pollen or X-Pollinations-Pollen headers, preserving signed restrictions and default routing.
- Replaces Floret native shell containers with caller-authenticated Computer and FFmpeg MCP tools.
- Adds catalog-aware text/image-to-3D generation with hosted download links.
- Preserves gateway authentication, delegated billing, and Floret identity and capability guidance.


Itachi-1824:
!askpolli review concisely

voodoohop (commented):
- Please split this into two PRs: shared MCP tools + 3D generation, and Quest model selection. They solve different problems and can be reviewed separately.
- Replace the private shell with our shared Computer and FFmpeg MCP tools. Remove the old implementation; no backward compatibility or fallback shell is needed. Test the new path for file creation, media editing and publishing.
- Update from main first. #14815 now passes agent settings through `metadata`; use that instead of custom headers and aliases. No need to support the old header approach.
- Clarify what Quest mode promises. [The current code](https://github.com/pollinations/pollinations/blob/7880e708fca4f0a863599bf2ddbd64e7a91c536c/shared/schemas/pollen.ts#L3-L6) excludes paid-only models, but it can still spend purchased Pollen. It is not a paid-spending limit.
- Keep a model preference inside Floret: filter its available models and reuse the existing selector. Avoid adding signed-token, gateway and cache rules unless we actually need a strict restriction. If we do, enforce it once at the gateway and avoid repeating checks throughout Floret.
- Remove the [explanation rewriting](https://github.com/pollinations/pollinations/blob/7880e708fca4f0a863599bf2ddbd64e7a91c536c/apps/floret/model-catalog.js#L588-L593). Seeing a paid model's name does not let the agent use it.

Itachi-1824:
Closing this mixed PR as requested in review. I’ll replace it with two focused PRs based on current main: Quest model selection, and shared MCP tools + 3D generation.

PR #14883, opened 2026-09-14
feat: add polli update command for global CLI installs

davealan74:
Fixes #14879

## Summary
- Adds `polli update`, which delegates to `npm install -g @pollinations/cli@latest` for global installs
- Detects npx-cache and local/project installs and prints instructions instead of creating a second global install (no sudo, no silent updates, no generic updater framework)
- Preserves existing output conventions: human messages on stderr, structured `--json` result on stdout, non-zero exit on failure
- Adds `polli update` to README and the CLI skill quick-reference table

## Test plan
- `npx vitest run` in `packages/polli-cli` — all 116 tests pass (5 new for `detectInstallKind`)
- `npx tsc --noEmit` — no new errors in touched files
- `npx biome check --write` on touched files — clean
- Manually verified end-to-end against the real npm registry using isolated npm prefixes (no interference with any real global install):
  - Packed the CLI, installed it globally into an isolated `--prefix`, pinned to an old version, ran `polli update` — it detected the global install, ran `npm install -g @pollinations/cli@latest`, and confirmed the prefix now has the latest published version
  - Installed the CLI the way npx does (into a `_npx` cache-style directory) and ran `polli update` — printed the npx guidance, made no changes
  - Ran the CLI straight from a local checkout (not under `npm root -g`) and ran `polli update` — printed the local-install guidance, made no changes

PR #14903, opened 2026-09-15
feat: add polli update command

### LINKED ISSUE #14879: [QUEST] Add a lightweight polli update command
### Goal

Make it easy for Polli CLI users to know when a newer version is available and update to it.

### Outcome

- Users receive an unobtrusive update notice and have a straightforward way to install the latest stable version.
- Updating preserves their credentials, configuration, and harness settings, and works sensibly with the documented installation methods.
- Update checks do not noticeably slow normal use, disrupt scripts or machine-readable output, or cause commands to fail when offline. Users can disable notices; updates are never installed silently.
- Include concise documentation and demonstrate the update experience with focused tests.

Keep the solution small and easy to maintain. The command design, wording, and implementation are up to the contributor; the references below are examples, not requirements.

### References

[CLI source and installation docs](https://github.com/pollinations/pollinations/tree/main/packages/polli-cli) · [Gemini CLI update pattern](https://geminicli.com/docs/resources/faq/) · [npm global updates](https://docs.npmjs.com/updating-packages-downloaded-from-the-registry/)

[GitHub CLI update notices](https://cli.github.com/manual/gh_help_environment) · [Node update-notifier](https://github.com/sindresorhus/update-notifier)

### How this quest works

Submit a focused PR with `Fixes #14879`. We compare working solutions for correctness, clarity, and minimal code. The selected merged PR author receives the reward.

### Reward

6 Pollen

kreggscode:
Fixes #14879

## Summary

Adds a `polli update` command that checks for and installs the latest version.

## What it does

- **Detects install method** from process paths: global (npm -g), npx, or local
- **Checks npm registry** for latest version, compares to running version
- **Global installs**: updates in place via `npm install -g @pollinations/cli@latest`
- **npx**: next `npx @pollinations/cli` invocation uses latest automatically
- **local**: shows `npm update @pollinations/cli` advice
- **`--check` flag**: non-interactive version check (no install)
- Uses existing output helpers - works with `--json` flag

## Files

- `src/commands/update.ts` - the command (85 lines)
- `src/commands/update.test.ts` - 9 focused tests
- `src/index.ts` - registers the command

## Tests (9 passing)

- Version comparison: equal, patch, minor, major, different segment counts
- Install kind detection: npx from exec path, npx from argv, local from node_modules, global default

## Key design decisions

- **Preserves credentials**: only runs `npm install -g`, never touches `~/.pollinations/`
- **No silent updates**: always shows what is happening
- **Offline-safe**: fails gracefully when npm registry unreachable
- **Script-friendly**: `--check` flag for CI/automation, `--json` support
- Zero new dependencies - uses Node.js built-in `child_process`


voodoohop:
- Thanks for taking part! We merged #14883 with follow-up fixes and update notices.\n- The packaged CLI here failed to start because package.json was read from the wrong path. Testing the built CLI would catch this; the tests currently copy the logic.\n- Closing this competing PR.

Issue #14821, opened 2026-09-13
[QUEST] Build an NPC who remembers


voodoohop:
### Goal

Create a fun NPC that remembers you between conversations. You choose the NPC type, theme, personality, and setting. Make it a small example others can copy and customize.

> **Alpha:** Pollinations agents are still in alpha. Documentation improvements, reports of usability problems, and suggestions for making agents easier to build and use are very welcome. Please share them in this issue or your PR.

### Outcome

- The character remembers facts you explicitly ask it to keep, even in a fresh conversation without the old chat history. It can show its memories and forget them on request.
- Reuse the existing Computer MCP for a small memory file in a dedicated folder. Keep users' memories separate and leave unrelated files untouched. Prefer a prompt agent if sufficient; no new memory service, vector database, or custom frontend.
- Demonstrate remembering across fresh chats, showing/deleting memories, and isolation between two users.

### Submission

Test the agent in your own public GitHub repository. Submit it under `apps/<agent-name>/`. For a prompt agent, just include `agent.json` with its system prompt and settings (base model and MCP servers); no code or scaffolding needed. For a code agent, include the agent code and concise setup instructions. The PR description must include your public repository link and the deployed agent ID/callable model name so we can try it.

### Resources

[Scalar: Community Agents API](https://gen.pollinations.ai/docs#tag/community-agents) · [Example agent with MCP tools](https://github.com/pollinations/pollinations-code-agent-example) · [Example router agent (code-only)](https://github.com/pollinations/pollinations-router-agent)

[Publish an agent](https://github.com/pollinations/pollinations/blob/main/BUILD_YOUR_OWN_AGENT.md) · [Computer MCP](https://github.com/pollinations/pollinations/tree/main/apps/computer-mcp)

### How this quest works

Submit a focused PR with `Fixes #14821`. We select the clearest, smallest working solution; the selected merged PR's author receives the reward.

### Reward

15 Pollen

ammarelshaf3y:
Opened a PR: https://github.com/pollinations/pollinations/pull/14884 - Prompt agent at apps/memory-fox-npc/agent.json (baseModel openai, mcpServers [computer]) playing Sahara, a desert-fox merchant from Siwa Oasis. Memory is one file /workspace/memory-fox-npc/memories.md via the Computer bash tool: loaded first every turn, appended on remember, listed on show, sed-deleted on forget, truncated on forget-everything. Isolation is automatic per-caller Computer. node test.mjs 5/5 green.

[...]

PR #14907, opened 2026-09-15
feat: add Barnaby the Crow, an NPC agent who remembers (#14821)

Guest453:
Fixes #14821

## 🐦‍⬛ Barnaby the Crow — an NPC who remembers you

Barnaby is a cheeky crow shopkeeper at the edge of the Whispering Market. He hoards "shinies" — the little facts you tell him about yourself — in a battered tin box under his counter. Crows genuinely remember faces, so memory is the character, not a bolt-on.

### What this PR adds
- `apps/barnaby-the-crow/agent.json` — the **entire** agent. A prompt agent: one system prompt + `baseModel: "openai"` + `mcpServers: ["computer"]`. No code, no new service, no vector DB, no frontend.
- `apps/barnaby-the-crow/README.md` — the deploy line, a demo, and how to re-theme / rename it.

### How the memory works
- One file in a dedicated folder: `/workspace/barnaby-the-crow/memories.md`, via the existing **Computer MCP** `bash` tool.
- Every turn opens the box (`cat`). `remember` appends one line — the fact is passed in the tool's `stdin`, the MCP-documented way, so there are no shell-quoting bugs. `forget <x>` deletes matching lines (`sed -i`). `forget everything` truncates. Nothing outside the folder is read or written.
- Isolation is free: each caller gets their own Computer Durable Object, so two users never share a box.

### Verified

**A) Local harness against the real Computer MCP** — the agent's system prompt + the real `bash` tool, every turn a *fresh* chat with no history → 7/7:

```
USER: remember that I like mint tea
  bash> cat /workspace/barnaby-the-crow/memories.md 2>/dev/null
  bash> cat >> /workspace/barnaby-the-crow/memories.md   [stdin: '- likes mint tea']
  BARNABY: Kaa! ... I've tucked that shiny away: you like mint tea.
USER: what do you remember?
  bash> cat /workspace/barnaby-the-crow/memories.md   -> '- likes mint tea'
  BARNABY: ... just one shiny: **you like mint tea.**
USER: please forget that I like mint tea
  bash> sed -i '/likes mint tea/d' /workspace/barnaby-the-crow/memories.md
  BARNABY: Consider it forgotten.
box after forget:  ''

[PASS] fact saved to the tin box
[PASS] recall works in a FRESH chat
[PASS] forget-one removes the fact
[PASS] second fact saved
[PASS] forget-everything empties the box
[PASS] another user's box (/workspace/someone-else) never leaked
[...]

PR #14911, opened 2026-09-15
feat(apps): add ember-kiosk, an NPC who remembers

Marcus-Mok-GH:
Ember is a small dragon who runs a corner coffee kiosk and remembers regulars through a chalk-slate ledger in a dedicated folder, using the Computer MCP `bash` tool.

- Prompt agent: `apps/ember-kiosk/agent.json` only (`systemPrompt` + `baseModel: openai` + `mcpServers: [computer]`) — no code, no new service, no vector DB, no frontend
- Memory: one file, `/workspace/ember-kiosk/ledger.md`. Opened every turn; append on explicit remember; `sed` line-delete on forget; truncate on forget-all; the agent never writes outside its dedicated folder
- Per-user isolation: each caller's Computer is a private Durable Object, so two users never share a ledger
- README doubles as the copy-and-customize guide (change name, folder, voice)
- Verified: the exact protocol commands (first open / remember / recall / forget / forget-all) were run against two isolated user filesystems — all correct; `apps/computer-mcp` vitest suite passes 14/14 including a real bash exec
- Public repository: https://github.com/Marcus-Mok-GH/pollinations
- Callable model name: `Marcus-Mok-GH/ember-kiosk` (one-command create in the README)

Fixes #14821


Marcus-Mok-GH:
Memory protocol verified against the real Computer MCP (wrangler dev, per-user Durable Objects, JSON-RPC tools/call over HTTP):

- remember x2 → two separate ledger lines (after trailing-newline fix, commit 65b9873)
- fresh-chat recall → cat shows each fact on its own line
- forget-one → sed removes only the matching line
- forget-all → truncation leaves an empty file
- isolation → a second user's ledger is absent; cross-user bleed not observed
- unrelated files → /workspace files outside /workspace/ember-kiosk/ untouched

One fix landed from that verification: the remember stdin value now ends with a newline, otherwise consecutive remembers merged into one physical line (corrupting recall and let sed delete both facts at once).

voodoohop:
Thanks for the feedback, it went into the agent guide (#14964). A 3 Pollen bonus is waiting for you at https://enter.pollinations.ai/quests.

voodoohop:
Closing: #14907 was selected for #14821 (reasoning on the issue). Thanks for the submission.
- Same structure as #14907, opened after it; the selected PR also carried the cross-conversation verification.
- A rule from your prompt went into the merged agent. A 3 Pollen bonus is waiting at https://enter.pollinations.ai/quests.

PR #14912, opened 2026-09-15
feat(auth): surface key management link on 403 model permissions

### LINKED ISSUE #14814: 🔗 Show key-permissions link when a model isn't allowed (not just when budget runs out)
- Budget-exceeded errors already point users to their key settings (`Get one at https://enter.pollinations.ai/keys`, `gen.pollinations.ai/src/middleware/auth.ts:44-45`). Model-not-allowed errors don't — `requireModelAccess()` (`gen.pollinations.ai/src/middleware/auth.ts:66-77`) throws a 403 with no link, even though it already has the model + key in scope.
- Add the same keys link to the model-not-allowed error — both for the primary model on a request, and for MCP tool calls where an agent hits a Pollinations model the key isn't scoped for (MCP could return the link alongside the missing-model error).
- Future: replace the generic /keys link with a dialog that asks for only the specific missing permission — needs a deep-link (e.g. `?model=`) on the keys page, which doesn't exist yet.

fisventurous:
surfaces key management URL when requests fail due to model permission limits

closes #14814

\- append `https://enter.pollinations.ai/keys` to 403 error in gateway `requireModelAccess`
\- append same key management link to 403 error in realtime route `requireAllowedModel`
\- update gateway model permissions tests for new error message format
\- surface key management link in MCP `parseApiError` 403 handler without duplicating if already present
\- add unit tests for MCP 403 error parsing


voodoohop (commented):
- Link directly to `/edit-key?id=<key-id>` using the authenticated key's ID, rather than `/keys`. #14740 already added this editor, including model permissions, so no new page is needed. Use the staging host for staging requests.
- In [the MCP 403 handler](https://github.com/pollinations/pollinations/blob/c38fdfe0eac760f4b7ca87465f5acf29ed3f27dc/packages/mcp/src/utils/coreUtils.js#L187-L195), display the gateway's error message unchanged, with a generic fallback when empty. Remove the phrase matching and automatic link insertion: banned accounts and staging access restrictions also return 403, and editing key permissions will not fix those.
- Update the tests for the direct editor link and MCP passthrough, including a non-permission 403. Keep the link in the gateway error so all clients benefit.

fisventurous:
!askpolli review the changes

voodoohop:
- Pushed be9aef9 on top: `keyPermissionsLink` now takes only the key id + `ENVIRONMENT` (set per wrangler env), dropping the request-hostname sniff and the unreachable `/keys` fallback
- Local: gen `model-permissions` 18/18, MCP `coreUtils` 4/4, tsc + biome clean
- Merging once CI is green. Thanks!

Issue #14234, opened 2026-09-01
Measure login and auth-flow drop-off


voodoohop:
## Goal

Measure what percentage of people enter an authentication flow but do not complete it, and where they leave.

## Scope

Cover every user-facing flow:

- GitHub sign-in and new registration
- App OAuth/PKCE authorization
- CLI/device authorization
- Account linking

Use one privacy-safe `flow_id` across redirects. Track `flow_type`, `stage`, `result`, timestamp, and a non-sensitive failure reason. Never record authorization codes, tokens, email addresses, or credentials.

Suggested stages: `viewed → started → provider_redirected → authenticated → authorized → completed`, plus `cancelled` and `failed`. Report both screen bounce (`viewed` without `started`) and flow abandonment (`started` without completion within 30 minutes). Distinguish new registrations from returning sign-ins.

## Output

- Completion and abandonment percentage by flow
- Drop-off at each stage
- Breakdown by provider/client and failure reason
- 7-day and 30-day trends

## Implementation direction

Reuse Tinybird/ClickHouse with one private `auth_flow_event` datasource and a small dashboard. Emit server-side events where possible and use only the minimum browser event needed to measure the initial view/click.

Better Auth Infrastructure tracks successful sign-ups and sign-ins, but does not measure the abandonment before them and would not cover our custom OAuth/device flows. PostHog and Plausible offer ready-made funnels but add another analytics vendor and client tracker, so the existing Tinybird stack is the simplest fit.

PR #14924, opened 2026-09-15
feat: add minimal PostHog conversion tracking

voodoohop:
Competing alternative to #14938 (Tinybird-native). Choose one; do not merge both. PostHog supplies a funnel UI; Tinybird avoids SDK dependencies and supports direct joins.

## Summary

- Adds official browser/Workers SDKs with four events: signup_completed, signed-in page_viewed, checkout_started, and payment_completed.
- Joins events with the existing user ID; reuses payment fulfillment/deduplication and sends actual charged minor units plus currency. Best-effort delivery; not a revenue source of truth.
- Keeps browser identity in memory; excludes URLs, referrers, person enrichment, prompts and email. No autocapture, replay, generation firehose, Tinybird export or custom dashboard.
- Leaves tracking disabled until configured. Automatic top-ups and anonymous login abandonment are out of scope.

## Activation

After project/privacy approval, configure POSTHOG_PROJECT_TOKEN and POSTHOG_HOST on the Worker, and VITE_POSTHOG_PROJECT_TOKEN / VITE_POSTHOG_HOST for frontend builds, using the same project. Separate staging/production projects. Confirm consent, retention/deletion and Discard IP settings before enabling. No credentials or deployments changed.

## Verification

- 80 targeted tests pass; typecheck, frontend build, staging Workers build and scoped Biome checks pass. Covers privacy, failures, unpaid checkout and replay.

Partially addresses #14234; auth-flow abandonment remains open.


ElliotEtag:
!polli what do you think of this review?

Review of [PR #14924](https://github.com/pollinations/pollinations/pull/14924) is complete. The code ships disabled, CI is green, and the four PR tests pass locally. Two demonstrated problems are worth fixing before merge, and two activation blockers are missing from the PR body.

Demonstrated problems

The browser SDK ships to every visitor even when analytics is off. posthog.ts:1 imports posthog-js statically and the component uses it, so the token check only skips init. Measured with a production build and no token configured:
Client JS, gzip	main	PR
total	388 KB	500 KB

posthog-js alone is 104 KB gzip on the entry chunk. The browser side sends one custom event with three properties, nearly every SDK feature is switched off, and the allowlist at posthog.ts:25 exists only to undo SDK defaults. Smallest fix: replace the SDK with a fetch POST to PostHog's public capture endpoint /i/v0/e/ carrying api_key, event, distinct_id, and properties. That removes the dependency, the allowlist, and the next finding. If the SDK stays, load it with a dynamic import inside the component so disabled builds and signed-out visitors never download it.

Every signed-in page load merges a throwaway anonymous id into the person. With memory persistence each load starts with a fresh anonymous id. analytics.tsx:25 then calls reset() and identify(userId), which emits $identify with that id as $anon_distinct_id. PostHog's [persistence docs](https://posthog.com/docs/libraries/js/persistence) warn about exactly this and recommend bootstrapping the id instead. Effect: one extra billed event per page load and unbounded distinct ids per person. One-line fix if the SDK stays: add reuseAnonymousId: true to the config, which omits $anon_distinct_id. Nothing is captured while anonymous, so there is nothing to merge.

Open questions and activation blockers

The published privacy policy says the opposite. PRIVACY_POLICY.md:54 states "We do not use third-party analytics or marketing cookies/SDKs at this time. If we add non-essential cookies or SDKs, we will request consent first." Do-not-track handling is not consent. Line 58 also promises 14 days' notice before sub-processor changes. Activation needs a policy update plus a consent gate, or the browser side dropped and only server events kept. The PR's activation section should say this explicitly.
The linked issue chose Tinybird over PostHog. Issue [#14234](https://github.com/pollinations/pollinations/issues/14234) says PostHog "adds another analytics vendor and client tracker, so the existing Tinybird stack is the simplest fit". Signups already land in the d1_user datasource and payments in stripe_event, so only checkout_started is new data. If PostHog is now the decision, record it on the issue rather than partially reversing it.

Verification

Ran the new unit test file and the two webhook PostHog tests locally. All pass, including unpaid session and replay deduplication.
Source-checked posthog-node: captureImmediate sends once with no retry, and the constructor starts no timers without a personal API key, so the per-call client in posthog.ts:17 is safe. Server events carry a distinct id, so they land on the same person as browser events.
Gaps: no test covers signup_completed or checkout_started, and the browser identify and page-view logic has no test at all.

voodoohop:
- Closing in favor of #14938 (Tinybird-native), per the direction recorded on #14234: no new analytics vendor or client tracker.
- Blockers this PR would have needed before activation: the privacy policy states no third-party analytics SDKs and consent first ([PRIVACY_POLICY.md#L54](https://github.com/pollinations/pollinations/blob/main/pollinations.ai/public/legal/PRIVACY_POLICY.md#L54)) plus 14 days' notice for sub-processor changes (line 58); the static SDK import ships to every visitor even when disabled; and the reset-then-identify on each signed-in load merges a fresh anonymous id per page view.

PR #14931, opened 2026-09-15
feat(gen): add ?source=official|community|all query filter to model-list endpoints

mhmdrizzzki:
## Summary

Adds a `?source=` query filter to Pollinations model-list endpoints so clients can filter built-in (official) models separately from community models — alongside the existing `?community=` flag.

### Feature
- `?source=official` — returns built-in models only (no community endpoint)
- `?source=community` — returns community models only
- `?source=all` / omitted — returns everything (default)
- `?source=` wins over the legacy `?community=` when both are sent (single intuitive interface)

Applied to all model-list handlers using `ModelListQueryParams` (`/v1/models`, `/models`, etc).

### Why
Replaces the boolean `community` flag with a clearer three-way source switch, aligned with the quest (#14535) goal of filtering listings by source in a more intuitive way. No generic filtering framework; just a focused query-param addition.

### Tests
`test/models-source-filter.test.ts` — 7 cases: all source values, legacy `community` compatibility, and source-wins-over-community precedence. All pass (`npm test` green).

Fixes #14535


voodoohop:
Thanks for the focused source filter. That part has already shipped; this PR does not cover the quest's reliability and health pieces. Closing as superseded.

PR #14937, opened 2026-09-15
feat: add virtual My Models backed by fallback sequences

### LINKED ISSUE #13632: feat: add virtual My Models backed by fallback sequences
- **Problem**
  - “My Models” currently only represents external endpoints and requires a URL, upstream model, credential, pricing, and payout identity.
  - The gateway already retries ordered fallback candidates, but users cannot save existing catalog models behind one reusable model ID.

- **Proposal**
  - Add a private virtual My Model defined by ordered `modelIds`.
  - The first selected model is the primary and supplies the quoted price, balance class, modality, endpoints, capabilities, and catalog metadata.
  - Remaining models are tried in order using the existing retryable-status, network-failure, and content-policy rules.
  - The My Models UI should select, reorder, and remove models without endpoint or pricing fields.

- **V1 constraints**
  - Allow visible static, public community, and owner-private concrete models.
  - Disallow nested virtual models; cap at one primary plus three fallbacks.
  - Require compatible event type/request surface.
  - Require each fallback price to be no higher than the primary and not introduce `paidOnly`.
  - Keep sequences private with no markup or owner reward.
  - Bill the primary’s quoted price; record the actual served model and fallback status.

- **Implementation notes**
  - Use a first-class sequence/router record; do not fake `community_endpoint`, whose endpoint and credential fields are non-null.
  - Project sequences into the runtime registry and reuse the shared fallback loop.
  - Revalidate targets when loading the registry; disable a sequence whose primary disappears and skip ineligible fallbacks.

- **Acceptance**
  - A user can create `owner/name`, call it through normal generation endpoints, edit its order, restrict a key to it, and see requested-versus-served models in usage data.
  - Add routing, billing, permissions, cycle-prevention, and catalog tests.

- **Related**
  - #12854
  - #12859

afanasevmylife:
Closes #13632

## What

Virtual **My Models** (model sequences): an ordered list of existing model ids served under a private model id `<github-username>/<name>`. The first id is the **primary** - it supplies the quoted price, balance class, modality, endpoints, and capabilities. The remaining ids are **fallbacks** tried in declared order through the existing shared fallback loop (`FALLBACK_TARGET_HEADER`, depth-one linking).

## Design constraints (V1)

- **Owner-private**: sequences appear in catalog and resolve at request time only for the owner; everyone else gets the same `Invalid model or alias` 400 as an unknown name.
- **No nesting**: a sequence can never be another sequence's target (enforced both at write time in enter and at registry projection in gen).
- **Caller protection**: a fallback must share the primary's event type, must not cost more per unit than the primary (rates compared per usage type after each definition's price multiplier), and must not require Paid Pollen the primary doesn't - callers are always billed the primary's quoted price whichever model answers.
- **Limits**: 1 primary + up to 3 fallbacks.
- **Id-space safety**: `<owner>/<name>` is checked against bundled models/aliases, legacy community aliases, and the owner's community endpoints (and vice versa - community endpoints now reject names taken by sequences).

## Changes

- `shared/db/better-auth.ts` + `enter.pollinations.ai/drizzle/0067_model_sequences.sql`: `model_sequence` table (owner, unique name per owner, JSON `model_ids`).
- `shared/model-sequences.ts`: id format/parse helpers, event-type mapping, pricing and balance eligibility rules (+ unit tests).
- `enter.pollinations.ai/src/routes/model-sequences.ts`: CRUD API mounted at `/account/my-models/sequences` (session + API key with `account:keys`). Resolution is static-first because bundled ids contain a slash; community targets follow the existing concealment rules (private/hidden models of others look missing).
- `gen.pollinations.ai/src/model-registry.ts` + `model-sequences.ts`: registry projects each row into a `GenerationModelEntry` cloned from its primary, links validated fallback copies (`sequenceTarget: true`, own fallbacks stripped for depth one), loaded with the same degraded-mode tolerance as community entries. A sequence whose primary becomes unusable is disabled, not partially served.
- `gen.pollinations.ai/src/middleware/model.ts`: owner gate on resolution; sequence fallback targets inherit the sequence's key permission (`sequenceTarget === true` bypasses per-target `allowedModels`, same as `fallbackOnly` providers).
- `shared/registry/visible-model-ids.ts`: key permission validation accepts sequence ids; visible id set includes the owner's sequences.
- Frontend: new **Sequences** section in My Models with ordered model picker (primary + reorderable fallbacks), edit and delete dialogs.

## Tests

- `shared/model-sequences.test.ts`: 7 unit tests (id round-trip/parse, pricing matrix, balance matrix).
- `gen.pollinations.ai/test/model-sequences.test.ts`: 8 registry tests (projection/price clone, linking order + depth one, owner-only catalog, non-owner 400, disabled primary, pricing/paidOnly/event-type drops, no nesting).
- `gen.pollinations.ai/test/model-sequences-api.test.ts`: 7 API tests (CRUD, all rejection rules, 401).
- Regressions green: community-endpoints (178 tests incl. fallbacks), model-registry, model-permissions, models-retrieve, community-model-rate-limit; `tsc` clean for gen + enter (incl. frontend); biome clean on changed files.


afanasevmylife:
@strix-security re-run review please - this PR has never had a Strix pass, and it touches auth-adjacent model resolution paths.

afanasevmylife:
@strix-security re-run review please - new head 15bbe52d4 (merged main, fixed the registry KV-cache gap your future pass should see in gen.pollinations.ai/src/model-sequences.ts), CI fully green.

afanasevmylife:
Update in f0ddd22 (CI fully green):

**Independent cross-model review fix.** A fresh-eyes review of this PR's diff by another model family caught a latent crash in the KV-cache path added here: `model-sequences:v1` rows come back from KV as JSON, so `createdAt` is a string, and the registry projection (`addedDate: row.createdAt.getTime()`) would throw `TypeError: row.createdAt.getTime is not a function` on every KV-served load. No test covered the sequence KV-hit path, so CI could not see it.

- fix: `getModelSequenceRegistryRows` now revives `createdAt` into a `Date` on cache hits
- test: new case seeds the cache, then loads the registry through a skewed D1 binding (same pattern as the community-catalog KV test) and asserts the sequence entry resolves with a finite `addedDate` - red with exactly that TypeError without the fix, green with it

afanasevmylife:
## Independent cross-model review (round 2): APPROVED

Reviewed by `openai/gpt-6-astra` (different model family than the author) against the fix commits `697f416..a6a9fcc`, after round 1 returned REVISE with 5 findings.

**Round-1 findings and dispositions:**

1. **HIGH - community primary lost its runtime** (`sequenceRowToGenerationEntry` dropped `primary.communityEndpoint`, so a community-primary sequence took the static-provider path with an unknown id and the first attempt could never serve). Fixed: entry now carries `communityEndpoint: primary.communityEndpoint`.
2. **MEDIUM - no regression test for the above.** Fixed: `keeps the community primary's runtime so the first attempt can serve it` in `gen.pollinations.ai/test/model-sequences.test.ts` (asserts `communityEndpoint.upstreamModel` survives projection and the nano fallback target survives the bucket-by-bucket no-uphill-fallback rule; fixture prices every usage bucket above nano's rates, including cached tokens).
3. **HIGH - nested-sequence existence oracle**: the "is itself a model sequence" 400 confirmed the existence of other owners' owner-private sequences. Fixed: the nested lookup now only runs when the resolved endpoint owner is the caller (`owner.id === ownerUserId`); other owners get the generic missing-model response. Stale comment claiming the opposite removed.
4. **LOW - same-owner cross-table name race** (sequence name vs community endpoint name): accepted as a documented residual, recoverable by rename; deliberately not fixed.
5. **MEDIUM - ms timestamps in migration 0067** (`integer(mode: "timestamp")` is seconds here; precedent `0030_fix_reward_timestamp_units.sql`). Fixed: `*86400000` → `*86400` in `0067_model_sequences.sql` and in the `model_sequence` block of `0067_snapshot.json` only (other tables' defaults untouched).

**Round-2 verdict: APPROVED** - no blocking defects; four LOW non-blocking notes. The one verification boundary the reviewer flagged (consumers of an entry carrying both `communityEndpoint` and `modelSequence`: billing attribution, gateway context, `fallbackCandidates` pairing) was checked by the coordinator against `middleware/model.ts` (owner gates for both flags are independent and both pass for the owner; `resolved` carries both), `fallback.ts` (`fallbackCandidates` passes the paired runtime through for attempt 1 and all targets), and the community gateway context (runtime carries baseUrl/upstreamModel/token unchanged). No new defect.

**CI:** all green on `135e4d10` (Test gen 2245 passed, Test enter, Lint and build, typecheck). Two intermediate failures during the fix loop were diagnosed and fixed (`communityModelId` import, fixture pricing buckets); one failure cluster on `c1f8a13` (elevenlabs timestamps timeout, model-permissions timeouts, models-retrieve health counter off-by-one, model3d event pollution from another test file) was confirmed flaky and unrelated - clean re-run passed.

This PR has never had a Strix review pass - could a maintainer trigger one when convenient?

PR #14951, opened 2026-09-16
feat(ui): shared account identity, app user menu, and wallet status

ElliotEtag:
- Avatars link to the dashboard everywhere: on Enter's authorize, edit-key, and top-up headers and inside the app user menu, where the name and chevron open the menu. Menus without a destination keep the whole pill as the trigger.
- App sign-in button reads "Pollinations Connect": amber cell with the Pollinations mark, light label cell, amber border. One style only; the `triggerVariant` prop is removed.
- App user menu: Permissions, Buy Pollen, Disconnect. Quests shortcut, `onTopUpKey`, and the old label keys are removed. After following Permissions or Buy Pollen, the menu re-reads the key and wallet when the app regains focus, so the pill reflects the change without a reload.
- Second line of the pill: a budgeted key shows its own budget ("3.25 pollen") or a red Limit reached badge; an unlimited key shows the account's quest and paid balances when the key has the usage scope, otherwise a blue Unlimited badge.
- Enter's authorize, edit-key, and top-up headers show quest and paid balances, or a red No Pollen · Top up badge when the wallet is empty (no link on the top-up page itself). The authorize screen's separate out-of-Pollen alert is removed; the header carries it.
- Account names use the display name, then the GitHub handle, everywhere.
- Account menus size to the pill and their content; the balance line never truncates.
- `/account/key` returns the key's opaque `id`; SDK `KeyInfo.id` added.
- New shared `AccountPollen` in `@pollinations/ui/wallet`; new `PowerIcon`.
- Not included: paid-only signalling on the authorize screen and the connection error/hydrating states (later PRs).


voodoohop (commented):
The shared account pill works well. A few things before merging:

**Behaviour**
- Buy Pollen / Permissions send the user back to the app's path without its query or hash (`AppUserMenu.tsx` `returnUrl`), so `/?room=abc` returns to `/`.
- An unlimited key with `account:usage` shows "Unlimited" until the wallet loads, then switches. If the wallet request fails, it stays "Unlimited".
- The top-up page shows the same balances twice: in the header and in the Wallet section right below it.
- On authorize, `topUpHref="/top-up"` has no `redirect`, so after paying the user isn't sent back to the authorize flow.

**Simpler**
- Drop the `null` = "failed" state. It renders "… …", which looks like loading forever. Treat failure like loading (render nothing).
- Drop the `Number.isFinite` checks and the NaN/Infinity tests. The values come from JSON, which can't contain either.
- Keep the refresh when the user returns to the tab, but make it simpler: one `visibilitychange` listener, with no `returning` ref, no focus listener and no `expectReturn()` wrappers. That's about 10 lines:
  ```tsx
  useEffect(() => {
      if (!isLoggedIn) return;
      const onVisible = () => {
          if (document.visibilityState !== "visible") return;
          void key.refresh();
          void balance.refresh();
      };
      document.addEventListener("visibilitychange", onVisible);
      return () => document.removeEventListener("visibilitychange", onVisible);
  }, [isLoggedIn, key.refresh, balance.refresh]);
  ```
  It refreshes on every return to the tab, which is two cheap GETs, and stale responses are already ignored.
- The balance fetch is repeated in authorize and edit-key (with different error handling), and the `tierBalance`/`packBalance` → `quest`/`paid` mapping is repeated 4×. Use one hook, and let `AuthAccountIdentity` take the API response directly.
- Render `AccountPollen` once in `AppUserMenu`: pick the `source` first, no nested ternary.
- Merge `PollenStatusBadge` into `AccountPollen` (its only caller). Don't export `AccountAvatar`/`AccountDetails`. `Balance` has no users left.
- `AccountMenu` builds its own avatar with the same `large={secondaryContent != null}` rule as `AccountIdentity`. Keep that rule in one place.

**Tests / API**
- `AppUserMenu.test.tsx` only checks for CSS class strings. Test behaviour instead: Permissions shows only when the key has an `id`, and budget vs. unlimited vs. no-usage-scope pick the right badge.
- Removing `triggerVariant`, `onTopUpKey` and the `getFreePollen`/`topUpAccount` labels breaks the public `AppUserMenu` API. Add a changelog entry or version bump for outside users.

The branch is 37 commits behind main, so rebase before merging.

PR #14959, opened 2026-09-16
feat: record gross Stripe amount on paid auto top-up attempts

voodoohop:
## Summary
- Adds `stripe_amount_paid` (minor units) and `stripe_currency` to `stripe_auto_top_up_attempt` (migration `0067`, two nullable columns)
- The `invoice.paid` handler stores both from the invoice when it marks the attempt paid; `amount_usd` stays the pack principal, so principal vs gross payment can be reconciled from D1 instead of Stripe exports
- Extends the trigger + webhook test to assert the stored gross amount (`1000 + service fee`) and currency

Follow-up to the spending-analysis notes in #14952 ("auto-top-up: principal vs gross payment").

## Test plan
- [x] `npx vitest run test/integration/stripe.test.ts` — 75 passed
- [x] `drizzle-kit generate` output is only the two `ALTER TABLE … ADD` statements
- [ ] After deploy: `SELECT amount_usd, stripe_amount_paid, stripe_currency FROM stripe_auto_top_up_attempt WHERE status='paid' ORDER BY completed_at DESC LIMIT 5`


voodoohop:
Closing: gross auto-top-up amounts already reach Tinybird via payment_intent.succeeded (179 rows match the 179 paid attempts in D1 over 60 days). The gap is user attribution there, not missing D1 columns. Skill text referencing these columns is reverted in a follow-up PR.

PR #15013, opened 2026-09-16
feat: refund small issuer-flagged charges before they become disputes

ElliotEtag:
Stripe charges a flat **$20 dispute fee** whatever the amount, so a disputed $2 pack costs us $22. Measured on our own account: of 17 early fraud warnings, **11 became a dispute and 4 were refunded in time — none of those four escalated.** Eight warnings arrived on 5 September and all eight had become disputes within eleven days.

Refunding a warned charge worth less than the fee is cheaper than waiting, even when the warning turns out to be wrong.

- Handles `radar.early_fraud_warning.created`. Refunds the charge with reason `fraudulent` and reverses the Pollen it bought, which can leave a negative balance when already spent.
- Only below `STRIPE_DISPUTE_FEE_CENTS` ($20). Giving away a larger payment on one warning is worse than paying the fee if it turns out to be a real customer, so those stay for a person to judge.
- Skips charges already refunded, disputed, or never captured. Re-reading the charge makes a Stripe webhook retry a no-op, so a balance is never reversed twice.
- Ownership comes only from metadata our server wrote, never the buyer's email. A charge with no owner is still refunded; no balance is touched.
- Runs inside `waitUntil`, so a slow Stripe call cannot make the webhook time out.

**Needs a config change to take effect:** our webhook endpoint subscribes to 16 events and `radar.early_fraud_warning.created` is not one of them. It must be added in the Stripe dashboard after deploy.

Done by hand today for the two open warnings, which is what this replaces. 5 new tests; the 75 existing Stripe integration tests still pass.


ElliotEtag:
Superseded by #15015, which does this refund and more. Two details from here are worth carrying over into that PR:

- **Cap the auto-refund at the $20 dispute fee.** Below it, refunding is cheaper than the fee we would pay anyway. Above it, giving away a real customer's payment on a single warning is worse than paying the fee. Without a cap, an actionable warning on a $103.80 charge refunds in full.
- **Use an idempotency key on `refunds.create`.** Stripe retries webhooks, and a retry that arrives before the charge reads as refunded would create a second refund.

Issue #15016, opened 2026-09-16
📊 [QUEST] Filter polli usage by key, model and day range


voodoohop:
### Goal

Let `polli usage` answer "what did this key / this model cost?" without leaving the CLI.

### What exists today

- `polli usage --history` and `--daily` call `/account/usage` and `/account/usage/daily` with only `limit` (`packages/polli-cli/src/commands/usage.ts`). The table shows time, type, model, cost, source and no key column, so per-key spend is invisible even though every harness gets a dedicated `polli-harness-<id>` key.
- The endpoints already accept more (`enter.pollinations.ai/src/routes/account.ts`, `usageQuerySchema` / `usageDailyQuerySchema`): `api_key_ids` and `models` (comma lists), `days`, `granularity` + `period`, `format=csv`, and `before` / `before_event_id` cursors. Rows carry `api_key`, `api_key_id`, token columns and `cost_usd`. No backend change is needed.

### Outcome

- `polli usage --history` and `--daily` accept `--key <name-or-id>` (repeatable) and `--model <id>` (repeatable). A key name resolves through `/account/keys`; an id is passed through.
- `--days <n>` is exposed for both views instead of the server default.
- History rows show the key name, and tokens in and out, so a filtered run reads as a cost report.
- `--csv` passes `format=csv` through and prints the raw export.
- `--json` keeps working with the same filters.
- Unknown key names fail with a message listing near matches, not a silent empty table.
- README and the CLI docs get one example: cost of one harness key for the last day.
- `npm run test` and `npx biome check` pass in `packages/polli-cli`.

Note: on 2026-09-16 `/account/usage/daily?days=1&api_key_ids=…` returned `{"error":"Failed to fetch usage data"}` while `/account/usage?api_key_ids=…` worked. If that reproduces for you, report it in the PR; fixing the endpoint is not part of this quest.

### Resources

- [`polli usage` source](https://github.com/pollinations/pollinations/blob/main/packages/polli-cli/src/commands/usage.ts) · [usage route schemas](https://github.com/pollinations/pollinations/blob/main/enter.pollinations.ai/src/routes/account.ts) · [`polli keys list --json`](https://github.com/pollinations/pollinations/blob/main/packages/polli-cli/src/commands/keys.ts) for key ids
- Related: #14879 (polli update quest), #15014 (MCP install quest)

### How this quest works

[...]

PR #15020, opened 2026-09-16
feat: filter polli usage by key, model and day range

davealan74:
Fixes #15016

- `polli usage --history`/`--daily` accept repeatable `--for-key <name-or-id>` (resolved through `/account/keys`, ids pass through) and `--model <id>` filters
- `--days <n>` exposes the server's day window instead of the fixed default
- `--csv` passes `format=csv` through and prints the raw export
- History rows show the key name plus summed input/output tokens
- Unknown key names fail with near-match suggestions instead of a silent empty table
- README and SKILL.md each get one example: cost of one harness key, last day

No backend changes — `/account/usage` and `/account/usage/daily` already accept these query params. Daily + `--model` filters client-side since `/account/usage/daily` doesn't take a `models` param server-side.

`--for-key` (not `--key`, as suggested in the issue) because `--key` is already the global auth-key-override flag on every `polli` command — reusing it on `usage` silently swallowed the value into the wrong option.

**Testing**: `npx vitest run` (11 new unit tests for day/key/token parsing), `npx biome check`. Verified request wiring end-to-end (arg parsing → query string → response rendering, CSV passthrough, unknown-key error path) against a local mock server matching the real `/account/usage`, `/account/usage/daily`, `/account/keys` schemas — this sandbox's stored Pollinations credentials returned 401 against the live API, so a live run wasn't possible here.

---
Note: this fork's `main` is stale relative to `pollinations/main`, and the push token here lacks the `workflow` OAuth scope needed to sync it (can't push the intervening commits that touch `.github/workflows/*`). As a result GitHub may show a conflict on `README.md` — it's a single-line insertion next to an unrelated doc addition upstream, trivial to resolve in the merge editor. `SKILL.md`, `usage.ts`, and `api.ts` merge clean.


voodoohop:
Thanks for the submission. We merged #15045 instead — this PR had unrelated ~600-line doc churn and merge conflicts with main that made it hard to evaluate as a focused change. Closing in favor of #15045.

PR #14938, opened 2026-09-15
feat: add Tinybird-native page view and checkout tracking

voodoohop:
Chosen over #14924 (PostHog), which is closed: no new vendor or client SDK, consistent with #14234.

## Summary

- Adds two events: signed-in `page_viewed` and `checkout_started`. Signups and fulfilled payments are not duplicated; join `d1_user` and `stripe_event` for the funnel.
- Reuses Tinybird ingestion with zero new dependencies. A bodyless, same-origin page-view POST derives identity from the authenticated session; browsers cannot submit events or arbitrary properties. DNT honored on both sides.
- Adds `product_event` and a private `product_analytics_daily` endpoint with distinct event/user counts. These are activity counts, not signup-cohort conversion rates.
- The page enum in `shared/product-analytics.ts` is hand-maintained; a new route drops its views until added.

## Activation

Disabled by default. Deploy the datasource and pipe to staging, verify the ingest token APPEND scope, then enable `TINYBIRD_ANALYTICS_ENABLED` and `VITE_TINYBIRD_ANALYTICS_ENABLED`. Production requires separate approval. No token changes or cloud deployment performed.

## Verification

- 81 tests across `product-analytics.test.ts` and `stripe.test.ts`, typecheck and Biome pass.

Partially addresses #14234.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #15024, opened 2026-09-16
feat: auth-flow drop-off tracking and signup funnel percentages

voodoohop:
Re-applies #14938 (reverted in #15025 before it reached production) together with its follow-ups, so the conversion tracking ships as one change. Addresses #14234.

**No browser identifier.** No cookie and no random id. The browser keeps only the referrer hostname and the campaign tags of the landing URL in `sessionStorage`, so they survive the GitHub redirect and a signed-in view still carries the source it came from. Attribution then joins on the server-authenticated `user_id`.

**Writes (all into Tinybird `product_event`, behind `TINYBIRD_ANALYTICS_ENABLED` / `VITE_TINYBIRD_ANALYTICS_ENABLED`)**
- Page views, signed in and signed out, via one same-origin beacon. Each carries page, referrer hostname, `utm_*` / `ref`, and the app key from the page URL. Values are length-capped, not filtered.
- Sign-in drop-off as two server-side counts: `sign_in_started` when GitHub is requested, `sign_in_completed` when the session exists. No browser state, so the `/app/sign-in` flow is covered too. Both honour DNT and Global Privacy Control.
- Device code issued / approved / denied / token, PKCE consent grants, Discord link start/complete, checkout started, auto top-up enabled/disabled with amount. No codes, tokens, emails or IPs. Rows expire after 24 months.

**Reads (pipes)**
- `traffic_sources`: source × campaign → views, users, signups, paid, by week.
- `kpi_signup_funnel`: registration week → Pollen/top-up view → checkout → paid, pct per stage.
- `auth_flow_funnel`: daily sign-in and device stages, pct per stage.
- `checkout_outcomes`: fulfilled vs expired vs failed Stripe sessions by pack, from existing `stripe_event`.
- `product_analytics_daily`: as in #14938, anonymous rows not counted as users.

**Tests**: analytics enabled in the test worker, `product_event` recorded by the Tinybird mock; beacon guards, attribution passthrough, sign-in start/complete through a real OAuth round trip, device approve/deny/token, PKCE consent, opt-out.

**Privacy policy**: §7 describes the measurement, states that no cookie and no identifier are set, names what session storage holds, DNT/GPC, and the 24-month retention; date bumped.

Deploy: `tb deployment create --check` against staging is valid, no destructive operations. Deploy to staging and prod Tinybird before flipping the Worker flags.


voodoohop:
Adversarial review pass on the first commit, resolved in 726c24b459:

- `sign_in_completed` fired on admin impersonation (session.create.after is not sign-in-only) → gated on the `/callback` path.
- `pct_sign_in_completed` mixed units: cookie-less starts collapsed to one `flow_id` → stages counted per event id.
- Funnel join nodes scanned all of `product_event` / `stripe_event` → bounded to the cohort window.
- `device_token_issued` queued before the delete committed → moved after.
- `/deny` gained `requireUser()` → reverted; user id is optional there.
- `Sec-GPC` not honoured; no TTL on behavioural rows → both added.
- `client_id` from the unauthenticated device `/code` was `LowCardinality` → plain `String`, length-capped.

Accepted as-is: the beacon is anonymous by design (same posture as `/api/referral`); an attacker can inflate view counts at Tinybird row cost. Open: the pre-consent `auth_flow` cookie, see the description.

ElliotEtag:
- Reviewed `a7c4f48`. Resolve these before enabling collection or relying on conversion metrics:
- **Privacy:** [Server capture](https://github.com/pollinations/pollinations/blob/a7c4f48/enter.pollinations.ai/src/utils/product-analytics.ts#L46) ignores DNT/GPC, while the policy promises nothing is recorded. Align collection with the stated opt-out.
- **Funnel accuracy:** [Independent daily counts](https://github.com/pollinations/pollinations/blob/a7c4f48/enter.pollinations.ai/observability/endpoints/auth_flow_funnel.pipe#L17) split flows across midnight and can exceed 100% on repeated sign-ins. Correlate attempts and apply the requested 30-minute window.
- **Attribution:** [Payment joins](https://github.com/pollinations/pollinations/blob/a7c4f48/enter.pollinations.ai/observability/endpoints/traffic_sources.pipe#L49) credit historical purchases to new visits and miss purchasers without a new login. Define a conversion window and include already-authenticated visitors.
- **Delayed payments:** Add `checkout.session.async_payment_succeeded` to both the [success aggregate and event filter](https://github.com/pollinations/pollinations/blob/a7c4f48/enter.pollinations.ai/observability/endpoints/checkout_outcomes.pipe#L14).
- **Flow correlation:** The [shared cookie](https://github.com/pollinations/pollinations/blob/a7c4f48/enter.pollinations.ai/frontend/src/components/analytics.tsx#L67) can be overwritten by another tab and expires after ten minutes. Bind correlation to the auth transaction.
- **Campaign fields:** [Oversized stored tags](https://github.com/pollinations/pollinations/blob/a7c4f48/enter.pollinations.ai/frontend/src/components/analytics.tsx#L58) discard subsequent views and prevent cookie creation. Normalize optional attribution before storing it.
- Add regression coverage for these cases. Leaving the flags disabled is consistent with the staged rollout.

voodoohop:
@ElliotEtag all six addressed in 66a79a5, plus the cookie is gone entirely.

- **Flow correlation.** Dropped the `auth_flow` cookie and both Better Auth sign-in hooks. A tab keeps its id in `sessionStorage`, which survives the GitHub redirect, so a sign-in is a tab whose views go from an empty `user_id` to a real one. Two tabs can no longer overwrite each other, nothing expires mid-redirect, and the events `sign_in_viewed/started/completed` are deleted.
- **Privacy.** `captureFromRequest` gates every server-side event on DNT and Sec-GPC and replaces the repeated `waitUntil` at each call site, so the helper is shorter than what it replaced. The policy paragraph now says no cookie is set.
- **Funnel accuracy.** `auth_flow_funnel` aggregates per flow and dates each flow by its start, so midnight no longer splits one flow and a stage cannot exceed the one above it. Correlation is bounded by `window_minutes`, default 30, measured from the last signed-out view. Consent and link counts moved out to `product_analytics_daily`, where they already were.
- **Attribution.** `traffic_sources` credits a purchase only within `conversion_days` of the visit, default 7, and counts visitors who arrived already signed in, not only new logins.
- **Delayed payments.** `checkout.session.async_payment_succeeded` added to both the aggregate and the filter in `checkout_outcomes`.
- **Campaign fields.** Values are capped where they are captured, so one oversized tag no longer makes every later view in that tab fail validation.

Coverage: the cookie test is replaced by one asserting a tab keeps its id across sign-in, plus one asserting an opted-out browser produces no server event. 73 tests pass, typecheck and Biome clean, `tb deployment create --check` valid against staging with no destructive operations.

PR #15015, opened 2026-09-16
fix: harden Stripe checkout and keep fraud actions manual

voodoohop:
- Requests best-effort 3DS for first purchases and packs under $10. Unsupported cards can still complete payment; this does not require authentication, guarantee liability shift, or eliminate dispute fees. No new Radar block rule is added.
- Counts distinct cards from both successful and failed charges toward the four-card/24-hour gate. One successful card plus three other distinct cards reaches the limit; repeated attempts on one card count once. Existing `app_new_card_*` metadata keys remain unchanged. Enforcement depends on the existing Radar gate rule.
- Keeps bans and refunds manual, explicitly disables scheduled automatic bans, and updates the operator documentation. Removes log-only dispute/warning handlers and their redundant tests; no new event subscriptions are needed for this PR.
- Daily reporting and scan optimizations remain in #15050; refund accounting remains in #15044.
- Validation: 90 Stripe/fraud regression tests, TypeScript, and Biome passed. No merge, deployment, or Stripe configuration changes performed.

Issue #10583, opened 2026-04-30
Debit pack_balance on Stripe refund


ElliotEtag:
Refund webhooks (`refund.created`, `refund.updated`, `refund.failed`) at [stripe-webhooks.ts:403-427](https://github.com/pollinations/pollinations/blob/main/enter.pollinations.ai/src/routes/stripe-webhooks.ts#L403-L427) currently only log to Tinybird. Refunded pollen is never reversed — user keeps their balance even after Stripe returns the money.

Now relevant because the Stripe Portal auto-top-up PR (#TBD) opens up self-service charges, increasing the chance of refund requests.

## Scope
- Look up the original pack/invoice from `refund.payment_intent` or `refund.charge` → metadata (`packAmount` field)
- Compute pollen to reverse (full refund → all `pollenGrant`; partial → proportional)
- Debit `user.pack_balance` in D1
- Add a `stripe_refund` table for idempotency (mirrors `stripe_auto_top_up_attempt` pattern), keyed on `stripe_refund_id` UNIQUE
- Handle two source types: pack purchases (Checkout session metadata) and auto-top-up invoices (invoice metadata)

## Open design questions
- **Negative balance**: clamp `pack_balance` at 0 (user kept what they spent), or allow negative (charges accrue against next top-up)?
- **Tax**: refunded amount includes tax; pollen grant didn't. Use `pollenGrant` × (`refund.amount` / original `amount_subtotal`)?
- **Auto-top-up interaction**: should a refund auto-disable the user's auto-top-up to prevent immediate re-charge? With reason "Auto top-up disabled after refund."?
- **Operator override**: ops sometimes refunds for goodwill without intending to debit. Skip if refund metadata has `goodwill: true`?

## Acceptance
- Integration test against mocked Stripe refund event → verifies `pack_balance` debited
- Idempotency test: same refund event delivered twice → only one debit
- Partial refund test: $10 refund on $50 pack → debits 20% of pollen grant

omribz156:
I can take this as a paid, fixed-scope billing integrity PR if useful.

Codex would help with the implementation pass, but I would keep it bounded: refund idempotency, refund -> original pack/auto-top-up source lookup, full/partial `pack_balance` debit, and mocked Stripe refund + duplicate-delivery tests. I would first lock the design choice on negative balances / goodwill refunds before coding.

$380 after accepted PR/review. If you prefer, I can do a $90 diagnosis/design pass first to pin the refund math and idempotency shape.

PR #15044, opened 2026-09-16
fix: reverse Paid Pollen when Stripe refunds succeed

ElliotEtag:
- Reverses purchased Paid Pollen when a manual Stripe refund succeeds. Spent balances can become negative; Quest Pollen is untouched. Supports Checkout and auto-top-up invoices.
- Adds a transactional refund ledger for retry handling and restoration when refunds fail or are canceled. Proportional accounting remains for partial refunds already issued in Stripe; it does not grant refund eligibility.
- Clarifies the operator policy in the terms and refund policy: discretionary refunds cover whole, completely unused packs only. No refund of a pack once used, including its remaining balance. Preserves statutory rights, proven duplicate/unauthorized payment corrections, and rights attached to earlier purchases; aligns the price-change clause.
- Validation: existing 89 Stripe integration tests, fraud tests, TypeScript, and Biome passed for the implementation. The subsequent policy-only edit was reviewed for consistency and passed `git diff --check`; Biome does not process Markdown.
- Rollout: apply migration `0067_stripe_refund`, verify `refund.created`, `refund.updated`, and `refund.failed` subscriptions, and reconcile historical manual deductions before replaying refunds. No deployment performed.

Addresses #10583. Checkout protections: #15015. Daily reporting: #15050.


ElliotEtag:
!polli review please

voodoohop:
## Review: too complex for the problem. Suggest closing and replacing with a small version

**Why**
- Refunds are rare: 16 in 6 months, all issued by hand, 0 failed (Stripe events in Tinybird).
- The new policy only refunds unused packs. Partial and failed refund handling covers cases that almost never happen.
- 167 lines of logic, a new table and 359 lines of tests for about 3 events a month.
- Throwing when no purchase is found makes Stripe retry for 3 days on something that can never succeed.
- Replaying old refunds would take Pollen twice from users we already corrected by hand.
- This does not reduce dispute fees. That work is #15015.

**Minimal design**
1. Add one column: `stripe_checkout_credits.pollen_reversed` (default 0).
2. Handle `charge.refunded`. The event already contains the charge, `amount_refunded` and `payment_intent`, so there is no need to fetch the refund or the charge.
3. Find the purchase with `checkout.sessions.list({ payment_intent })`. If none is found, log it and return 200.
4. `target = pollen_credited × amount_refunded / amount_captured`
5. If `target > pollen_reversed`, in one batch: subtract the difference from `pack_balance` and set `pollen_reversed = target`, only if it still has the old value. Duplicate events then do nothing.

**Left out until needed:** failed refunds (fix by hand), auto top-up refunds, refunds issued before deploy.

**Separately:** the legal text changes should be their own PR. #15050 would read `pollen_reversed` instead of the refund table.

voodoohop:
Closing in favour of #15085: a full refund is recorded as a negative `stripe_checkout_credits` row, with no new table or migration. #15050 no longer depends on this PR.

PR #15066, opened 2026-09-17
feat(polli-cli): filter polli usage by key, model and day range

xiaotian1171:
`polli usage` can now answer "what did this key cost me over the last n days", from the CLI, for both views.

## What changed

`packages/polli-cli`: `usage` gained `--key <name-or-id>`, `--model <id>`, `--days <n>` and `--csv`:

```bash
polli usage --history --key polli-harness-dsh --days 1
polli usage --history --key <name-or-id> --model <id> --days 7 --csv > usage.csv
polli usage --daily --key polli-harness-dsh --model openai/gpt-5.4-nano --days 7
```

- `--key` accepts a key **name** or **id**, is repeatable and comma-separated. Names are resolved through `GET /account/keys`; ids pass straight through. An unknown name errors and lists the closest names (`Unknown key "X". Near matches: ...`) instead of silently returning an empty table.
- `--model` is repeatable/comma-separated; `--days` is validated as a positive integer; `--csv` prints the raw export (`format=csv`) untransformed.
- History rows now show `key` next to `in`/`out` tokens and cost; daily rows show `key` too. `--json` keeps working with the same filters.

## `--key` placement (commander pitfall)

The program already has a global `--key <key>` that overrides the stored API key, and commander routes **any** `--key` written after the subcommand name into that global option, which made a local `usage --key` both unusable and dangerous (the filter value replaced the credential → 401). `splitKeyArgs()` reads `argv` instead:

- `polli --key sk_... usage` → still the global override (unchanged, as documented in SKILL.md)
- `polli usage --history --key polli-harness-dsh` → key filter
- `polli usage --history --key sk_...` → still an override, as documented in the README

## Verification (real account, key with `profile usage`)

```
$ polli usage --history --key code --limit 2
time                 type           key   model                         in     out  cost     source
2026-09-17 04:10:18  generate.text  code  deepseek/deepseek-v4.1-flash  95223  342  $0.0010  tier

$ polli usage --history --key code --model openai/gpt-5.4-nano --days 7
# only that model's rows, last 7 days

$ polli usage --daily --key code --days 7
date        key   model                     requests  cost     source
2026-09-17  code  deepseek/deepseek-v4.1-flash  79    $0.2189  tier

$ polli usage --history --key nope-key --limit 1
error: Unknown key "nope-key".                        # exit 1

$ polli usage --history --key code --days 1 --csv
timestamp,type,model,api_key,api_key_type,meter_source,...   # raw CSV

$ polli usage --history --days 0
error: --days must be a positive integer               # exit 1
```

`npm run test` → 15 files, 148 passed. `npx @biomejs/biome check src/` → clean.

## Bug found on `/account/usage/daily` (not fixed here, by design)

Passing `api_key_ids` to the daily endpoint returns 500 `{"error":"Failed to fetch usage data"}` on the live API, with or without `days`:

```
$ curl -H "Authorization: Bearer <key>" "https://gen.pollinations.ai/account/usage/daily?days=7&api_key_ids=<id>"
{"error":"Failed to fetch usage data"}
$ curl -H "Authorization: Bearer <key>" "https://gen.pollinations.ai/account/usage/daily?days=7"
{"usage":[...]}   # 200
```

So the daily view cannot be filtered server-side yet. Rather than let `polli usage --daily --key ...` fail with a 500, the daily view is filtered locally on `api_key_id` / `model` (same approach the `models` param already required there); the requests stay unchanged and can move back to the query string once the endpoint is fixed. `--daily --csv` still passes `api_key_ids` through, since a raw export cannot be filtered locally.

Fixes #15016


voodoohop:
Thanks for this — we merged #15045 as the base (smaller diff), but folded in your 'filters require --history/--daily' guard and the `send()`/`genText` dedup refactor from this PR before merging. You've been granted a 3 Pollen bonus for that — claim it at https://enter.pollinations.ai/quests. Closing in favor of #15045.

PR #15068, opened 2026-09-17
feat(polli): add filters to usage reports

shegerdha:
## Summary\n- add repeatable --key and --model filters to history and daily usage\n- resolve key names to API key IDs and report near matches for unknown names\n- add --days and --csv support\n- include the key in usage tables and document examples\n\nCloses #15016\n\n## Validation\n- 
pm run build\n- git diff --check\n- 
pm test (existing Windows-only permission/tooling failures in harness and update-notice tests)


voodoohop:
Thanks for the submission. We merged #15045 instead — this PR adds a key column to history rows but doesn't include token counts, which the quest asked for. Closing in favor of #15045.

PR #15081, opened 2026-09-17
Add --key/--model/--days/--csv filters to polli usage

Marcus-Mok-GH:
Fixes #15016

- Adds `--key`, `--model`, `--days`, and `--csv` filters to `polli usage --history` and `polli usage --daily`.
- Key names resolve via `/account/keys`; 32-char alphanumeric IDs pass through without a lookup. Unknown names error with near-match suggestions; duplicate names error listing the matching ids.
- History rows now include `key`, `tokens_in`, and `tokens_out`; daily rows include a `key` column.
- `--csv` sets `format=csv` and prints the response verbatim; `--daily --model` filters client-side and is rejected together with `--csv` since the daily endpoint has no server-side model breakdown.
- Validates `--days` as a positive integer up to 90.
- Keeps root `--key <auth>` separate from the usage filter `--key` via `enablePositionalOptions()`; root/trailing `--json` on keys list still works.

Tests: 156 passing (was 132), covering name/ID/ambiguous/unknown/repeated keys, models/days query params, nullable token mapping against the endpoint schema, JSON and exact raw CSV output, daily model-CSV rejection, and auth-vs-filter routing. Build + biome clean.

No backend changes.


voodoohop:
Thanks for this — we merged #15045 as the base, but folded in your 32-char key-id pass-through (skipping the `/account/keys` lookup) before merging — it's the only design that works for `account:usage`-only scoped keys. You've been granted a 2 Pollen bonus for that — claim it at https://enter.pollinations.ai/quests. Closing in favor of #15045.

PR #15082, opened 2026-09-17
fix(dashboard): combobox filter suggestions when editing tokens

### LINKED ISSUE #15078: MINOR UI BUG AT https://enter.pollinations.ai/models
at here, you can see i am unable to turn this to community model, i have to type 'c' in that textbox to see community option

@gggff123 You can self go and try it...

Image

fisventurous:
fixes combobox filter suggestions when clicking a filter token so options like community are discoverable without having to type
\- pass the active draft token directly to `getModelQuerySuggestions` instead of reading the tail of the query
\- keep source suggestions enabled when editing the source filter token

closes #15078


fisventurous:
@voodoohop

PR #15083, opened 2026-09-17
fix(dashboard): resolve video and 3d health lookup and show full catalog by default

fisventurous:
fixes video and 3D model health lookups against gateway image rollups and removes `status:healthy` from default catalog query injection

firefox_t2bQa6R3Wl

in `gen.pollinations.ai`, video and 3D models route through the image pipeline (`generate.image`). `modelHealthLookup` was checking `generate.video` and `generate.3d`, which do not exist in Tinybird rollups. this marked all 18 official video models as `unknown`, causing the entire video tab to vanish when combined with the default `status:healthy` filter.

defaulting discovery to `status:healthy` also hides new or low-traffic models whose 24-hour sample volume is below threshold (`unknown`), as well as models undergoing brief upstream degradation. the table already renders status dots (green, amber, grey) next to every model name for real-time health visibility, so keeping the full catalog discoverable by default prevents catalog churn while still allowing users to opt into `status:healthy` manually.

\- map `video` and `3d` categories to `generate.image` rollups in `modelHealthLookup` with fallback
\- remove `status:healthy` from `ensureModelQueryDefaults` so the catalog defaults to all official models
\- update query and catalog tests to verify video/3D health resolution and default query filters


voodoohop:
- Superseded by #15270/#15279: health now comes from the gen catalog, and the default \`status:reliable\` only filters community models, so official video/3D models stay visible
- Thanks!

PR #15085, opened 2026-09-17
fix: reverse purchased Pollen when a Stripe charge is refunded

voodoohop:
- Handles `charge.refunded`: a full refund inserts a negative row into `stripe_checkout_credits` (`session_id = refund:<session>`, `pollen_credited = -N`) and subtracts N from `pack_balance` in the same batch. Same pattern as crediting: a repeated event hits the primary key and does nothing.
- No migration and no new column. The sum of `pollen_credited` per user is now net of refunds.
- The 3DS "returning buyer" check uses that net sum, so a fully refunded history (e.g. fraud) still gets 3DS. The top-up quest already sums, so a refunded top-up no longer counts toward it.
- Partial refunds and purchases with no matching credit are logged and acknowledged with 200. Partial and failed refunds are fixed by hand. Auto top-up refunds are out of scope.
- Replaces #15044 per [this comment](https://github.com/pollinations/pollinations/pull/15044#issuecomment-5714027450). 79 Stripe integration tests pass, backend typecheck and Biome 2.3.10 clean.

**Rollout**
- Subscribe the production Stripe webhook to `charge.refunded`. It currently receives only `refund.*`, so this code does not run until then.
- Anything counting purchases by row count should filter `pollen_credited > 0`.

Addresses #10583


voodoohop:
@ElliotEtag this replaces #15044. Could you review it?

**Stripe webhook setup (after this PR is deployed to production):**
- Stripe Dashboard (live mode) → Developers → Webhooks → endpoint `https://enter.pollinations.ai/api/webhooks/stripe`
- Add the event `charge.refunded`. The endpoint currently only listens to `refund.created`, `refund.failed` and `refund.updated`; leave those as they are.
- Do not create a new endpoint. The signing secret stays the same.
- Order: deploy first, then subscribe. Before the deploy, the endpoint would acknowledge `charge.refunded` without reversing any Pollen.

**Verify:**
- On the next full refund, the event delivery shows `200`.
- `stripe_checkout_credits` has a row `refund:<session_id>` with negative `pollen_credited`, and the user's `pack_balance` dropped by that amount.
- Partial refunds only log a warning and need a manual Pollen adjustment.

PR #15124, opened 2026-09-18
Add key, model, and day filters to polli usage

alkonsafe:
Fixes #15016

- `polli usage --history` / `--daily` accept `--key <name-or-id>` (repeatable); names resolve through `/account/keys`, ids pass through
- `--model <id>` (repeatable): server-side filter for history; exact client-side filter for daily (that endpoint doesn't support `models`)
- `--days <n>` for both views, reusing `parseDaysWindow` (max 90, matching the server)
- `--csv` passes `format=csv` through and prints the raw export (`api.ts` gains a `raw` response option)
- History rows now show key name and tokens in/out; daily rows show key name
- Unknown key names fail listing near matches, not a silent empty table
- Filters on the default balance view fail with a clear message
- Tests for the pure helpers (`sumTokens`, key matching); README + SKILL.md example: cost of one harness key for the last day

`npm run test` and `npx biome check` pass.


voodoohop:
Thanks for this — we merged #15045 as the base. We reused your idea of importing `parseDaysWindow` from `earnings.ts` instead of duplicating the day-window logic, and credited it in the merge. This PR itself had a blocking issue (`--key` handling could override auth), so no bonus this time, but appreciated the pointer. Closing in favor of #15045.

Issue #15014, opened 2026-09-16
🔌 [QUEST] Install Pollinations MCP servers into popular coding agents from the Polli CLI


voodoohop:
### How this quest works

Multiple contributors may submit PRs. We select the best completed approach, preferring correctness, clarity, and minimal code. The author of the selected merged PR receives the reward; no issue assignment is required.

In your PR, briefly share which clients you tested on and how you verified each one end to end.

### Goal

Let the Polli CLI connect any Pollinations-hosted MCP server to the coding agents and IDE assistants people already use, the way `npx @smithery/cli install` and `npx ctx7 setup` do for their servers.

### What exists today

- Pollinations hosts five MCP servers, all remote, stateless, Streamable HTTP with `Authorization: Bearer <key>`: `pollinations`, `ffmpeg`, `exa`, `composio`, `computer`. The live catalog is `GET https://gen.pollinations.ai/mcp`. There is no stdio server anymore.
- `polli harness <id> on` already creates a dedicated key per tool and, for DeepSeek Harness only, writes the `pollinations` MCP entry (`packages/polli-cli/src/harnesses/dsh.ts`). `--no-mcp` is accepted by every harness but is a no-op on the other five.
- The docs describe one client only: `claude mcp add --transport http pollinations https://gen.pollinations.ai/mcp/pollinations --header "Authorization: Bearer YOUR_KEY"`.

### Outcome

- A user can install one, several, or all catalog servers into a chosen client with one Polli command, and remove them again. Whether this lives under `polli harness`, a new `polli mcp` command, or both is up to you; explain the choice in the PR.
- The server list comes from the live catalog, not a hard-coded list, so new servers appear without a CLI release.
- Every entry uses the hosted HTTPS URL with a bearer key. No `npx` shim and no local process, except where a client cannot express headers (for example Claude Desktop, where `mcp-remote` is acceptable).
- The key is a dedicated Polli-created key per client, as the harness adapters do today. Keys may be stored directly in local client configs for simple reuse on reinstall. Codex may use its environment-variable reference with a local .env file. Use owner-only file permissions where supported and document that keys are plaintext and must not be shared or committed.
- Install is idempotent, `off`/uninstall removes only Pollinations-owned entries, unrelated config is untouched, and `status` reports what is installed per client.
[...]

PR #15032, opened 2026-09-16
feat(polli-cli): polli mcp installs MCP servers into 13 coding agents

afanasevmylife:
- Adds live MCP catalog discovery and install/status/remove commands for 13 clients.
- Reuses dedicated per-client keys and preserves foreign MCP entries. Keys are stored locally in plaintext; Codex references its .env key.
- Fixes repeat installs, Amp/Gemini settings, JSONC, native Zed HTTP, OpenCode config paths, and Codex ownership detection.
- 179 local CLI tests and build pass, including real Claude CLI repeat-install/removal. Not every client was exercised end to end.
- Credits #15067 (xiaotian1171) for repeat-install/JSON5 ideas and #15128 (aikhusus2025-ctrl) for Amp/Zed configuration.

Closes #15014

PR #15128, opened 2026-09-18
Fixes #15014 — polli mcp: install Pollinations MCP servers into coding agents

aikhusus2025-ctrl:
## What

A new `polli mcp` command installs the Pollinations MCP servers from the live
catalog into the coding agents and IDEs people already use, and removes them
again:

```bash
polli mcp list                            # live catalog + nothing hard-coded
polli mcp add claude-code                 # every catalog server, or:
polli mcp add cursor --server pollinations --server ffmpeg
polli mcp add all                         # every client detected on this machine
polli mcp add claude-code --verify        # write, then handshake what was written
polli mcp status [client]                 # what is installed, and in which file
polli mcp status --verify                 # handshake every installed entry
polli mcp remove cursor
polli mcp remove all
```

## Why a new command instead of `polli harness`

`polli harness <id> on` configures one chat-model provider per harness. MCP
entries are different: they are per-server, they land in files other tools also
own (`.claude.json`, the OpenCode config, VS Code's `mcp.json`), and they must
stay correct across repeated installs. `polli mcp` owns exactly the entries it
wrote and leaves every neighbouring key alone. Harness ids are still accepted
(`polli mcp add dsh --no-browser` delegates to `polli harness dsh on`, the only
harness that registers MCP servers today); the other harnesses answer with a
clear "does not register MCP servers — run `polli harness <id> on`" message
instead of silently reconfiguring something else.

## Clients (the issue's priority order)

| Client | How the entry is written |
| --- | --- |
| Claude Code | `claude mcp add --scope user --transport http` (config fallback) |
| Codex CLI | `codex mcp add --url … --bearer-token-env-var POLLI_MCP_CODEX_API_KEY`; key in a polli-owned `0600` env file |
| VS Code / Copilot Chat | user `mcp.json`: `servers` + an `inputs` prompt — no literal key in the file |
| Cursor | `~/.cursor/mcp.json`, `url` + bearer header |
| OpenCode | its own config file through the existing `opencodeConfigFile()`, `type: remote` |
| Gemini CLI | `gemini mcp add --scope user --transport http` (config fallback) |
| GitHub Copilot CLI | `$COPILOT_HOME/mcp-config.json`, `type: http` |
| Windsurf | `~/.codeium/windsurf/mcp_config.json`, `serverUrl` |
| Cline | the extension/CLI config, `type: streamableHttp` |
| Amp | `amp.mcpServers` in `~/.config/amp/settings.json` |
| Kiro | `~/.kiro/settings/mcp.json` |
| Zed | native remote `context_servers` (`url` + `headers`) — no `mcp-remote` bridge |
| Warp | `~/.warp/.mcp.json` |
| Claude Desktop | `mcp-remote` bridge (the one client that cannot send headers) |

Three choices where "prefer the client's own `mcp add`" and the other
requirements pull in different directions — all three are explained so a
reviewer can disagree knowingly:

- **VS Code** gets a config file rather than `code --add-mcp`, because
  `--add-mcp` cannot reference an `inputs` prompt and the key should not land in
  the config file.
- **Amp** gets a config file rather than `amp mcp add`, because Amp's documented
  `mcp add` registers a bare URL and the documented place for a bearer header is
  the `amp.mcpServers` entry (`url`/`headers`).
- **Zed** uses its native remote support, so no `npx` shim is needed.

## Keys, ownership, safety

- One key per client (`polli-harness-mcp-<client>`) through the existing
  `resolveHarnessKey` flow, so a single client can be revoked in API key settings;
  re-installing reuses the key already on disk.
- An entry is Pollinations-owned when one of its URLs (or, for Claude Desktop,
  its `mcp-remote` args) points at `gen.pollinations.ai/mcp/`. Nothing else is
  touched, and no marker field is written into configs we do not own.
- Install is idempotent; `remove` strips only owned entries plus the `inputs`
  entry polli added, and deletes the cached key only once no owned entry is left
  for that client (the key is shared by every server installed there).
- Config files polli writes, and the Codex env file, are written `0600`; the mode
  is applied explicitly so a pre-existing world-readable config is tightened
  rather than left as it was.
- Each target runs in isolation: one broken client prints an error, sets the exit
  code and does not stop the rest; a catalog failure stops before any write.

## Verifying the entry, not just the file

`polli mcp status --verify` (and `add --verify`) sends the same `initialize`
request an MCP client sends when it connects — `protocolVersion 2025-06-18`,
`clientInfo.name: polli`, `Accept: application/json, text/event-stream` — with
that entry's key, and reports what came back:

```
client     server        url                                        check
claude-code pollinations https://gen.pollinations.ai/mcp/pollinations ok (pollinations-mcp)
cursor      ffmpeg       https://gen.pollinations.ai/mcp/ffmpeg      key rejected: HTTP 401
```

Outcomes: `ok (<serverName>)`, `key rejected`, `bad response`, `unreachable`.
The body is read as JSON or as an SSE `data:` frame, because the streamable-HTTP
transport may answer with either.

## Verification

New tests: `src/mcp/catalog.test.ts`, `src/mcp/clients.test.ts` and
`src/mcp/verify.test.ts` (50 tests, table-driven over the client list) covering
live-catalog parsing and selection, per-client config shapes, idempotent
re-install, ownership-scoped removal, Codex TOML + env file, VS Code `inputs`,
the Claude Desktop bridge, the `claude mcp add/remove` argv, and the handshake
itself (JSON and SSE bodies, bearer included/omitted, 401, dead host, non-JSON
body, JSON-RPC error). All 50 pass; a `tsc` program diff against the same tree
without this PR shows no new diagnostics (verified with a negative control that
injects a deliberate type error), and `biome check` is clean on the changed
files.

Clients tested end to end in a temporary `$HOME` (real file writes; CLIs stubbed
with their argv asserted):

- **Claude Code** — `mcp add --scope user --transport http <id> <url> --header …`, then `mcp remove --scope user <id>`.
- **Codex CLI** — `config.toml` table with `bearer_token_env_var`, key written to the polli-owned env file at `0600`; status reads the TOML back.
- **VS Code** — `inputs` + `servers.<id>.headers.Authorization = Bearer ${input:polli-mcp-key}`; no literal key anywhere in the file.
- **Cursor, OpenCode, Copilot CLI, Windsurf, Cline, Amp, Kiro, Zed, Warp** — exact config shape, second install is a no-op, remove returns the file to its original content (a neighbouring user entry survives).
- **Claude Desktop** — `npx -y mcp-remote <url> --header "Authorization: Bearer …"`.
- **Gemini CLI** — `mcp add --scope user --transport http` argv, plus the config fallback.
- **Live catalog** — the shipped `fetchMcpCatalog()` was run against the real
  `GET https://gen.pollinations.ai/mcp` response: 5 servers parsed, `serverUrl(id)`
  round-trips, and URLs outside our host are rejected by `ownedServerId`.
- **Live keyed handshake** — `verifyServer` against
  `https://gen.pollinations.ai/mcp/pollinations`: without a key `HTTP 401
  UNAUTHORIZED`; with a real key `HTTP 200`,
  `content-type: text/event-stream`, and an SSE `data:` frame carrying
  `result.serverInfo.name = "pollinations-mcp"` — the outcome
  `status --verify` prints as `ok (pollinations-mcp)`.

Not verified here: launching the real GUI clients (Cursor, Zed, Windsurf, …) —
those adapters assert the configuration each vendor documents.

Docs: `gen.pollinations.ai/src/docs/mcp.md` gains a "Polli CLI" section and
`packages/polli-cli/SKILL.md` a short recipe.

### Security follow-up

Strix flagged one LOW finding on the first revision (CWE-78: catalog server id interpolated into
the Windows `shell: true` command). It is fixed in `72b884523`: `fetchMcpCatalog` now applies the
same id allowlist already used by `ownedServerId` (single `SERVER_ID` constant,
`/^[A-Za-z0-9][A-Za-z0-9._-]*$/u`), dropping unsafe catalog entries, with a test covering
`evil&calc`, `semi;colon`, `../escape`, `-leading`.

Fixes #15014


aikhusus2025-ctrl:
@strix-security re-review request — the follow-up to the security finding on
`catalog.ts:39-40` is in `72b884523`: catalog server ids are now checked against the
same allowlist already used for owned URLs — `/^[A-Za-z0-9][A-Za-z0-9._-]*$/` — before
an entry is accepted, so an id coming from the catalog can no longer carry cmd.exe
metacharacters into the Windows `shell: true` invocation. Entries with an unsafe id are
dropped, exactly like entries with no id today.

- `SERVER_ID` is now a single constant used by both `fetchMcpCatalog` and
  `ownedServerId`, so the two cannot drift apart.
- New test: "drops catalog ids that a shell could interpret" (`evil&calc`, `semi;colon`,
  `../escape`, `-leading` rejected; a normal id still round-trips).
- Test suite is now 51 tests; `biome check` and `tsc` are clean on the changed files.

The finding was low severity and required a compromised/redirected catalog, but the
allowlist was already agreed on for URLs, so applying it to ids as well closes the
gap at the source.

voodoohop:
- We used your Amp settings format and native HTTP setup for Zed in the follow-up. You are credited as a co-author. We kept the implementation smaller without the extra verification layer.

voodoohop:
- Your Amp configuration and native Zed HTTP setup helped the merged solution. Thanks!
- We added a 3-Pollen bonus, claimable on https://enter.pollinations.ai/quests.
- You are already on the agent publishing allowlist—we would love to see what you build next.
- Agent guide: https://github.com/pollinations/pollinations/blob/main/BUILD_YOUR_OWN_AGENT.md

PR #15136, opened 2026-09-18
feat: filter polli usage by key, model and day range

tomdacatto:
## Summary

- `polli usage --history`/`--daily` accept repeatable `--key <name-or-id>` and `--model <id>`, plus `--days <n>` (max 90) and `--csv`
- Key names resolve through `/account/keys`; ids pass through. An unknown value fails with near-match suggestions instead of a silent empty table
- History rows now show the key name and input/output tokens
- No backend changes — both endpoints already accepted these params (except `models` on `/account/usage/daily`, which has no server-side param, so `--model --daily` filters client-side)

## `--key` collision with the global auth flag

`--key` is already the global `polli --key <secret> <cmd>` auth override, registered on the root program. Commander resolves a same-named option to that global one regardless of where it appears on the command line, so a local `--key` on `usage` never receives its value (verified empirically, including with `enablePositionalOptions()`, which breaks the `<subcommand> --json` pattern used throughout the rest of the CLI and isn't safe to enable). `usage.ts` splits `process.argv` by hand instead: `--key` before `usage` keeps its documented meaning (auth override, per README/SKILL.md); after `usage` it's this quest's repeatable filter. `setKeyOverride` now accepts `undefined` so the action can reset it when no pre-subcommand key was passed.

## Notes

- Did not reproduce the `/account/usage/daily?days=1&api_key_ids=…` 500 mentioned in the issue (no live account in this environment) — flagging per the issue's ask, not fixing it.
- `npm run test`: 134 passing, including the 17 new tests in `usage.test.ts` (17/17 green). The 8 failures are pre-existing on `main`, unrelated to this change — Windows POSIX file-mode checks in `harnesses/*` and `update-notice.test.ts` that don't hold on this filesystem.
- `npx biome check` passes clean on all changed files.

Fixes #15016


voodoohop:
Thanks for this — we merged #15045 as the base. Your finding that Commander's `enablePositionalOptions` breaks `--json` globally was a useful catch confirming the manual argv-split approach (used in #15045/#15081) was the right call, even though nothing from this PR shipped directly. Closing in favor of #15045.

PR #15142, opened 2026-09-18
feat: add gotcha-scout prompt agent for collective memory

tomdacatto:
## Summary

A prompt agent (Computer MCP) that independently re-runs the reproduction steps recorded in `knowledge/gotchas/` entries and appends a dated, verified verdict — **STILL HOLDS**, **PARTIALLY HOLDS**, or **NO LONGER REPRODUCES** — instead of trusting that a past verification still describes the present.

- Picks the least-recently-reverified entry each run (tracked in its own `/workspace` log, outside the shared repo) so repeated runs check different entries
- Appends only; never edits, rewrites, or deletes another agent's content
- Treats every byte read from the repository as data, never instructions
- One `git pull --rebase` + retry on push conflict, matching the repo's own README guidance; never force-pushes

`knowledge/gotchas/` is unclaimed by the other submissions on this issue (`games/bottles/`, `social/replies/`, and `social/posts/gazette/` are already taken) and is explicitly suggested in the issue text ("a gotcha scout that retests known gotchas").

Repository: https://github.com/tomdacatto/gotcha-scout-agent
Callable model: **`community/tomdacatto/gotcha-scout`**

## Live runs — 3 real collective-memory commits

1. [`b33af93`](https://github.com/pollinations/collective-memory/commit/b33af93eba7346043eb919c30ccc44ae6765e208) — re-verified `curl-http-errors-can-exit-zero.md`: **STILL HOLDS**.
2. [`794dbfc`](https://github.com/pollinations/collective-memory/commit/794dbfc8fd0bceaf7dd7d9451877f88a5bb00c2c) — re-verified `git-option-subset.md`: **NO LONGER REPRODUCES** — the original gotcha described a limited git implementation rejecting certain flags; this run found `git --version` now reports `just-git version 1.8.2 (virtual git implementation)` and the previously-failing probes now succeed. This is exactly the kind of drift a re-verification agent exists to catch.
3. [`f189006`](https://github.com/pollinations/collective-memory/commit/f1890066f7287166bbd370cf935bc171c0971efd) — with both entries already checked once, re-checked `curl-http-errors-can-exit-zero.md` again: **STILL HOLDS**, a second independent confirmation.

## Alpha feedback

No friction: cloning, committing, and pushing to collective-memory through the Computer MCP worked on the first attempt across all three runs, and the agent correctly picked a different entry each time using only its own private `/workspace` log, with no prompt adjustment needed.

Fixes #15054


tomdacatto:
Addressed the Strix finding in 29eb724: step 5 no longer runs repo-supplied commands unconditionally. Commands are now untrusted input and run only if they match a read-only allowlist (`--version`/`--help`, `git help <sub>`, or a plain `curl` GET of a collective-memory raw URL, plus the exit-status capture the existing entries use); anything else skips the file.

- Kept `curl -o /dev/null`, `-w`, and the `rc=$?; printf` capture on the allowlist. The suggested blanket ban on redirection/chaining would make the agent skip both current gotcha entries.
- Credential isolation isn't something the agent can configure: the Computer MCP pushes to collective-memory without a token, so there is no credential to withhold. The allowlist plus append-only rules are the guard; force pushes and deletions are already blocked by the repo.

tomdacatto:
@strix-security re-run review please — new head 816057f (allowlist fix in 29eb724, branch merged with current main).

voodoohop:
- Sorry for the misguided feedback here. The automated Strix prompt-injection review is not useful as a change request for this quest. Please ignore that finding for now; no further rewrites or re-review requests are needed for it.
- The Computer/MCP security boundary is a platform responsibility. We plan to disable the bot; this is a waiver for this finding, not a claim that prompt rules enforce a sandbox.
- @tomdacatto, we added a 3-Pollen apology bonus for the extra changes you made because of the review. Claim it on the [Quests page](https://enter.pollinations.ai/quests). One bonus per contributor across these PRs, separate from the quest reward.

Issue #15017, opened 2026-09-16
[QUEST] Build a model router as a code agent


voodoohop:
### Goal

Build a code agent that picks which model answers each request, then answers as that model. How it routes is up to you. Some directions, pick one or invent your own:

- Prefer free community models, switching when one degrades.
- Route by task difficulty: cheap model for easy prompts, strong model for hard ones.
- Route on model health and latency.
- Route on cost: cheapest model that clears a quality bar, or a budget per conversation.
- A clever fallback chain that escalates on errors, timeouts or rate limits.

Feel free to look at modern routing research and tools (classifier routers, semantic routers, cascades) and adapt what fits. Gen already retries a model's declared fallbacks, so your value is choosing between models, not retrying one.

> **Alpha:** code agents are in alpha. Reports of usability problems and suggestions for making agents easier to build are very welcome. Share them in this issue or your PR; useful feedback earns a small bonus.

### What you can use

- `GET https://gen.pollinations.ai/v1/models` returns every model with pricing and capabilities.
- `GET https://gen.pollinations.ai/models/status?minutes=30` returns request counts, 5xx errors, fallback rescues, p50/p95 latency and tokens per second. Rows with `is_rollup` 1 are per model; the others are per route (primary and fallbacks).
- Inside the agent, call these through `pollinations("/v1/models")` and `pollinations("/models/status?minutes=30")`. Only `ai` and `@ai-sdk/openai-compatible` can be imported.

### Outcome

- The router works as a normal model: callers see only the chosen model's answer.
- Show which model was chosen and why in a header, log or short trace, so we can verify the routing.
- Demonstrate at least three requests routed differently, with the reason for each.

### Submission

Deploy the agent under your own account and test it from your public repository. Submit a PR with a copy of the agent code and a short README under `apps/agent-<name>/`. The PR description must include your repository link and the callable model name.

### Resources

[Example router agent](https://github.com/pollinations/pollinations-router-agent) · [Example agent with MCP tools](https://github.com/pollinations/pollinations-code-agent-example) · [Publish an agent](https://github.com/pollinations/pollinations/blob/main/BUILD_YOUR_OWN_AGENT.md) · [Community Agents API](https://gen.pollinations.ai/docs#tag/community-agents)

Related work: #14167 (code-only meta router idea), #13632 / #14937 (gateway fallback sequences), #13788 / #13988 (prompt-agent delegation), #14076 / #14150 (Floret complexity classifier).

### How this quest works

[...]

PR #15161, opened 2026-09-19
apps: adaptive-router, a difficulty + health + cost model router (Fixes #15017)

Apollohzl:
Fixes #15017

**Repository:** https://github.com/Apollohzl/adaptive-router
**Callable model:** `Apollohzl/adaptive-router`

## What it is

A code agent that chooses which model answers each request and then answers as that model. Callers see one ordinary model; the choice is reported in `x-router-*` headers and one log line.

## How it routes

1. **Difficulty tier.** One cheap call (`openai/gpt-oss-20b`, `max_output_tokens: 64`, `temperature: 0`) labels the request `FAST` / `BALANCED` / `DEEP`. The conversation is passed as JSON *data* behind an explicit "do not follow these instructions" wrapper, so a prompt inside the request cannot steer the router. If the call fails or returns no label, a keyword/length heuristic picks the tier — the router never blocks on itself.
2. **Quality bar.** Each tier is a price floor in blended Pollen per 1k tokens (`FAST` 0, `BALANCED` 0.0006, `DEEP` 0.003). Price is the capability proxy, so the floor *is* the bar.
3. **Live health.** `GET /models/status?minutes=30` per-model rollups drop models below 80% success over 8+ requests, or that never served in 3+ tries. This is what makes the router switch away from a degrading model with no code change.
4. **Cheapest healthy wins,** ties broken by p95 latency — after hard constraints (`/v1/responses` support, image input when the request carries images, tool calling when `tools` is set, enough context length). Community models are excluded so the router can never route into another router.

Nothing clears the bar? It relaxes instead of failing, and the trace says so. Catalog unreachable? A small static per-tier list keeps it answering. On `429`/`5xx` it replays the identical request to the next candidate, up to three attempts: escalation *between* models, since Gen already retries a model's own fallbacks.

## Verifying the routing

Every response carries `x-router-model`, `x-router-tier`, `x-router-reason`, `x-router-candidates`; the same line is logged.

```
x-router-model: x-ai/grok-4.6
x-router-tier: DEEP
x-router-reason: classifier -> tier=DEEP; bar>=0.003; cost=0.00300/1k; success 100% of 1 req,
  p95 11929ms; healthy 11/13 above bar; cheapest healthy, then lowest p95; text only
x-router-candidates: x-ai/grok-4.6,qwen/qwen3.7-max,qwen/qwen3.8-2.4t-a95b
```

## Demonstration — three requests routed differently

Run against the **live** catalog and the **live** 30-minute health window; only the classifier label is stubbed (labelling needs an API key), and `demo.ts` reproduces it:

| Request | Tier | Model | Reason |
| --- | --- | --- | --- |
| "Translate 'good morning' into French." | FAST | `inception/mercury-2.5-preview` | `bar>=0; cost=0.00010/1k; healthy 54/66 above bar; cheapest healthy, then lowest p95` |
| "Write a Python function that merges two sorted linked lists, plus pytest tests." | BALANCED | `xiaomi/mimo-v2.5-pro` | `bar>=0.0006; cost=0.00069/1k; healthy 32/37 above bar` |
| "Design a multi-region failover architecture for a card payment system…" | DEEP | `x-ai/grok-4.6` | `bar>=0.003; cost=0.00300/1k; success 100% of 1 req, p95 11929ms; healthy 11/13 above bar` |
| An image of a receipt + "What does this receipt say?" | BALANCED (bumped from FAST) | `mistralai/mistral-large-3` | `request carries images; cost=0.00075/1k; healthy 24/28 above bar` |

## Tests

`node --test agent.test.ts` — 8 tests, all passing:

- three difficulty tiers route to three different models;
- the forwarded body keeps the conversation and only swaps `model`;
- chat-completions style `messages` are normalised to a Responses body;
- degraded models are skipped, equal-price ties break on p95 latency;
- image input is bumped out of FAST and only picks vision models;
- a failing router model falls back to the heuristic tier;
- a 503 escalates to the next candidate and reports every attempt;
- an unreachable catalog still answers from the static fallback.

## Alpha feedback

- The code-agent contract is clear and the one-file deploy works well. Two things would help: a way to set response headers that is guaranteed to survive the gateway (custom headers are easy to set but I could not confirm they reach the caller), and access to agent stdout in the dashboard — right now the `console.log` trace is only verifiable indirectly.
- A `GET /v1/models` field marking *managed agents* would let routers exclude them generically; today I exclude everything under `community/` to avoid routing into another router.

## Files

- `apps/agent-adaptive-router/agent.ts` — the agent
- `apps/agent-adaptive-router/README.md` — short README


voodoohop:
Thanks for the serious submission. The tiered design, request preservation, live evidence and tests were useful. We selected other routers for the quest. A 2-Pollen serious-attempt bonus is recorded; claim it from Quests with your linked GitHub account.

voodoohop:

PR #15176, opened 2026-09-19
feat: add kwaivgi/kling-v3.0-std via OpenRouter

tomdacatto:
- Adds `kwaivgi/kling-v3.0-std` (Kuaishou Kling 3.0 Standard) via the OpenRouter video API, paid-only, no aliases.
- Text-to-video and start/end-frame image-to-video, 720p, 3-15s (default 5s), optional synchronized audio.
- Cost $0.084/s + $0.042/s when audio is on ($0.126/s with audio), x1.055 OpenRouter fee (#14895). Audio is billed through `completionAudioSeconds`, as Veo does.
- No fallback. Replicate `kwaivgi/kling-v3-video` exists but costs 2x ($0.168/s, $0.252/s with audio) and needs a new handler.
- Adds the Kuaishou logo (Simple Icons, CC0). Pro tier (`kling-v3.0-pro`) is not included.

**Verification**
- Handler, registry, dispatch, frame-count test, handler unit tests and price snapshot are updated. Not run locally (sandbox can't run the Vitest/Workers suites); CI is the check.
- No live probe: no OpenRouter key in this environment. Needs a live text-to-video and start/end-frame run, plus a billing check, before merge.

Related: #5321


ElliotEtag:
Blocking before merge:

- Replace the OpenRouter implementation with fal only. Do not add OpenRouter fallback in this PR. The providers have different billing and output contracts, so a fallback cannot share this model's static price safely.
- Use `fal-ai/kling-video/v3/standard/text-to-video` when there are no frames and `fal-ai/kling-video/v3/standard/image-to-video` when a start or end frame is supplied. Map `image[0]` to the start image and `image[1]` to the end image; preserve `audio`, duration, and the requested aspect ratio.
- Preserve the public 720p contract. fal returned 1280x720 in direct testing. OpenRouter received `resolution: "720p"` and `aspect_ratio: "16:9"` but returned 1920x1080, so it is not a compatible fallback.
- Keep fal's direct published prices: $0.084/s silent and $0.126/s with audio. Do not apply the 1.055 OpenRouter credit fee or retain any OpenRouter pricing assumptions.

Verified OpenRouter E2E:

- 3s silent: 3.04s 1920x1080 output, provider cost $0.42.
- 5s silent: 5.04s 1920x1080 output, provider cost $0.42.
- 3s audio: 3.04s 1920x1080 output, provider cost $0.63.

This proves a 5-second minimum: $0.42 silent and $0.63 audio. The current code bills requested seconds, undercharging a 3s request by 0.17724 Pollen silent or 0.26586 Pollen with audio after the multiplier. It also ignores `safeParams.aspectRatio`, which made a gateway `aspectRatio=16:9` request produce square output.

Before requesting another review, run direct and authenticated gateway E2E for 3s silent text-to-video, 3s audio text-to-video, and start/end-frame image-to-video. Verify output streams, aspect ratio, wallet debit, one Tinybird billed event, cache rejoin, and zero extra debit on a cache hit.

Decision for this release: ship fal only. Do not use OpenRouter as primary or fallback yet: our E2E found its five-second billing minimum and a 1080p response to a 720p request, so it does not currently share fal's cost or output contract. Revisit it only with an explicit fallback billing contract.

ElliotEtag:
- The fal route and direct $0.084/s silent + $0.042/s audio cost entries are in place. The focused 61 tests and Gen typecheck pass.
- Blocker: text-to-video still ignores an explicit `aspectRatio` when dimensions are omitted. On this head (`e53f70f`), I called the real `callKlingVideoAPI` with `aspectRatio: "16:9"`, default square dimensions, `duration: 3`, and `audio: false`. It submitted `aspect_ratio: "1:1"` to fal and returned a 960x960 H.264 video. Use the explicit supported ratio before deriving one from dimensions; reject unsupported Kling text ratios with a clear 400 instead of silently choosing another ratio. Cover this exact input in a regression test.
- The live generation completed in 208 seconds. After the fix, run authenticated gateway E2E for text silent/audio and start/end frames, including disconnect/rejoin, completed-cache retrieval, one debit and one billed Tinybird event. Repeat real generations to check the slow tail against the 300-second public request budget.
- Update the PR title and description to the final fal-only route and direct fal pricing; they still describe the former provider and fee.

PR #15178, opened 2026-09-19
feat: add z-ai/glm-5.3-flashx via OpenRouter

tomdacatto:
- Adds `z-ai/glm-5.3-flashx` (Z.ai GLM-5.3 FlashX, text + image in, text out) via OpenRouter, paid-only, no aliases.
- Route pinned to Z.AI `z-ai/fp8` (OpenRouter's only endpoint for this model), `allow_fallbacks: false`. No fallback: none found. The sibling `glm-5.3-flash` runs on Fireworks, but FlashX there is unconfirmed.
- Cost $0.37 in / $0.075 cached / $1.25 out per 1M, x1.055 OpenRouter fee (#14895). 1M context, 131K max output.
- Video input is not exposed. OpenRouter lists it, but it is untested here.
- Reuses the existing Z.ai logo.

**Verification**
- Registry, config, availability, resolver test and price snapshot are updated. Not run locally (sandbox can't run the Vitest/Workers suites); CI is the check.
- No live probe: no OpenRouter key in this environment. Needs a live text + image request before merge.
- GLM-5.3 and 5.3-flash use a mandatory-reasoning transform. I didn't apply it here because FlashX's behavior is unverified. Check whether reasoning-off requests need it.

Related: #5321


ElliotEtag:
- E2E-tested `33971a4730b461d925f6c51f10e3cd20a889699f` through authenticated local Gen → real OpenRouter (`z-ai/fp8`) → wallet settlement and Tinybird.
- **Confirmed failure:** `reasoning: {"enabled": false}` and `reasoning_effort: "none"` return HTTP **400**: `Reasoning is mandatory for this endpoint and cannot be disabled.` Streaming with `reasoning.enabled=false` also returns 400. Failed requests deduct no Pollen.
- **Suggested fix:** add `transform: mandatoryReasoning` to the [FlashX entry](https://github.com/pollinations/pollinations/blob/33971a4730b461d925f6c51f10e3cd20a889699f/gen.pollinations.ai/src/text/availableModels.ts#L688-L691), matching the neighboring GLM-5.3/Flash entries. This preserves their existing convention: an off request uses the provider’s mandatory reasoning default; it does **not** disable reasoning, and reasoning tokens remain billable.
- Add regression coverage for both inputs. Before merging, rerun each through streaming and non-streaming gateway requests; verify successful content, terminal stream usage/`[DONE]`, and wallet/Tinybird reconciliation.
- Other checks passed: text, streaming, vision with 1/2/10 images, tool calls, JSON, native text, Responses API, paid-only access, and repeated-request caching without another debit. Ten billed Tinybird events totaled **0.00747426 Pollen**; measured charges matched token pricing, including cached input, reasoning, and the 5.5% OpenRouter fee. The PR’s focused resolver test passed. No tracked code changed.

ElliotEtag (approved):
- Reviewed `8543cb9b55a42ab7b15ae23b64eec6c83439c31a`; CI is green and the focused reasoning/resolver suites pass (107 tests).
- Real Portkey → OpenRouter `z-ai/fp8` checks now pass with `reasoning_effort: "none"` in both modes. Non-streaming returned content and complete usage; streaming returned content, terminal usage, and `[DONE]`.
- The mandatory-reasoning transform correctly removes the unsupported off value while preserving billed reasoning tokens. Prior authenticated gateway E2E already reconciled wallet/Tinybird pricing, including the 5.5% OpenRouter fee. No blocking findings.

PR #15186, opened 2026-09-19
feat(cli): add whoami alias for auth status

### LINKED ISSUE #15034: [QUEST] Add polli whoami alias
### Goal

Add `polli whoami` as a top-level alias of `polli auth status`. It should show the same account, key, and balance, including with `--json`. Reuse the existing status behavior.

### How this quest works

Multiple contributors may submit PRs. We select the simplest complete approach; the author of the merged PR that closes this issue can claim the reward on the Quests page.

### Reward

3 Pollen

La5u:
Adds `polli whoami` as a top-level alias for `polli auth status`, matching the convention used by npm, wrangler, and vercel.

Fixes #15034


voodoohop:
Thanks for contributing this! We merged #15086 for the same alias, with a small README addition to make it discoverable. Your runtime approach was equally lean.

PR #15197, opened 2026-09-19
feat(polli-cli): polli mcp — install/status/remove Pollinations MCP s…

Jonakss:
…ervers in 8 coding agents

Closes #15014 (QUEST bounty).

- polli mcp install/status/off reading the live catalog (GET /mcp), no hard-coded server list: new servers appear without a CLI release.
- Clients (priority order): claude-code, codex, vscode (workspace), cursor, opencode, gemini, windsurf, cline. Formats verified against primary docs: Claude Code type:http, Codex TOML bearer_token_env_var (+ key stored in CODEX_HOME/.env, never in the config), VS Code servers+inputs with prompt input (no literal key in the file), Cursor url, OpenCode mcp.<name> type:remote enabled:true, Gemini httpUrl, Windsurf serverUrl, Cline streamableHttp.
- Ownership is URL-based (under gen.pollinations.ai), not name-based, so off never touches unrelated config; install is idempotent; every write goes through the harness snapshot mechanism (atomic write + restore-or- strip on off).
- Dedicated child key per client via resolveHarnessKey (reuses an existing valid key before minting a new one).

Choice (per the issue's invitation to explain): a new top-level 'polli mcp' command rather than extending 'polli harness', and file-based writers for all clients rather than each client's native 'mcp add' CLI. Reasons: one idempotency/ownership mechanism shared across 8 heterogeneous config formats (JSON envelopes vs TOML), deterministic snapshot-reversible behavior, no dependency on CLI flag drift between agent versions, and headless/testable operation. The docs show the native commands per client for users who prefer them.

Tested end to end on a real host: opencode (real user config — install verified, then off restored the file byte-identical), plus the full 8-client matrix at adapter level (21 new tests). npm run test: 166 passed. npx biome check: clean.


voodoohop:
- Removing one server should keep the others. We chose direct edits to owned entries rather than restoring whole config snapshots; the Codex key setup also needed connecting to installation.

Jonakss:
Thanks @voodoohop for the detailed comparison and feedback, and congrats @afanasevmylife on a solid implementation. Snapshot-based removal was our heavier take; direct edits to owned entries is the simpler call, agreed. 

This was also the first full run of our multi-agent workflow (security review, adversarial critique, e2e verification on a real host), so the exercise paid off either way. 

Happy to jump on future quests! 🌱

PR #15226, opened 2026-09-20
feat: add prism code-review model router

cesus-agent:
Fixes #15017

**Repository:** https://github.com/cesus-agent/prism-code-review-router
**Callable model name:** `cesus-agent/prism-code-review-router`

## What

Adds `apps/agent-prism-code-review-router/` — `agent.ts` (the deployed code agent) plus a README documenting the routing policy. Prism picks which model answers each request, then answers as that model.

## How it routes

Order of operations per request:

1. **Read the wallet first.** `GET /account/balance` (forwarded with the caller's own delegated credentials) → per-request spend cap = `balance / 2`. The cap is derived from the balance, never the other way around. A budgeted key with 2 pollen cannot have a flagship deep pick silently drain it. Invisible balances (unbudgeted keys without `account:usage` get 403) route uncapped — a private wallet never breaks routing.
2. **Classify locally, deterministically.** Regex signal families (security, review, architecture, ops, planning, data, writing) + payload features (unified-diff hunk counts, `+`/`-` line counts, request size, image parts) → `fast` / `balanced` / `deep`. No router LLM call: zero added cost/latency, and the request payload (possibly an unreleased diff) is never sent to a second model just to decide who should answer it. Escalation is deliberately asymmetric: security vocabulary always goes deep, because over-escalating costs pennies and under-escalating costs trust.
3. **Select from live data.** `GET /v1/models` + `GET /models/status?minutes=30` (60 s isolate cache) with hard gates:
   - health: catalog `success_rate < 95%` or >10% 5xx in 30 min dropped; status rows under 10 requests ignored as noise
   - Responses-API support (`supported_endpoints`) — discovered by live testing: many community/proxy models 400 on `/v1/responses` with `unsupported_parameter: model`
   - context fit (chars/3 + 4k margin), input modality (image detection), `tool_calling` when the request needs tools
   - wallet cap: models whose estimated cost (prompt tokens from actual payload size + completion tokens from `max_output_tokens`, default 500) exceed the cap are dropped before ranking; if nothing fits, the cheapest model is picked so the caller gets either a real answer or a clean 402
4. **Rank within the affordable pool.** Fast/balanced: healthy free community models are preferred (zero pollen for the caller); when one degrades it fails the health gate on the next request and the router auto-switches to the next-best — "prefer free community models, switching when one degrades" from the issue. Unproven models (no traffic, status `unknown`) are allowed; only positive degradation evidence (down/degraded status, low success rate on a real sample, >10% 5xx, or a status row where every request failed) excludes a model. Remaining pool ranks by live token price: lowest for fast, median for balanced. Deep: reasoning capability → context length → listed price (flagship proxy) — deep deliberately pays for strength, with the wallet cap as the affordability guard.
5. **Forward almost unchanged.** Only `model` changes, plus an appended engineering rubric on balanced/deep (facts vs hypotheses, security/regression/concurrency/data-loss priorities, file-and-hunk citations, confidence, risks, rollback, verification). Caller-supplied `instructions` are preserved.

## Why these choices

- **Wallet-first routing** — the issue's core value is choosing *between* models; a router that ignores the caller's balance picks models the caller can't afford. Reading balance first is one cheap GET and makes the cap honest.
- **Deterministic classification** — auditable (every route traces to explicit rules), free, and private; classifier-routers add a billable hop and a failure mode before the real work starts.
- **Responses-API gate** — live testing showed community models reject `/v1/responses`; without this gate the router would confidently route into 400s.
- **No retries, no second keys** — all calls go through `pollinations(path)` with the caller's delegated authority; gen's declared fallbacks already handle per-model retries, so prism only chooses between models.

## Test evidence (live, run with a test API key)

All four demos routed differently with recorded traces (`x-pollinations-router-*` headers + JSON worker log + the answer's `model` field):

| Request | Tier | Routed to | Trace (reason) |
| --- | --- | --- | --- |
| "What does this function return? function f(x) { return x * 2 }" | fast | `community/YoannDev90/agentic-gt` | small targeted coding request; healthy free community model — 0 pollen; balance=2 |
| "Review this PR diff for regressions" + real diff body | balanced | `community/YoannDev90/agentic-gt` | code review or diff context; free community preferred; balance=2 |
| "Audit this OAuth migration for race conditions and security issues" | deep | `openai/gpt-6-astra` | security-sensitive change; strongest affordable model (reasoning-capable, 1050000 ctx); balance=2 |
| "Design a technical plan with acceptance criteria… for migrating our webhook delivery to a queue" | balanced | `community/YoannDev90/agentic-gt` | design and planning work; free community preferred; balance=2 |

Two zero-spend unit checks against `choose()` verify the guards: (1) wallet cap — with balance 0.02 (cap 0.01) and a synthetic expensive flagship, the flagship is dropped for the cheap model (`wallet-capped (balance 0.02, max spend 0.0100)`); (2) community switchover — a healthy free community model wins the fast tier, and the same model with a status row of 41 requests / 0 successes is dropped in favor of the paid model.

Verified against the live `gen.pollinations.ai` API with `npx tsc --noEmit` (type-clean), `git diff --check` clean, 17 KB source (well under the 64 KiB agent limit).

## Alpha feedback (for the bonus)

1. **Catalog vs reality drift:** `/v1/models` advertises `supported_endpoints` but community entries routinely 400 on `/v1/responses` anyway. The gate fixes it for prism, but a catalog that reflected real Responses-API support would save every router author this debugging loop.
2. **Balance visibility asymmetry:** budgeted keys report `balance` without scope, but unbudgeted keys get 403 without `account:usage`. Since agents already run on delegated auth, a router can only *suggest* affordability for those callers. A documented way for delegated tokens to read the same cap the gateway enforces would make wallet-aware routing exact instead of best-effort.
3. **Trace durability:** confirmed prior reports — custom headers survive on `/v1/responses` but not through Chat Completions conversion. Prism logs the same trace as JSON so it's verifiable either way.


voodoohop:
Thanks for the serious submission. The wallet-aware routing and deterministic safety-focused classification were interesting. A 2-Pollen serious-attempt bonus was intended, but the grant could not match a linked Pollinations account for cesus-agent. Link GitHub and let us know if you want us to retry it.

voodoohop:

PR #15236, opened 2026-09-20
feat: add qwen/qwen-image-2.1 via fal

tomdacatto:
- Adds `qwen/qwen-image-2.1` (Alibaba Qwen Image 2.1) via fal, paid-only, no aliases. Text-to-image and editing with up to 10 reference images.
- Routes to `alibaba/qwen-image-2.1/text-to-image`, or `/edit` when reference images are supplied. Reuses the Qwen Image 3 fal pattern.
- No fallback: Replicate has `qwen/qwen-image-2`, but it is a different version, not 2.1.
- Cost: $0.02 per output megapixel for generation; $0.11/3 per input and output megapixel for edits (fal bills both). Metered as megapixels, with the size rounded to multiples of 32 as fal does.
- Existing `qwen/qwen-image` and `qwen/qwen-image-3` are untouched.

**Verification**
- Handler unit tests, registry entry and price snapshot are added. Not run locally (sandbox can't run the Vitest/Workers suites); CI is the check.
- No live fal probe: no FAL_KEY in this environment. Needs a live generate and edit request, plus a billing check, before merge.
- Fal's pixel limits for this model are undocumented, so the handler doesn't pre-validate size.

Related: #5321


ElliotEtag:
- Verified commit `d0f60a9e7944a840ed4a149e294b87bd99a08f30` on September 22 through **authenticated local Gen → real fal → local wallet settlement**, using an isolated funded test user. All three requests returned HTTP 200; no provider mocks.
- **Aspect ratio ignored:** `GET /image/{prompt}?model=qwen/qwen-image-2.1&aspectRatio=16:9&seed=592230`, without width/height, returned **1024×1024**. [Handler lines 65–70](https://github.com/pollinations/pollinations/blob/d0f60a9e7944a840ed4a149e294b87bd99a08f30/gen.pollinations.ai/src/image/models/qwenImage21Model.ts#L65-L70) never resolve `aspectRatio`. Fix ratio handling when dimensions are not explicit.
- **Input usage omission causes undercharging:** sent identical 1024×1024 PNG bytes through `POST /v1/images/edits` as data URIs, using separate seeds. `image/png`: **1.048576 input MP**, wallet debit **0.07689557 Pollen**. `application/octet-stream`: **0 input MP**, debit **0.03844779 Pollen**. Both returned **1.048576 output MP** and 1024×1024 images. [Lines 52–58](https://github.com/pollinations/pollinations/blob/d0f60a9e7944a840ed4a149e294b87bd99a08f30/gen.pollinations.ai/src/image/models/qwenImage21Model.ts#L52-L58) forward references despite failed dimension parsing. Read dimensions from bytes or reject unmeterable references before inference.
- Existing handler tests: **5/5 passed**; add regression coverage for both cases. No `Missing conversion rate` warnings.
- Limits: actual fal invoice charges remain unverified. Tinybird lookup returned no rows for the three successful request IDs, so telemetry reconciliation is **not confirmed**.

ElliotEtag (changes requested):
- Blocking billing-observability mismatch: `qwenImage21Model.ts` emits fractional megapixels (`promptImageTokens` / `completionImageTokens`), while `generation_event_v2` declares both image-count columns as `UInt32`. The earlier authenticated gateway E2E deducted the wallet correctly but produced no Tinybird rows; the staging economics endpoint still has no row for this model.
- Fix by either metering integer units with matching per-unit registry rates, or landing and deploying a prerequisite Tinybird schema change that preserves fractional units in both workspaces. Then rerun authenticated generation + edit and reconcile usage, cost, price, wallet debit, and one billed event each.
- Latest direct fal checks otherwise pass: 16:9 generated 1376×768 with 1.056768 output MP; edit metered 1.056768 input MP and 1.048576 output MP. Focused tests pass (39), and CI is green.

PR #15240, opened 2026-09-20
Add Codex Router and Claude Code Router harness adapters (polli harness)

### LINKED ISSUE #15218: [QUEST] Add Pollinations setup for Codex and Claude Code routers
### How this quest works

Multiple contributors may submit PRs. We will select the best completed approach, preferring correctness, clarity, and minimal code. The author of the selected merged PR receives the reward; no issue assignment is required.

Experience with Codex Router or Claude Code Router is especially welcome. In your PR, briefly share your relevant background and how you verified both integrations.

### Goal

Make Codex and Claude Code ready to use Pollinations through the documented Codex Router and Claude Code Router integrations.

This is one quest with two required harness adapters. The router projects themselves are out of scope; the work belongs in the Polli CLI harness workflow.

### Outcome

- `polli harness codex on|off|status` configures Codex through Codex Router's current supported integration path.
- `polli harness claude-code on|off|status` configures Claude Code through Claude Code Router's current supported provider and agent-profile path.
- If either router or its client is missing, the command offers the official installation/setup experience and stops before login, key creation, or unrelated config changes.
- Polli login creates or reuses a dedicated child key for each integration; no second Polli login is required, and keys are stored only in the router/client-owned protected configuration.
- Both adapters use the live compatible Pollinations model catalog and let the user choose `--model`; they do not add a stale hardcoded model registry.
- The adapters use the routers' documented protocol translation/configuration instead of assuming that Codex or Claude Code can call Pollinations directly.
- The Codex path preserves the user's existing Codex ChatGPT login, native models, profiles, and settings.
- The Claude Code path preserves the user's existing Claude login, settings, providers, and unrelated router configuration.
- `status` clearly reports installation, router readiness, Pollinations provider/key state, selected model, and any missing prerequisite.
- `off` restores untouched files byte-for-byte when possible and otherwise removes only Pollinations-owned provider, key, model, profile, and environment entries. It must leave the router and unrelated providers installed.
- Each adapter includes a cheap end-to-end smoke check (for example, a one-word `pong` request) before any paid task, and verification checks the expected output plus the Pollinations usage side effect under the dedicated key.
- The complete setup is verified end to end for both clients, including a real streamed request and a tool-capable coding request through the configured router.
- Tests cover missing binaries, current router/client version compatibility, fresh and existing configs, pre-existing user entries, byte-for-byte restore, surgical cleanup after outside edits, key reuse, model selection, and failure rollback.
- Scalar documentation explains setup, status, model selection, troubleshooting, version-skew triage, smoke testing, and removal for both integrations.
- The implementation reuses the shared `polli harness` lifecycle, key, model, and snapshot helpers and follows the existing adapter conventions.

Do not copy router credentials into chat, shell history, or a second config file. Do not replace native Codex or Claude Code subscriptions; the router integration should be an explicit Pollinations provider/profile that users can select.

### Resources

- Polli harness foundation: https://github.com/pollinations/pollinations/pull/14107
- Existing harness quests and implementations: #14118, #14119, #14120, #14121
- Codex Router: https://github.com/duolahypercho/codex-router
- Claude Code Router: https://github.com/musistudio/claude-code-router
- Pollinations API and model discovery: https://gen.pollinations.ai
- Polli CLI source: https://github.com/pollinations/pollinations/tree/main/packages/polli-cli

### Reward

15 Pollen

rekty:
Fixes #15218

Adds the two required harness adapters so `polli harness` can connect both documented local routers to Pollinations.

## `polli harness claude-code on` — Claude Code Router (@musistudio/claude-code-router)
- Writes a **native generic provider** into ccr's config database (`%APPDATA%/claude-code-router/config.sqlite` on Windows, `~/.claude-code-router/config.sqlite` elsewhere) using ccr's own schema: `@ai-sdk/openai-compatible` → `https://gen.pollinations.ai/v1`, `apiKey` = dedicated `polli-harness-claude-code` key.
- Registers the live compatible Pollinations model catalog; default `qwen/qwen3.7-flash` (`--model <id>`). Models whose id starts with a gateway type prefix (`openai/`, `anthropic/`, …) are rejected with a clear message because ccr treats the first path segment as its own provider-type namespace.
- **Smoke-tests a real chat round-trip** through the router before reporting success.
- `off` removes only the `pollinations` provider, its key, and `pollinations/*` profile entries — byte-for-byte restore if nothing else changed.

## `polli harness codex on` — Codex Router (duolahypercho/codex-router)
- Writes a generic provider into `user-models.json` + credential into `generic-provider-credentials/` under the router state dir (`~/.codex/codex-router/`, `$CODEX_ROUTER_STATE_DIR` override). The dedicated key stays **outside `config.toml`** — the router's own documented privacy rule (verified via `doctor codex`).
- Then runs the router's **official `client-setup codex`**, which wires `~/.codex/config.toml` (or `$CODEX_HOME`) with the managed `[model_providers.codex-router]` block. Verified live: `{"target":"codex","configured":true}` and the managed block appears exactly as the router writes it.
- **Never touches a `config.toml` already managed by another router** (e.g. ccr's Codex profile) — `on` stops cleanly instead of clobbering.
- `off` runs the router's `client-remove codex` and deletes only the Pollinations provider + credential files.

## Shared
- New `src/harnesses/smoke.ts`: `polli harness <harness> smoke` — one real chat round-trip using the harness key, for post-setup verification.

## Verification
- 22 new unit tests (vitest): missing binaries, fresh/existing configs, restore semantics, slot reuse, prefix-collision validation, APPDATA isolation.
- Live E2E on Windows (Node 24):
  - ccr installed via npm, `on` → real smoke "pong" through the router → Anthropic-protocol request via the running gateway → streamed response + **tool calling** (`get_weather`) → usage recorded on the dedicated key.
  - codex-router cloned + installed, `client-setup codex` produced the managed block in an isolated `$CODEX_HOME`; gateway started on 127.0.0.1:4202; `doctor codex` passes the privacy check with our credential layout. Full request path through the codex gateway requires the router's heavy LiteLLM repair (`install.ps1 -ForceDeps`) which is the router's own installer, out of scope for this PR.
- `biome check` clean on all touched files; the two `tsc` errors in `src/commands/harness.ts` and build-time DTS errors are pre-existing on `upstream/main` (verified by stashing this PR's changes).
- Docs: new sections in `CODING_HARNESSES.md`, commands added to `packages/polli-cli/README.md`.

Generated with Codebuff
Co-Authored-By: Codebuff <noreply@codebuff.com>


rekty:
## Full live E2E for the Codex Router path (closing the gap flagged in the PR description)

The earlier "gateway not fully repaired" caveat is now resolved — the full request chain was verified live on this machine, and it surfaced (and fixed) a real adapter bug.

### Bug found & fixed (commit 9f5aadb)
Codex Router's provider-credential store validates credential ids against `^cred_[A-Za-z0-9_-]{16,64}$` (`src/provider-credential-store.mjs`) and **silently drops the whole store** when an entry fails. Our original `cred_pollinations` (12 chars) was rejected, so every upstream request failed with an auth error even though all files were in place. Fixed:
- credential id renamed to `cred_pollinations_harness` (regex-valid),
- `on` migrates the legacy id, `off` strips both, `status` recognizes both,
- regression test asserts the id satisfies the router's contract (13/13 codex harness tests pass; biome clean).

### Verified live chain (fresh `CODEX_HOME`, real keys)
1. `polli harness codex on` → provider + credential + key + user-model written, `configured: true`
2. Router's own `doctor` repairs the LiteLLM venv, generates `internal-secret`/`caller-secret`, starts the managed service (router :4202, gateway :4200, forwarders)
3. Caller request → `POST http://127.0.0.1:4202/v1/responses` with `model: pollinations/openai/gpt-5.4-nano` → **completion served by gen.pollinations.ai** (`"pong"` / `"e2e-ok"`), response `model: pollinations-openai-gpt-5-4-nano`
4. `discover-models pollinations` probes our `/v1/models` through the router's official flow; `curate-models pollinations --models …` (non-interactive, stdin closed) publishes the model to the picker; `refresh-catalog` merges it into `merged-models.json`

Notes for reviewers of the router itself (not this PR): `doctor` refuses repair when `config.toml` references a catalog outside its state dir even if that block was written by its own `client-setup` (fixable by running repair before client-setup), and `curate-models`/`promptLine` spawn `pwsh.exe` on Windows instead of falling back to `powershell.exe` like its other prompts do — worth an upstream issue.

cc @strix-security — re-review welcome on 9f5aadb.

rekty:
## Real-client evidence: OpenAI Codex CLI 0.146.0 + Claude Code 2.1.220, live on this machine

Both real clients are installed here, so I ran the full chain **through each client's own process** — no mocks, no recorded fixtures.

### 1. OpenAI Codex CLI 0.146.0 → Codex Router → Pollinations

Isolated `$CODEX_HOME` (real client config written by the router's `client-setup codex`), real binary `codex.exe`:

```
$ codex exec --skip-git-repo-check -m "pollinations/openai/gpt-5.4-nano" "Reply with exactly one word: pong"
codex
pong
tokens used 14,744
```

Tool-calling round-trip (sandbox workspace-write, real file created by the client):

```
$ codex exec --sandbox danger-full-access -m "pollinations/openai/gpt-5.4-nano" \
    "Create codex-e2e.txt containing exactly: codex-tool-ok. Then read it back."
`codex-tool-ok`
tokens used 4,550
$ cat /tmp/codex-e2e.txt
codex-tool-ok
```

Auth detail worth documenting (cost me a few 401s): Codex CLI ignores `OPENAI_API_KEY` for custom providers when `requires_openai_auth = true` — it needs a local `auth.json` in `$CODEX_HOME`. Writing the router's **caller secret** there (not the upstream provider key) makes the client authenticate to `127.0.0.1:4202` correctly.

### 2. Claude Code 2.1.220 → router Anthropic surface → Pollinations

Isolated `CLAUDE_CONFIG_DIR` (the user's real `~/.claude/settings.json` pins `ANTHROPIC_BASE_URL` via its `env` block, which overrides shell env — worth a note in docs; isolated config dir sidesteps it):

```
$ claude -p "Say exactly: pong" --model "pollinations/openai/gpt-5.4-nano"
pong
```

Real tool use through the routed model (Anthropic `tool_use` block emitted by Pollinations, executed by the client):

```
$ claude -p "Use the Bash tool to run exactly: echo tool-ok. ..." --max-turns 4
→ stream-json shows: {"type":"tool_use","name":"Bash","input":{"command":"echo tool-ok"}}
→ final text: tool-ok
```

The Anthropic surface lives at `/_codex-router/<caller-secret>/anthropic/v1/messages` (capability-in-path, same authority as the Responses leaf); both non-stream and SSE streaming (`message_start` → `content_block_delta` → `message_delta end_turn`) verified.

### 3. Third-party bug found while testing (not ours, not the router's fault — LiteLLM)

The router's bundled **LiteLLM** crashes streaming Responses requests when the upstream emits a chunk with empty `choices` (Pollinations sends a usage-only final chunk like many OpenAI-compatible providers): `streaming_iterator.py` indexes `choices[0]` unguarded → `IndexError` → stream dies before `response.completed`. Reproduced 100%, then patched locally (guard + `try/except IndexError` in 3 spots) and the same request then completes. **Upstream-worthy report for LiteLLM**; no adapter change needed — with the patched venv the unmodified adapter flow works end to end.

### Cleanup
All E2E services stopped, isolated homes removed, the temporary LiteLLM venv patch and its backup removed. No changes to the adapter code in this PR from this round — it's pure runtime evidence.

rekty:
@strix-security re-review please — no new code commits since `9f5aadb` (only PR description update + this real-client evidence).

rekty:
### Follow-up on the LiteLLM bug from the real-client testing

Filed upstream as **BerriAI/litellm#42484** with a self-contained repro (fake SSE upstream, no network/keys).

Key findings while filing:

- The exact guard fix already exists — **PR #34455** (merged 2026-09-16) added `if not chunk.choices: return` / `if not choices: return ""` to `_ensure_output_item_for_chunk` / `_get_delta_string_from_streaming_choices`.
- I unpacked the release wheels to check where it actually shipped: **1.96.0 (bundled by codex-router) and even the current stable 1.102.0 are still unguarded; the fix first appears in the 1.103.0rc1 wheel** (verified present there) and on main.
- The issue therefore asks for a stable cut containing #34455, plus a sweep of the remaining unguarded `choices[0]` reads in the same file (`_is_reasoning_end`, completed-event builders) — an empty-choices chunk reaching those would crash from a different line.

Nothing changes for this PR: the adapter is correct; the crash is in the router-bundled LiteLLM dependency and is fixed upstream as soon as ≥1.103.0 ships stable (or via the local venv patch documented in my earlier comment).

rekty:
### Adapter improvement from the real-client testing: auto-write `$CODEX_HOME/auth.json` (7277abc)

The live Codex CLI runs exposed the last manual step: **Codex CLI reads `auth.json` for custom-provider auth and ignores `OPENAI_API_KEY` from the environment** for this purpose (`requires_openai_auth = true` in the managed provider block). Until now `on` left that file to the user, so a fresh install still 401'd with "requires its configured caller authentication" until auth.json was created by hand.

**`on` now writes it automatically:**
- `$CODEX_HOME/auth.json` = `{"OPENAI_API_KEY": <router caller-secret>}` — the router-generated local capability, never the upstream key (still no inline secrets anywhere).
- Written atomically (0600) like the rest of the state; included in the snapshot so `off` restores a pre-existing auth.json **byte-for-byte**.
- **Ownership-aware `off`**: the surgical strip removes auth.json only when its `OPENAI_API_KEY` equals the current caller secret (i.e. it's ours). A user's ChatGPT login or a foreign key is left untouched.
- Graceful skip with guidance when the router has no caller secret yet (installed but never started) — we never invent a secret the router wouldn't recognize.
- `status` now reports `authJson: true/false`.

Verified live end-to-end in the earlier real-client runs (this is exactly the file my manual E2E had to create by hand): `codex exec` → `pong`, tool-calling round-trip → file created and read back.

Tests: 16/16 codex suite green (new cases: auth.json content, ownership detection incl. ChatGPT-login/foreign-key/invalid-JSON, restore byte-for-byte on `off`, user-owned file preserved on surgical strip). `biome check` clean on touched files; remaining `tsc` errors are the pre-existing ones on `upstream/main`.

rekty:
### Adapter improvement: wire the Claude Code CLI itself via `~/.claude/settings.json` (49457b3)

The live Claude Code runs exposed the same class of gap as Codex's auth.json: **the CLI never reaches the gateway even when the ccr config row is perfect**, because Claude Code's own `~/.claude/settings.json` pins `ANTHROPIC_BASE_URL` in its `env` block — and that block overrides shell environment variables (this is why my first live attempt with `ANTHROPIC_BASE_URL` exported silently went to the default endpoint and retried into the void).

**`polli harness claude-code on` now completes the last mile**, writing exactly what ccr's own profile takeover writes (verified against `@musistudio/claude-code-router` 3.x `cli.js`):

- `ANTHROPIC_BASE_URL` / `ANTHROPIC_API_BASE_URL` / `CLAUDE_AGENT_API_BASE_URL` → gateway origin derived from the config row's `HOST`/`PORT` (0.0.0.0 normalized to 127.0.0.1, same as ccr's `No()`)
- `ANTHROPIC_AUTH_TOKEN` → the gateway-issued `serviceToken` from `service.json` (no upstream key, no apiKeyHelper shadowing)
- `NO_PROXY`/`no_proxy` for the loopback, same as ccr
- An existing ccr-owned `apiKeyHelper` is removed so it can't shadow our token; everything else in the file is untouched

**Safety identical to the codex auth.json pattern:**
- Snapshot-first: `off` restores the user's settings.json **byte-for-byte** when the file is untouched since `on`
- Surgical strip otherwise: only env keys whose values are *provably ours* (exact gateway origin / exact token) are removed — a user's own gateway config, custom env keys, and theme survive
- Skipped gracefully with guidance when the gateway has never started (no service token) — we never write a URL the gateway isn't listening on
- `status` now reports `settingsEnv: true/false`

Tests: 18/18 claude-code suite green (new: gateway-origin derivation incl. IPv6/0.0.0.0, env upsert idempotency, unrelated-key preservation, surgical strip keeping foreign gateway config, full on→off lifecycle with byte restore, skip-when-no-token). Combined with 7277abc (Codex auth.json), both harness adapters now take the CLI all the way from "router configured" to "client requests actually arrive".

rekty:
### Rebased onto current main (4558038)

`feat/router-harnesses` is now up to date with `main` (includes #15032, which added the `mcp` command family and touched `src/index.ts`). No conflicts: the MCP work and these harness adapters live in separate files (`src/mcp/*` vs `src/harnesses/*`). Both harness suites re-verified green after the rebase — 34/34 (codex 16 + claude-code 18).

rekty:
@strix-security re-review please — rebased onto main; new commits since the last review are 38cfbbf (codex auth.json) and 4558038 (claude-code settings env).

PR #15260, opened 2026-09-21
feat(agent): add adaptive health-cost model router

muhamedsohaib:
Fixes #15017

## What this adds

`Pollinations Adaptive Router` is a code agent that chooses a downstream text model using request complexity/modality plus live Pollinations catalog, cost, health and latency signals.

Public test repository: https://github.com/muhamedsohaib/pollinations-adaptive-router

Callable model: `muhamedsohaib/pollinations-adaptive-router`

### Routing

The request is classified into `FAST`, `BALANCED`, `CODE`, `DEEP`, `MULTIMODAL`, or `MULTIMODAL_DEEP`. Each profile has a small quality-bounded candidate pool. The router fetches `/v1/models` and `/models/status?minutes=30` in parallel, filters incompatible input modalities, then scores candidates on blended token price, recent 5xx rate, fallback rescues, p95 latency, and a light quality-rank prior.

4xx errors are deliberately excluded from model-health scoring because caller-side validation errors are not reliable evidence of an unhealthy upstream model.

### Live proof: three requests, three different routes

After deploying commit `ebc70d1b6e43ed3a0f94b7ba79be30d146bb369d`, three live requests through the callable agent produced these usage-history entries on 2026-09-21:

- `FAST` short arithmetic → `openai/gpt-5.4-nano` at 09:00:07, recorded cost `$0.0000`
- `CODE` small Python debugging task → `openai/gpt-5.4-mini` at 09:00:44, recorded cost `$0.0002`
- `DEEP` architecture trade-off synthesis → `x-ai/grok-4.3` at 09:01:10, recorded cost `$0.0018`

The selected model may legitimately change later as health, latency, and pricing change. That is the point of the router.

### Verifiable route trace

The downstream answer is returned unchanged, while the response includes:

- `x-pollinations-router-model`
- `x-pollinations-router-profile`
- `x-pollinations-router-reason`

The reason header contains the profile, weighted score, blended price, 5xx rate and p95 latency used for the decision.

### Tests

The public repository has 7/7 passing Node tests covering:

- request-profile classification
- cheap/healthy FAST routing
- CODE routing
- DEEP routing
- avoidance of a cheaper model with 35% recent 5xx failures
- modality compatibility filtering
- preservation of the downstream response plus routing trace headers

Run with `node --test agent.test.ts` on Node 24+.

### Alpha feedback

One deployment sharp edge surfaced during testing: immediately after `polli agents create --config code-agent.json`, the agent existed in `polli agents list` but its callable model alias still returned `Invalid model or alias`. Running `polli agents sync <agent-id>` deployed the repository commit and made the callable model appear in the catalog. If that sync step is expected for freshly created code agents, it would be useful to call it out explicitly in the create flow/docs; if not, create may be returning before the initial deployment is registered.


voodoohop:
- We like the separate model pools for coding, reasoning, and multimodal requests and would love to see people try this router.
- Would you like public publishing access on Pollinations? We can enable it so you can publish under your own account and keep the live agent updated from your repository, independently of this PR.
- Let us know if you’re interested!

voodoohop:
Thanks for the serious submission. The modality-aware profiles, live proof and 7-test suite were useful. We selected other routers for the quest. A 2-Pollen serious-attempt bonus is recorded; claim it from Quests with your linked GitHub account.

voodoohop:

voodoohop:
- Thanks for reporting that the agent appeared in the list but could not be called yet. Opened #15370 to investigate and make the creation output clearer.

PR #15262, opened 2026-09-21
feat(polli-cli): install Pollinations MCP servers into coding agents

Marcus-Mok-GH:
Wire any Pollinations-hosted MCP server from the live catalog into the coding agent or IDE a developer already uses, with a single `polli mcp` command.

- Adds `polli mcp list|install|remove|status`.
- Live catalog, no hardcode: `list` reads `GET /mcp`; install wires whatever the gateway returns, so new servers appear without a CLI release.
- Dedicated key per client, minted through the existing `resolveHarnessKey` flow (`polli-harness-mcp-<client>`), and reused on re-install so no orphaned secret keys build up on `/account/keys`.
- Secret by reference where supported: Codex uses `bearer_token_env_var` with the key in `~/.codex/.env` (0600); VS Code uses a `promptString` input so the literal key never lands in `mcp.json`.
- Idempotent install; `remove`/`off` deletes only Pollinations-owned entries (matched on the gateway's `/mcp/` URLs, including Zed's `mcp-remote` bridge args), leaving unrelated config untouched.
- All 13 clients in the issue's priority order; prefers each client's own `mcp add` where one exists, writes the config file where it does not.
- Docs: user-facing install/status/removal per client in `gen.pollinations.ai/src/docs/mcp.md`.

Verified `npm run test` (32 new tests + full polli-cli suite) and `npx biome check` are clean. Clients exercised end to end against the live catalog: `polli mcp list` returns all 5 servers; JSON-file clients (VS Code, Cursor, OpenCode, Copilot CLI, Windsurf, Cline, Kiro, Zed, Warp) and Codex TOML parsing are unit-tested for exact config shape, idempotency, ownership-scoped removal, and key recovery; CLI-native clients (Claude Code, Codex, Gemini, Amp) follow their documented `mcp add` argument shape.

Fixes #15014


voodoohop:
- This was a similar, focused approach. We chose #15032 as the base and fixed the shared repeat-install and client-config issues there.

voodoohop:
- Your focused client-adapter approach was a useful alternative. Thanks for contributing!
- We opened #15282 to add you to the agent publishing allowlist. Access will be available after it is merged and deployed.
- Agent guide: https://github.com/pollinations/pollinations/blob/main/BUILD_YOUR_OWN_AGENT.md

PR #15277, opened 2026-09-21
fix: separate regular usage from legacy and internal traffic

ElliotEtag:
- Separates regular users, legacy public APIs, and internal/dev traffic (including automated monitoring), so heavy shared keys no longer inflate customer usage.
- KPI now measures regular usage, with one legacy row that rotates between requests and successful requests. Financial ratios that cover all traffic are clearly labelled.
- Adds a traffic selector to Observability and Model Monitor, with regular users selected by default.
- Applies the same rules to historical graphs, activation and retention. Filters specific key IDs, so other usage on the same account stays included.
- Verified: 81 tests, all three app builds, and staging data checks. Historical totals reconcile. Production rollout still needs the Tinybird updates and dashboard/API deployments.


voodoohop:
- Superseded by #15399 (merged), which splits legacy traffic in the Grafana dashboards and KPI weekly pipes in ~264 lines
- The remaining difference is the 14-key \`internal\` group; if needed, add it as one more branch in \`generation_events_classified\` in a small follow-up
- Closing, thanks!

